PR #7, opened 2024-06-26
Fixes #2: Improve map functionality to load more stops on pan and zoom

### LINKED ISSUE #2: Load More Stops On Panning or Zooming
Right now, the map only loads stops on first load. We need to improve this by continuing to load more stops whenever the user pans the map (i.e. moves left/right/up/down) or zooms in or out. Check out GoogleMaps.svelte's loadStopsForLocation function to see how to load more stops from the server. The `(lat, lng)` params should come from the current center of the Google Map. 

It's ok to leave all of the existing stops on the map—i.e. for now we are only going to _add_ more stops to the map as the user pans and zooms.

The only thing to watch out for is the possibility of loading _duplicate_ stops. We must not load a stop twice on the map. For instance, if the server returns the stop "15th Ave E & Galer St" from two invocations of `loadStopsForLocation()`, we should ensure that only one marker for that stop is added to the map.

## Acceptance criteria

- [x] Add event handlers for pans and zooms
- [x] Call `loadStopsForLocation` whenever the event handlers fire and append its results to the map. (there should probably be some sort of debounce applied to this)
- [x] Make sure we don't load duplicate stops

tarunsinghofficial:
Fixed #2 

**Tasks done:**

- Added a debounce function to ensure the loadStopsAndAddMarkers function is not called more frequently than once every 300 milliseconds.
- The loadStopsAndAddMarkers function fetches stops for the current center of the map and adds markers after checking for duplicates.
- The markerExists function checks if a marker for a specific stop ID already exists.

**Screenshot:**
bandicam2024-06-2611-01-36-299-ezgif com-video-to-gif-converter

### DIFF
--- a/src/components/map/GoogleMap.svelte
+++ b/src/components/map/GoogleMap.svelte
@@ -3,8 +3,8 @@
   import {
     PUBLIC_OBA_GOOGLE_MAPS_API_KEY as apiKey,
     PUBLIC_OBA_GOOGLE_MAPS_MAP_ID as mapID,
-    PUBLIC_OBA_REGION_CENTER_LAT as lat,
-    PUBLIC_OBA_REGION_CENTER_LNG as lng
+    PUBLIC_OBA_REGION_CENTER_LAT as initialLat,
+    PUBLIC_OBA_REGION_CENTER_LNG as initialLng
   } from '$env/static/public';
 
   import {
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
great work! 👍

PR #33, opened 2024-08-16
Added the feature for hiding the bus route on back to button

### LINKED ISSUE #32: Hide the bus route when the ArrivalDeparture component disappears, instead of the StopPane
## User story

As a transit rider,
I only want to see a route map for as long as I have 'focused' on a single route,
so that I can be more efficient at figuring out where bus lines travel.

## Acceptance criteria

- [x] Upon clicking the back button to go from ArrivalDeparture.svelte back to StopPane, the route map overlay should disappear and stops should reappear.

tarunsinghofficial:
Fixes #32

### DIFF
--- a/src/components/map/GoogleMap.svelte
+++ b/src/components/map/GoogleMap.svelte
@@ -15,9 +15,10 @@
 	import StopMarker from './StopMarker.svelte';
 	import RouteMap from './RouteMap.svelte';
 
-	export let selectedTrip;
+	export let selectedTrip = null;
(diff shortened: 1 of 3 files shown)


aaronbrethorst (changes requested):
Almost there! When I click the "Back" button from the trip details back to the stop pane content, I am seeing the route polyline disappear as expected, but the stops aren't reappearing until I close the modal.

aaronbrethorst (approved):
great work! 👍

PR #36, opened 2024-08-22
Enhancement/polyline-arrow-symbol

Ahmedhossamdev:
This PR introduces the addArrowToPolyline function, which adds a red arrow to a polyline on Google Maps like android app. This can help users easily see the direction of the bus, improving the overall ux. The arrow repeats every 50 pixels along the polyline, and it can be adjusted.

### DIFF
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -39,6 +39,7 @@
 
 			if (shape) {
 				polyline = await createPolyline(shape);
+				addArrowToPolyline();
 				polyline.setMap(map);
 			}
 		}
@@ -52,6 +53,25 @@
 				addStopMarker(stopTime, stop);
 			}
 		}
+	}
+
+	function addArrowToPolyline() {
+		const arrowSymbol = {
+			path: google.maps.SymbolPath.FORWARD_CLOSED_ARROW,
+			scale: 2,
+			strokeColor: '#FF0000',
+			strokeWeight: 3
+		};
+
+		polyline.setOptions({
+			icons: [
+				{
+					icon: arrowSymbol,
+					offset: '100%',
+					repeat: '50px'
+				}
+			]
+		});
 	}
 
 	function addStopMarker(stopTime, stop) {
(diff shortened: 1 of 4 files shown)


Ahmedhossamdev:
without-arrow

To

with-arrow

src/components/map/RouteMap.svelte
```diff
@@ -54,6 +55,25 @@
 		}
 	}
 
+	function addArrowToPolyline() {
```
aaronbrethorst, line 58: I think this should be moved into `src/lib/googleMaps.js`, and take the polyline as a param.

aaronbrethorst (changes requested):
nice addition. please move the code into the googleMaps.js file as indicated!

src/components/map/RouteMap.svelte
Ahmedhossamdev: Done

src/lib/googleMaps.js
```diff
@@ -154,10 +156,33 @@ export async function createPolyline(shape) {
...
 
 	return polyline;
 }
+
+/**
+ * Adds an arrow to the polyline.
+ * @param {google.maps.Polyline} polyline - The polyline to which the arrow will be added.
+ */
+export function addArrowToPolyline(polyline) {
+	const arrowSymbol = {
```
tarunsinghofficial, line 172: Adding a check at the beginning of the function to validate the `polyline` parameter will be great, right @Ahmedhossamdev @aaronbrethorst? Something like this:

```
if (!(polyline instanceof google.maps.Polyline)) {
        console.error('Invalid polyline');
        return;
}
```

src/lib/colors.js
```diff
@@ -0,0 +1,4 @@
+export const COLORS = {
```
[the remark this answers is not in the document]
Ahmedhossamdev: awesome! I will try it

src/lib/googleMaps.js
Ahmedhossamdev: Yep, that’s a great idea for checking the polyline parameter. Moving to TypeScript could also improve type safety and help as the project grows.

aaronbrethorst (approved):
Nice work! 👍

PR #47, opened 2024-09-04
Fix: ModalPane never underlap the top of the window or navbar

### LINKED ISSUE #45: The modal pane should never underlap the top of the window or navigation bar
## User story

as a user with a small laptop screen,
I want to be able to see the close button on stop panes,
so that I can more effectively navigate nearby stops

## Acceptance criteria

- [x] Ensure that the y coordinate of the top of the modal pane is always lower than the y coordinate of the bottom of the navigation bar.
- [x] Make sure the modal pane (or its contents) properly scroll via overflow: hidden when necessary
- [x] Make sure that we don't end up with nested scrollbars inside of the modal pane

tarunsinghofficial:
Fixes #45 

**Screenshot:**
image

### DIFF
--- a/src/components/navigation/ModalPane.svelte
+++ b/src/components/navigation/ModalPane.svelte
@@ -29,7 +29,7 @@
 			use:keybinding={{ code: 'Escape' }}
 			class="close-button"
 		>
-			<FontAwesomeIcon icon={faX} class="dark:text-white font-black text-black" />
+			<FontAwesomeIcon icon={faX} class="font-black text-black dark:text-white" />
 			<span class="sr-only">Close</span>
 		</button>
 	</div>
--- a/src/components/navigation/ModalPane.svelte
+++ b/src/components/navigation/ModalPane.svelte
@@ -3,7 +3,7 @@
 	import { FontAwesomeIcon } from '@fortawesome/svelte-fontawesome';
 	import { faX } from '@fortawesome/free-solid-svg-icons';
 	import { keybinding } from '$lib/keybinding';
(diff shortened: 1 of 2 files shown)


src/components/navigation/ModalPane.svelte
```diff
@@ -34,15 +34,22 @@
...
 <style lang="postcss">
 	.modal-pane {
-		@apply absolute bottom-0 left-0 max-h-[40rem] w-full overflow-y-scroll bg-transparent px-2 shadow-lg md:max-w-prose;
+		@apply absolute bottom-0 left-0 max-h-[40rem] w-[25em] overflow-y-scroll bg-transparent px-2 shadow-lg md:w-full md:max-w-prose lg:w-full;
 		@apply rounded-lg border-b-[1px] border-[#C6C6C8] bg-[#F3F2F8] dark:border-[1px] dark:border-[#C6C6C8] dark:border-opacity-15 dark:bg-black;
 	}
 
+	.modal-content {
+		@apply overflow-y-hidden;
+		max-height: calc(100vh - 6.6em);
```
aaronbrethorst, line 50: 👏

aaronbrethorst (changes requested):
This looks great! I see one change that needs to be made before it can be merged: the contents of the modal pane aren't scrolling as expected, so I can't see any content in the pane that isn't within the viewport.

In the GIF below, the window content moving back and forth is me trying to scroll the stop pane.

aaronbrethorst (approved):
great work 👍

Issue #51, opened 2024-10-17
Load stops for the entire viewport


aaronbrethorst:
# User story

as a transit rider,
I want to see my entire screen covered with stops when I open the web app,
so I can feel confident that my area has a bus stop nearby

Right now, we load stops for the center of the viewport but the size of the viewport can easily be greater than what will be loaded and returned by the API.  We should load stops across the entire viewport.

# Screenshot

<img width="1552" alt="image" src="https://github.com/user-attachments/assets/117b2934-af6c-4672-adb9-e7c1547a1534">

# Concept

I added a circle around the area that is originally loaded. I'd love to see us calculate similarly sized bounding boxes (or circles) that can load stops for the rest of the viewport in the least number of API calls. We should cache these results locally based on the current zoom level so we don't repeat API calls.

![bubbles](https://github.com/user-attachments/assets/2340bd02-db00-441f-984d-b27a1e43f868)

PR #54, opened 2024-10-19
Feat/bounding-box-cache

Ahmedhossamdev:
#51

### DIFF
--- a/src/components/map/MapView.svelte
+++ b/src/components/map/MapView.svelte
@@ -36,11 +36,12 @@
 	let stopsCache = new Map();
 
 	function cacheKey(zoomLevel, boundingBox) {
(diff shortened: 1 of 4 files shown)


aaronbrethorst (changes requested):
I like the way you've structured this. I think it looks good, but I see a lot of cache misses, even when I drag into the same area over and over. 

image

Can you think of ways we could increase the cache hit rate?

Ahmedhossamdev:
@aaronbrethorst 

I think we can set the dight after the point to be fixed to 1 or 2 digits only, and we can increase the radius as well; this will improve the cache hit for sure, and if users want to load more data, they can invalidate the cache by zooming in.

PR #62, opened 2024-10-22
feat: added the responsiveness to the header

tarunsinghofficial:
**Issue:**
Currently, the header is not responsive, and users aren't able to use the header on small screens like mobiles, etc.

**Fixes:** #6 
Added the responsiveness to the header/navbar.

**Screenshot:**
image
image

### DIFF
--- a/src/components/navigation/Header.svelte
+++ b/src/components/navigation/Header.svelte
@@ -6,7 +6,7 @@
 	} from '$env/static/public';
 
 	import ThemeSwitcher from '$lib/ThemeSwitch/ThemeSwitcher.svelte';
-	import { fly } from 'svelte/transition';
+	import MobileMenu from './MobileMenu.svelte';
 
 	let isMobileMenuOpen = false;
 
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
looks good! please rebase off of the latest commits in `main`

src/components/navigation/Header.svelte
```diff
@@ -6,38 +6,113 @@
...
 	</div>
-	<div>
+
+	<div class="hidden md:flex">
 		<ThemeSwitcher />
 	</div>
 </div>
+
+{#if isMobileMenuOpen}
+	<div
```
aaronbrethorst, line 77: this should all be extracted into a new MobileMenu component.

aaronbrethorst (changes requested):
Looks good! Please extract the mobile menu into a new component and I think it'll be ready to merge.

aaronbrethorst (approved):
👍 thanks!

PR #63, opened 2024-10-23
Feat/i18n-language-detection

Ahmedhossamdev:
Added internationalization with the svelte-i18n package.

If this setup is okay, I will continue with localization for different languages.

https://www.npmjs.com/package/svelte-i18n

### DIFF
--- a/src/components/navigation/Header.svelte
+++ b/src/components/navigation/Header.svelte
@@ -6,12 +6,20 @@
 	} from '$env/static/public';
 
 	import ThemeSwitcher from '$lib/ThemeSwitch/ThemeSwitcher.svelte';
-
+	import { t } from 'svelte-i18n';
+	import '$lib/i18n';
 	let headerLinks = null;
 
 	if (PUBLIC_NAV_BAR_LINKS) {
 		headerLinks = JSON.parse(PUBLIC_NAV_BAR_LINKS);
 	}
+
+	const translationKeys = {
+		Home: 'home',
+		About: 'about',
+		Contact: 'contact',
+		'Fares & Tolls': 'fares_and_tolls'
+	};
 </script>
 
 <div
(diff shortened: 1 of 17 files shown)


src/components/navigation/Header.svelte
```diff
@@ -6,12 +6,20 @@
...
-
+	import { t } from 'svelte-i18n';
+	import '$lib/i18n';
 	let headerLinks = null;
 
 	if (PUBLIC_NAV_BAR_LINKS) {
 		headerLinks = JSON.parse(PUBLIC_NAV_BAR_LINKS);
 	}
+
+	const translationKeys = {
```
aaronbrethorst, line 17: we'll need to come up with a different solution for these links in the header. these items are meant to be dynamic and configurable by a transit agency. Right now, they're being driven by a simple JSON string in the .env file, but we'll need to come up with a better solution for making that information localizable. For now, please delete this and leave the home, about, contact, and fares & tolls links unlocalized.

src/components/search/SearchField.svelte
```diff
@@ -70,7 +71,7 @@
 					clip-rule="evenodd"
 				/>
 			</svg>
-			<span class="sr-only">Search</span>
+			<span class="sr-only"> </span>
```
aaronbrethorst, line 74: please put this back, but localized.

aaronbrethorst (changes requested):
this looks great! there are a couple small tweaks, but I'm super impressed with how quickly you tore through this :)

src/components/search/SearchField.svelte
Ahmedhossamdev: oh sorry, I forgot to localize it

src/components/navigation/Header.svelte
Ahmedhossamdev: I think we can adjust the env variables so that every item has a key, value, and route or action

for example 


PUBLIC_NAV_BAR_LINKS='[
  { "key": "home", "value": "Home", "route": "/" },
  { "key": "about", "value": "About", "route": "/about" },
  { "key": "contact", "value": "Contact", "route": "/contact" },
  { "key": "fares_and_tolls", "value": "Fares & Tolls", "route": "/fares-and-tolls." }
]'

aaronbrethorst: The challenge is that the example PUBLIC_NAV_BAR_LINKS I created in the .env file will definitely _not_ match up with what any transit agency decides to put into this navigation bar. There's no way for us to predict what they will put in there.

aaronbrethorst (approved):
🔥 🔥 🔥 🔥  looks great!

PR #69, opened 2024-10-26
Feature/realtime-bus-locations-on-route-select

Ahmedhossamdev:


src/components/map/RouteMap.svelte
```diff
@@ -46,5 +55,7 @@
 				mapProvider.addStopMarker(stop, stopTime);
 			}
 		}
+
+		currentIntervalId = await fetchAndUpdateVehicles(routeId, mapProvider);
```
aaronbrethorst, line 59: what happens if `routeId` is null? I see you're grabbing it from `moreTripData?.routeId` on line 37, which suggests it can be null.

src/components/map/VehiclePopupContent.svelte
```diff
@@ -0,0 +1,47 @@
+<script>
+	export let nextDestination;
+	export let vehicleId;
+	export let lastUpdateTime;
+	export let nextStopName;
+	export let predicted;
+
+	function formatLastUpdated(timestamp) {
```
aaronbrethorst, line 8: this feels like it should be extracted into the `$lib/formatters.js` file, and it will also need to be internationalized/localized.

src/components/map/VehiclePopupContent.svelte
```diff
@@ -0,0 +1,47 @@
...
+		const seconds = secondsAgo % 60;
+
+		if (minutes > 0) {
+			return `${minutes} min${minutes > 1 ? 's' : ''} ${seconds} sec${seconds > 1 ? 's' : ''} ago`;
+		}
+		return `${seconds} sec${seconds > 1 ? 's' : ''} ago`;
+	}
+</script>
+
+// TODO: Add vehicle icon, add the nextStops times
```
aaronbrethorst, line 23: `<!-- TODO: Add vehicle icon, add the nextStops times -->` ;)

image

src/lib/Provider/GoogleMapProvider.js
```diff
@@ -146,6 +152,99 @@ export default class GoogleMapProvider {
...
+			zIndex: 1000
+		});
+
+		this.vehicleMarkers.push(marker);
+
+		const vehicleData = {
+			nextDestination: activeTrip.tripHeadsign,
+			vehicleId: vehicle.vehicleId,
+			lastUpdateTime: vehicle.lastUpdateTime,
+			nextStopName: this.stopsMap.get(vehicle.nextStop)?.name || 'N/A',
```
aaronbrethorst, line 184: is `'N/A'` the most appropriate value for this? It's ok if it is, but it should be i18n'd.

aaronbrethorst (changes requested):
this is looking really good! There are a few changes I'd like to see that I've noted in the code, but they're pretty minor all things considered.

Do you want to take care of the TODOs before this gets merged, or wait til after?

src/components/map/VehiclePopupContent.svelte
Ahmedhossamdev: You are correct; I was unaware of this file.

src/components/map/RouteMap.svelte
Ahmedhossamdev: Since the api requires the routeId, I do not think this can be null, but I did add that it will not draw any buses if there are no trips.


https://developer.onebusaway.org/api/where/elements/trip

Ahmedhossamdev:
Do you want to take care of the TODOs before this gets merged, or wait til after?

I think for me there is no difference between before or after the merge. Just let me know what you prefer <3'

aaronbrethorst (approved):
great work! 👍

PR #70, opened 2024-10-28
Feat/view-all-routes

Ahmedhossamdev:


src/components/oba/StopPane.svelte
```diff
@@ -91,27 +92,7 @@
...
-					<path
-						class="opacity-75"
-						fill="currentColor"
-						d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"
-					></path>
-				</svg>
-				{$t('loading')}...
-			</div>
-		</div>
+		<LoadingSpinner />
```
aaronbrethorst, line 95: 👍

src/routes/api/oba/routes/+server.js
```diff
@@ -0,0 +1,39 @@
+import oba from '$lib/obaSdk.js';
+
+// TODO: change this api after deployment of get all routes api
+export async function GET() {
+	try {
+		const agenciesResponse = await oba.agenciesWithCoverage.list();
```
aaronbrethorst, line 6: it doesn't look like you're caching this data on the server side yet, right? This should be run once per app launch—and then cached in memory, instead of reloaded on every request.

aaronbrethorst (changes requested):
this is looking really good! I have noted a couple behavioral changes I'd like to see before merging it in the PR. Also, here are a few things I noticed when playing around with the feature that should get fixed:

## Content underlap

image

## Desired outcome

The modal pane is smart enough to know what y coordinate it can grow to, and never overlaps the search UI. 

image

## Add a title to the View All Routes modal

It should be called "All Routes", and that text should appear above the search bar.

## The search bar doesn't seem to be working

image


image

src/components/oba/StopPane.svelte
Ahmedhossamdev: Hi @aaronbrethorst 

The search bar is working but before I PR, I also added search by rotueId, I think I will remove it from the search filter so we can search by short and long name and description only

src/hooks.server.js
```diff
@@ -0,0 +1,47 @@
...
+		const routes = await Promise.all(routesPromises);
+		return routes.flat();
+	} catch (error) {
+		console.error('Error fetching routes:', error);
+		return null;
+	}
+}
+
+async function preloadRoutesData() {
+	if (!routesCache) {
```
aaronbrethorst, line 33: 🔥 🔥 🔥

aaronbrethorst (approved):
looks great 🚀

PR #81, opened 2024-10-30
Rebuilds side pane UI to fix underlaps, etc.

### LINKED ISSUE #71: Modal Pane Content Underlap with Search
## Content underlap

image

## Desired outcome

The modal pane is smart enough to know what y coordinate it can grow to, and never overlaps the search UI. 

image

aaronbrethorst:
Fixes #71

The modal pane was feeling a bit wide, and looked a bit strange with its margins not lining up with the search box. This change reconsiders how we render the search box and modals to make them 'physically' smaller and feel more connected to the rest of the UI.

* Get rid of an extra header element on the map — now we only have the single header on the layout page
* Judiciously use `.pointer-events-none` to make large swaths of our UI clickable that had not been before
* Unify/DRY up our modal pane styling so that the search pane and modal panes look similar

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -3,14 +3,6 @@
 @tailwind utilities;
 
 @layer components {
-	.scroll-hidden {
-		-ms-overflow-style: none; /* Internet Explorer 10+ */
-		scrollbar-width: none; /* Firefox */
-	}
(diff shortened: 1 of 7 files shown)


Ahmedhossamdev:
Art <3

Issue #78, opened 2024-10-29
region wide alerts


aaronbrethorst:
This is very similar to the work that just landed in Android: https://github.com/OneBusAway/onebusaway-android/pull/1274

## Feature description

On web app launch, we'll query `${OBACO_API_BASE_URL}/alerts.pb` for new alerts. If a recent, high severity, region-wide alert is found, a modal will be displayed that shows the contents of the alert, and optionally includes a "More Info" button.

## Acceptance criteria 

- [x] Add an optional onebusaway.co API server base URL to the .env (and document in the README). it should include the region ID. e.g. https://onebusaway.co/api/v1/regions/1
- [x] Load the contents of `${OBACO_API_BASE_URL}/alerts.pb` and show the first recent, high severity, system wide alert in a modal dialog

(note: button color is incorrect — I was just too lazy to change it :P)
<img width="1264" alt="image" src="https://github.com/user-attachments/assets/64fbb5e3-a22f-4d3e-9703-ac617d1578a2">

## Example code

```
<script>
	import { Modal } from 'flowbite-svelte';
	let showModal = true;
</script>

<Modal title="All Transit Service Suspended" bind:open={showModal} autoclose>
  <p class="text-base leading-relaxed text-gray-500 dark:text-gray-400">
	King County Metro shut down all its buses. So did Pierce and Snohomish counties. Seattle-Tacoma International Airport briefly closed all three runways, an "unprecedented" situation. Paine Field closed. So did Highway 2 over Stevens
Pass. Garbage pickup was canceled. Museums, theaters, libraries, the zoo, the aquarium: all closed. Doctor appointments
moved online.
  </p>
  <svelte:fragment slot="footer">
	<div class="text-right flex-1">
		<Button color="alternative">Close</Button>
		<Button on:click={() => alert('Handle "success"')}>Learn More</Button>
	</div>
  </svelte:fragment>
</Modal>
```

## (Hopefully) Helpful notes

* This looks like a legit Protobuf library for JS: https://github.com/protobufjs/protobuf.js/
* Flowbite-Svelte has a bunch of great-looking prebuilt UI components we can use, including a modal: https://flowbite-svelte.com/docs/components/modal

Ahmedhossamdev:
Hi @aaronbrethorst 

I found also Js SDK for GTFS-realtime language bindings, they use also the ProtoBuf.js buffer support.

https://gtfs.org/documentation/realtime/language-bindings/nodejs/

PR #83, opened 2024-10-30
Feat/region-wide-alert

Ahmedhossamdev:
fixes: #78

### DIFF
--- a/src/routes/api/oba/alerts/+server.js
+++ b/src/routes/api/oba/alerts/+server.js
@@ -1,9 +1,15 @@
 import GtfsRealtimeBindings from 'gtfs-realtime-bindings';
-import { PRIVATE_OBACO_API_BASE_URL } from '$env/static/private';
+import { PRIVATE_OBACO_API_BASE_URL, PRIVATE_OBACO_SHOW_TEST_ALERTS } from '$env/static/private';
+import { buildURL } from '$lib/urls.js';
 
 export async function GET() {
 	try {
-		const response = await fetch(`${PRIVATE_OBACO_API_BASE_URL}/alerts.pb`);
+		const alertsURL = buildURL(
+			PRIVATE_OBACO_API_BASE_URL,
+			'alerts.pb',
+			PRIVATE_OBACO_SHOW_TEST_ALERTS && PRIVATE_OBACO_SHOW_TEST_ALERTS == 'true' ? { test: 1 } : {}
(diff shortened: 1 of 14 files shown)


src/routes/api/oba/alerts/+server.js
```diff
@@ -0,0 +1,66 @@
...
+		);
+	}
+}
+
+function isValidAlert(alert) {
+	return isAgencyWideAlert(alert) && isStartDateWithin24Hours(alert) && isHighSeverity(alert);
+}
+
+function isHighSeverity(alert) {
+	const isHighSeverity =
```
aaronbrethorst, line 45: let's refactor this with early returns:

```
function isHighSeverity(alert) {
	if (!alert) {
		return false;
	}

	const severity = getSeverityLevel(alert);
	return severity === GtfsRealtimeBindings.transit_realtime.Alert.SeverityLevel.SEVERE ||
		  severity === GtfsRealtimeBindings.transit_realtime.Alert.SeverityLevel.WARNING;
}
```

src/routes/api/oba/alerts/+server.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function isStartDateWithin24Hours(alert) {
+	const startDate = alert.activePeriod[0].start;
+	const now = Date.now() / 1000;
+	return startDate <= now && startDate >= now - 24 * 60 * 60;
+}
+
+function isAgencyWideAlert(alert) {
+	return alert.informedEntity && alert.informedEntity.length > 0;
```
aaronbrethorst, line 65: is this only `true` under the same circumstances as on Android? https://github.com/OneBusAway/onebusaway-android/pull/1274/files#diff-aeef62db59a9689cfc54bcdff19bfd5934dd4846e06c6a47f6d497bb7d25244fR108

aaronbrethorst (changes requested):
great work! I have a couple questions and small tweaks I'd like to see but otherwise I think this is ready to go :)

src/routes/api/oba/alerts/+server.js
Ahmedhossamdev: Yes it's the same as 

`
GtfsRealtimeBindings.transit_realtime.EntitySelector.fromObject(alert.informedEntity[0])
`

We should return true only if the agency ID exists, right?

Issue #101, opened 2024-11-03
Disable Map Rotation on Right-Click Drag in MapLibre Layer for OpenStreetMapProvider


Ahmedhossamdev:
![image](https://github.com/user-attachments/assets/bc3187d0-9424-49c0-b6d8-c9e2be3559a9)

PR #102, opened 2024-11-03
refactor: Disable Map Rotation on Right-Click Drag in MapLibre Layer for OpenStreetMapProvider

Ahmedhossamdev:
fixes: #101


aaronbrethorst (approved):
good catch! 👍

PR #105, opened 2024-11-04
Refactor/stops-icon-type-display

Ahmedhossamdev:
### 1. Enhanced Stop Data Structure
- Merged routes referenced directly with their corresponding route IDs in the stop during API fetching
- Each stop object now includes its associated route data, making it easier to access route details for future features

### 2. Simplified Stop Display Logic
- Removed reactivity for the `showAllStops` property
- All stops now display on the map by default when no route is selected, simplifying the display logic

### 3. Optimized Stops Location Fetch
- Prevented unnecessary stops-for-location fetch calls from running in the background when drawing a route on the map
- This optimization reduces redundant API calls and enhances performance

### DIFF
--- a/src/components/map/MapView.svelte
+++ b/src/components/map/MapView.svelte
@@ -187,9 +187,10 @@
 
 		if (s.routes && s.routes.length > 0) {
 			const routeTypes = new Set(s.routes.map((r) => r.type));
-			const prioritizedType =
-				routePriorities.find((type) => routeTypes.has(type)) || RouteType.UNKNOWN;
-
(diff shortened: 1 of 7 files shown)


aaronbrethorst (changes requested):
The code looks great, but I'm noticing one behavioral issue that is confusing me. Stop ID `40_99610` should be displayed as light rail, but it's showing up as a bus on the map.

image

aaronbrethorst (approved):
looks good! thanks for tracking down those final issues! <3

Issue #115, opened 2024-11-13
Disable Search Input When No Input Provided


Ahmedhossamdev:
Currently, the search input remains enabled even when there is no user input. To improve the user experience and prevent empty searches, we should disable the search input by default and only enable it when there is valid input.

![image](https://github.com/user-attachments/assets/f8aac663-9948-4a16-8383-9c23a7574a5a)

Acceptance Criteria:

- [x] The search button should be disabled when the input field is empty.
- [x] The search button should become enabled only when there is valid input provided.

PR #116, opened 2024-11-13
fix: prevent empty search submissions and disable button accordingly

Ahmedhossamdev:
fixes: #115


aaronbrethorst (approved):
looks good! 👍

PR #136, opened 2024-11-25
Build a homegrown Accordion Component

aaronbrethorst:
- [x] Sticky route schedule headers have a gap above them
- [x] The datepicker appears behind the sticky headers
- [x] On load, the schedules should not animate open — i.e. add an option to open without animation and use it on first page load.
- [x] All schedules should be expanded by default
- [x] The accordion should support showing multiple items at once
- [x] Show all routes/Collapse all routes buttons aren't working


Ahmedhossamdev:
Great work Aaron!

Issue #150, opened 2024-11-28
Wayfinder isn't working on Safari or Mobile Safari


aaronbrethorst:
You can repro this problem at https://pugetsound-wayfinder.onrender.com

If you don't have access to Safari or Mobile Safari for testing, you can _probably_ use Gnome Epiphany to repro the issue: https://apps.gnome.org/Epiphany/

<img width="882" alt="image" src="https://github.com/user-attachments/assets/6b505e24-a029-4779-9897-39c1d1066d76">

aaronbrethorst:
This is working in #152

aaronbrethorst:
Fixed by #152

Issue #140, opened 2024-11-25
Upgrade to Svelte 5


aaronbrethorst:
Let's upgrade to Svelte 5.

PR #152, opened 2024-11-29
Feat: Svelte 5 Migration

Ahmedhossamdev:
Fixes: #140, #150

### Key Changes:
- Used `$state` to define reactive variables instead of `let`.
- Replaced `$:` with `$derived` or `$effect` (Runes support)
- Used `$props` instead of `export let` for component properties.
- Changed event handlers from `on:click` to `onclick` for consistency, along with similar event changes.
- Replaced `slots` with `snippets`.
- Removed deprecated event dispatcher and replaced it by passing the function directly to child components.
- Renamed some functions for better clarity and naming conventions.
- Enhanced `routeClick` with a `flyTo` transaction for smoother navigation.

For more details, refer to the [Svelte v5 Migration Guide](https://svelte.dev/docs/svelte/v5-migration-guide).

- [x] Write a detailed PR description
- [x] Ensure everything works correctly


src/components/MapContainer.svelte
```diff
@@ -1,50 +1,31 @@
+<!-- @migration-task Error while migrating Svelte code: $$props is used together with named props in a way that cannot be automatically migrated. -->
```
aaronbrethorst, line 1: looks like there's some more work to do here.

aaronbrethorst (changes requested):
looks great! Everything looks like it's working perfectly, and you've fixed #150, too!

I see one comment from the migration task that needs to be addressed—either by deleting the comment or doing whatever work it wasn't able to do, and then we can merge.

src/components/MapContainer.svelte
Ahmedhossamdev: No, I forgot to remove the comment. It appeared because the migration script failed to remove $$props, but it has been migrated now to Svelte 5 so there is no errors .

Ahmedhossamdev:
> looks great! Everything looks like it's working perfectly, and you've fixed https://github.com/OneBusAway/wayfinder/issues/150, too!

Haha, I’m glad it got fixed with the migration! I didn’t specifically fix #150, but it’s great to see that everything worked out. 😄 @aaronbrethorst

aaronbrethorst (approved):
terrific work, @Ahmedhossamdev

Issue #147, opened 2024-11-25
 Dark Mode Issue (stop/:id) Page


Ahmedhossamdev:
![image](https://github.com/user-attachments/assets/64e6c443-5f1d-4ec0-9cfd-e3305e06e583)

Ritesh051:
@Ahmedhossamdev whats the main issue there ?

Ahmedhossamdev:
@Ritesh051 

When dark mode is enabled, we expect to see the dark mode enabled in the entire app and pages, with a black color.

Ritesh051:
Ohh it's pretty easy, assign me to this issue.

Ahmedhossamdev:
Done

Ritesh051:
ok Got it .

Ritesh051:
The issue has been fixed. I have raised a Pr.

Ahmedhossamdev:
Hi @Ritesh051,

Thanks for your effort on this PR! Good work, but after reviewing it, it seems the problem has not been fully fixed.
Additionally, we use Svelte 5 syntax, so instead of on:click, we use onclick.

Our dark mode works well, so there's no need to edit the theme switcher. Instead, you should update the colors using Tailwind CSS on the required pages.

Feel free to reach out if you have any questions! 😊

Ritesh051:
ok, I'll fix it.

Issue #146, opened 2024-11-25
Dark Mode Issue (schedule-for-stop) Page


Ahmedhossamdev:
![image](https://github.com/user-attachments/assets/f9d425ba-0d36-43e9-8956-1e0eccf0e2ad)

Issue #145, opened 2024-11-25
Dark Mode Issue (Stop Pane)


Ahmedhossamdev:
![image](https://github.com/user-attachments/assets/6869b45d-8afe-4924-8f3c-572462df7d12)

PR #161, opened 2025-01-08
Fix Dark mode styling for /stop/[stopid]

Kentokamoto:
This addresses the following issues: #145, #146, #147 
Dark mode was not being rendered correctly when enabled. This adds TailwindCSS styling.

The shades of gray and white are not very formulaic but I went with what "looked right". Might be worth discussing some pre-defined colors for both light and dark so we don't continue to choose from every color option available.

Screenshot 2025-01-07 at 20 25 18

Screenshot 2025-01-07 at 20 42 00

### DIFF
--- a/src/components/schedule-for-stop/RouteScheduleTable.svelte
+++ b/src/components/schedule-for-stop/RouteScheduleTable.svelte
@@ -27,34 +27,43 @@
 	}
 </script>
 
-<div class="overflow-x-auto">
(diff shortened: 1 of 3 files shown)


Ahmedhossamdev:
Hey @Kentokamoto, 

Thanks for tackling the dark mode issues! I agree the colors feel a bit random, but we can clean that up later once we figure out the right palette. Awesome work for now, though! 🙌

PR #168, opened 2025-01-17
Primary Color Customization

### LINKED ISSUE #155: Enable Primary Color Customization via Environment Variables
**Story**:

As an agency admin, I want to set a primary color for my agency’s web app so that it aligns with our branding and provides a personalized experience for our users. By being able to define this color through a simple environment variable, I can ensure that our web app reflects our unique identity.

- [x] Add a new environment variable (e.g., APP_PRIMARY_COLOR) to manage the primary color.
- [x] Use a default color if the environment variable is not set.
- [x] Update the codebase to read and apply this color as the primary color across the app.

tarunsinghofficial:
PR Fixes #155 

**Changes Done:**
- Implemented primary color customization throughout the app in major components including buttons, texts, backgrounds, etc using the APP_PRIMARY_COLOR environment variable.
- Changed the background UI of RouteModal and StopPane.

Screenshots:

image
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ See `.env.example` for an example of the required keys and values.
 - `PUBLIC_OBA_REGION_NAME` - string: (required) displayed in the header.
 - `PUBLIC_OBA_LOGO_URL` - string: (required) The URL of your transit agency's logo.
 - `PUBLIC_NAV_BAR_LINKS` - JSON string: (required) A dictionary of the links displayed across the navigation bar.
+- `PUBLIC_APP_PRIMARY_COLOR` - string: (required) The hex color code for the application's primary brand color. Must be wrapped in quotes (e.g., "#214666").
 
 ### OBA Server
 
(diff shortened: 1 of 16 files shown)


aaronbrethorst (changes requested):
Setting `APP_PRIMARY_COLOR` as an environment variable is confusing to me: we don't set any other configuration values via environment variable. 

I could see this making sense if it was possible to set `APP_PRIMARY_COLOR` within the `.env` file but that doesn't seem to work. Please let me know if I'm missing something here, and also please add documentation to the README.

tarunsinghofficial:
> Setting `APP_PRIMARY_COLOR` as an environment variable is confusing to me: we don't set any other configuration values via environment variable.
> 
> I could see this making sense if it was possible to set `APP_PRIMARY_COLOR` within the `.env` file but that doesn't seem to work. Please let me know if I'm missing something here, and also please add documentation to the README.

@aaronbrethorst You're right that setting the primary color via environment variable might seem inconsistent with our other configuration patterns. However, I've found that it does work when properly configured.

It wasn't working earlier because, without the quotes, the color value was being interpreted as a comment, and without dotenv, the environment variables weren't being properly loaded into the Tailwind configuration.

aaronbrethorst (changes requested):
Looking good, but there are a few more changes that need to be done before this can be merged:

* Secondary brand color needs the same treatment as primary color
* Add examples of both to `.env.example`

tarunsinghofficial:
> Looking good, but there are a few more changes that need to be done before this can be merged:
> 
> * Secondary brand color needs the same treatment as primary color
> * Add examples of both to `.env.example`

Done the changes. Ready to 🚀 :)

aaronbrethorst (approved):
looks good, nice work Tarun

Issue #163, opened 2025-01-15
Calendar Widget Positioning/Layout Issue in Route Schedule View


tarunsinghofficial:
## Issue Description
There appears to be a display issue in the route schedules calendar view.

## Steps to Reproduce
1. Navigate to any stop schedule like (1_50468/schedule).
2. Access the schedule view
3. Observe the calendar display

# Expected Behavior
- Calendar should be fully visible within its container and not be hidden due to other components
- All dates of the month should be clearly visible and accessible
- Calendar dates should have proper spacing and layout within the page

**Screenshot**
![image](https://github.com/user-attachments/assets/7ae2e1f5-afe5-4ca9-ab09-db32cc859da6)

tarunsinghofficial:
Hi @aaronbrethorst. The PR #169 fixes this issue, so we can close this now.

PR #169, opened 2025-01-21
UI Improvements in Route Schedule

### LINKED ISSUE #164: Route Schedules View - Responsive Design Issues and UX Improvements
## Issue
There are two main concerns in the Route Schedules view:

1. Responsiveness Issues:
- The time display in the schedule table shows awkward spacing ("02 , 33")
- The layout may not be optimally using the available space

2. UI/UX Concern:
Currently implementing both "Show All Routes" and "Collapse All Routes" as separate buttons, which may not be the most efficient UI pattern.

## Expected Behavior
**Responsiveness:**
- Time/minutes should be properly formatted without extra spacing
- Layout should adapt smoothly to different screen sizes

**UI Controls:**
Recommend switching to a single toggle button that:
- Shows "Show All Routes" when routes are collapsed
- Changes to "Collapse All Routes" when routes are expanded
- Reduces visual clutter
- Makes the interaction more intuitive

**Screenshots:**

image

image

tarunsinghofficial:
Fixes #164, #163 

**Screenshots:**

**Before**
image
image

**After**
image
image

### DIFF
--- a/src/routes/stops/[stopID]/schedule/+page.svelte
+++ b/src/routes/stops/[stopID]/schedule/+page.svelte
@@ -153,7 +153,7 @@
 				</h2>
 
 				<div class="mb-4 flex gap-4">
-					<div class="min-w-32">
+					<div class="z-20 w-[30%] min-w-32">
 						<Datepicker bind:value={selectedDate} inputClass="w-96" />
 					</div>
 
@@ -167,7 +167,7 @@
 				</div>
 
 				<div
-					class="flex-1 rounded-lg border border-gray-200 bg-white p-6 dark:border-gray-700 dark:bg-black"
+					class="flex-1 rounded-lg border border-gray-200 bg-white p-2 dark:border-gray-700 dark:bg-black"
 				>
 					{#if emptySchedules}
 						<p class="text-center text-gray-700 dark:text-gray-400">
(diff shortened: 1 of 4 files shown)


aaronbrethorst (changes requested):
I can't repro #163, so I'm not sure which change you've made that was supposed to fix that. In general, you should strive to submit a fix for a single issue in a single PR. If that doesn't make sense because you're accumulating several tiny fixes, then each commit should fix a single issue. 

# Sticky header 

I see a gap above the sticky header, which is a regression. 

image

# underlined minutes

The new display for minutes is confusing to me because good links on the web are always underlined and these are not links. I am open to the possibility of a new visual treatment, but this isn't going to be it.

# Show all routes button appears to be broken at first

When I first land on the schedule page, all of the routes are expanded, which means that Show All Routes does nothing.

src/routes/stops/[stopID]/schedule/+page.svelte
```diff
@@ -143,22 +152,21 @@
 				</h2>
 
 				<div class="mb-4 flex gap-4">
-					<div class="min-w-32">
+					<div class="z-20 w-[30%] min-w-32">
```
Comment (aaronbrethorst) [inline src/routes/stops/[stopID]/schedule/+page.svelte:155]: this causes a regression on mobile devices.

image

aaronbrethorst (approved):
looks good! I noted one issue below, but I'll fix that in a followup PR.

src/routes/stops/[stopID]/schedule/+page.svelte
Comment (tarunsinghofficial) [reply src/routes/stops/[stopID]/schedule/+page.svelte:155]: Okay. Thanks for solving this :)

Issue #160, opened 2025-01-07
Add basic survey feature support 


aaronbrethorst:
This is the first plan for designing the survey feature in WayFinder.

## Overview
- Two survey types (like in Android):
  1. Popup survey.
  2. Survey triggered when a user clicks on a stop.
- In WayFinder, clicking a stop will show a popup asking if the user wants to take the survey.
- Use browser local storage to prevent the survey from appearing multiple times for the same user.

## Components
- Will use [Flowbite Svelte](https://flowbite-svelte.com/docs/components/forms) for form elements (checkboxes, radios, inputs, etc.).

## References
- [OneBusAway Android Survey Issue #1173](https://github.com/OneBusAway/onebusaway-android/issues/1173)

Ahmedhossamdev:
Could you take a look and let me know if this looks good or if you’d like to change anything? 
@aaronbrethorst

lambaaryan011:
Hi @Ahmedhossamdev are you still working on this issue

Ahmedhossamdev:
> Hi @Ahmedhossamdev are you still working on this issue 

Yep !

PR #170, opened 2025-02-06
Feat/surveys

Ahmedhossamdev:
Fixes: #160

### DIFF
--- a/src/lib/Surveys/surveyUtils.js
+++ b/src/lib/Surveys/surveyUtils.js
@@ -1,0 +1,123 @@
+
+import { showSurveyModal, surveyStore } from '$stores/surveyStore.js';
+
+
+
+export async function loadSurveys(stopId = null, userId = null) {
+    try {
+        const response = await fetch(`/api/oba/surveys?userId=${userId}`);
+        if (!response.ok) throw new Error('Failed to fetch surveys');
+
+        const data = await response.json();
+        const validSurveys = getValidSurveys(data.surveys);
+
+        let selectedSurvey = null;
+
+        if (stopId) {
+            selectedSurvey = getValidStopSurvey(validSurveys, stopId) || getShowSurveyOnAllStops(validSurveys);
+        } else {
+            selectedSurvey = getMapSurvey(validSurveys);
+        }
+
+        surveyStore.set(selectedSurvey);
+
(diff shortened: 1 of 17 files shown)


src/lib/Surveys/surveyUtils.js
```diff
@@ -0,0 +1,121 @@
...
+	return surveys.filter(
+		(survey) =>
+			new Date(survey.end_date) > now &&
+			!localStorage.getItem(`survey_${survey.id}_answered`) &&
+			!localStorage.getItem(`survey_${survey.id}_skipped`)
+	);
+}
+
+export function getValidStopSurvey(surveys, stop) {
+	return (
```
aaronbrethorst, line 39: this function feels too clever to me. I'd prefer to see it written in a more traditional style (i.e. with a for-of loop), instead of this deeply nested quasi-functional style.

src/lib/Surveys/surveyUtils.js
```diff
@@ -0,0 +1,121 @@
...
+	} catch (error) {
+		console.error('Error loading surveys:', error);
+	}
+}
+
+export function getValidSurveys(surveys) {
+	const now = new Date();
+	return surveys.filter(
+		(survey) =>
+			new Date(survey.end_date) > now &&
```
aaronbrethorst, line 32: it's valid for a survey to not have an end date, which makes it not appear on-screen as this filters it out as being invalid.

src/lib/Surveys/surveyUtils.js
```diff
@@ -0,0 +1,121 @@
+import { showSurveyModal, surveyStore } from '$stores/surveyStore.js';
+
```
aaronbrethorst, line 2: Please write a couple basic unit tests for each of the functions in this file. That will help us ensure that there are no regressions going forward, and that we're all on the same page with regard to product behavior.

aaronbrethorst (changes requested):
This looks like a great start! Let me know when you have a chance to work through the changes I mentioned in surveyUtils.js, and I'll continue reviewing!

Ahmedhossamdev:
Hi @aaronbrethorst,

You can continue reviewing now!

aaronbrethorst (approved):
looks good! great work

PR #176, opened 2025-02-16
Add plausible analytics

Ahmedhossamdev:
Fixes: #175


src/routes/+page.svelte
```diff
@@ -190,6 +205,8 @@
 	});
 </script>
 
+<!-- <PlausibleAnalytics  domain={"api.pugetsound.onebusaway.org/stop"} enabled={true}    /> -->
```
aaronbrethorst, line 208: delete me :)

src/routes/api/events/+server.js
```diff
@@ -0,0 +1,47 @@
+export async function POST({ request }) {
+	try {
+		const {
+			domain,
+			name,
+			url,
+			referrer,
+			props,
+			apiHost = 'https://plausible.io'
```
aaronbrethorst, line 9: I'd like to see this design changed slightly: the domain and apiHost can both be pulled directly from the environment variables instead of being passed in from the frontend.

aaronbrethorst (changes requested):
great work! I'd like to see a couple small changes before we merge, but otherwise good to go!

aaronbrethorst (approved):
great work! :fire:

Issue #179, opened 2025-02-19
Add stop/route level service alerts


aaronbrethorst:
# User Story

As a transit rider,
I want to see service alerts displayed on my stop,
so that I know if there is a disruption to my transit service.

# iOS Equivalent UI

<img src="https://github.com/user-attachments/assets/3b9b33f2-daa8-4cdf-a450-c5c6f8a2c8b7" width="320">

# Requirements

* For this API request: https://api.pugetsound.onebusaway.org/api/where/arrivals-and-departures-for-stop/40_99603.json?key=test see the key path: `.data.references.situations` - each of the objects in this array are a Service Alert (which are referred to as `Situation`s for historical and frankly uninteresting reasons inside of OBA).
* For each Service Alert, check the `activeWindows` to see if the alert is currently applicable. Talk to Amr about a bug he's been dealing with on Android around the time unit of these values. Note that the `to` value may be 0, null, or simply missing. * Any of these three values means 'no end date'.
* If the number of applicable Service Alerts > 0, then render a new section in the StopPane called Service Alerts, which displays up to 3 alerts. If the number of alerts > 3, then show a 'Show More' button below the section that will display the rest of the alerts.
* Each alert shows the `summary` as the title and the `description` as the body. The body should be capped at no more than 3 lines of text (Tailwind has a helper utility class to assist with this. Let me know if you need help finding it.)
* Clicking on an alert triggers the appearance of a Modal that shows both the title (i.e. `summary`) and the full body (i.e. `description`).
* Each service alert may be localized into several languages, but let's just ignore this for the time being. Assume that each service alert is only available in a single language for now.

PR #194, opened 2025-02-23
Feat/service-alerts

Ahmedhossamdev:
Fixes: #179


package.json
```diff
@@ -22,6 +22,7 @@
 		"@sveltejs/adapter-node": "^5.2.9",
 		"@sveltejs/kit": "^2.5.27",
 		"@sveltejs/vite-plugin-svelte": "^4.0.0",
+		"@tailwindcss/line-clamp": "^0.4.4",
```
aaronbrethorst, line 25: 👍

src/components/navigation/ModalPane.svelte
```diff
@@ -26,7 +33,7 @@
 				<button
 					type="button"
 					onclick={closePane}
-					use:keybinding={{ code: 'Escape' }}
+					use:keybinding={isModalOpen === false ? { code: 'Escape' } : null}
```
aaronbrethorst, line 36: Can this be simplified to `isModalOpen ? null : {code: 'Escape'}`?

src/components/service-alerts/serviceAlertsHelper.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Normalizes a timestamp value.
+ * If the difference between now and the timestamp is smaller when interpreted as milliseconds,
+ * then it's assumed to be in milliseconds; otherwise, it's in seconds and converted to milliseconds.
+ *
+ * @param {number|null|undefined} time - The timestamp to normalize.
+ * @returns {number} Normalized timestamp in milliseconds.
+ */
+export function normalizeTimestamp(time) {
```
aaronbrethorst, line 21: can you add a couple tests for this function?

aaronbrethorst (changes requested):
Looking good! The one other change I'd like to see made besides the feedback I provided in here is a bug fix for this scenario:

1. Open a stop with a service alert.
2. Click on the first service alert in the list so the modal appears.
3. Press Escape on your keyboard to dismiss the modal.

Expected: the modal closes but the stop pane is still open.
Actual: the stop pane closes too.

src/components/navigation/ModalPane.svelte
Ahmedhossamdev: Yes, I made it like this for the debug only xDD

aaronbrethorst (approved):
great work 🔥

Issue #192, opened 2025-02-22
Don't Show "View Schedule" on the standalone Arrivals and Departures page


aaronbrethorst:
The "View Schedule" button is redundant here. However, it should continue to appear on the map view. 

<img width="1624" alt="Image" src="https://github.com/user-attachments/assets/c68fce68-5d07-4067-ab87-23ca3cbd62ec" />

batwithhat04:
@aaronbrethorst 
Hello Sir,
I'm Avinash Singh, a full stack developer from India, i've huge interest in this project, could you please assign me this issue

PR #195, opened 2025-02-24
feat: conditionally render schedule link in StopPane component

iqbalcodes6602:
added a check on view schedule button to contionally render it only when the tripselected variable is there

fixes: #192


aaronbrethorst (approved):
looks good!

aaronbrethorst (changes requested):
looks good! except that listing failed. please go back and run the linter.

Issue #190, opened 2025-02-22
Search Pane tabs should use branding colors


aaronbrethorst:
The tabs in the SearchPane should use the branding colors, not the orange color shown below.

Look for `PUBLIC_APP_PRIMARY_COLOR` to see examples of how to use it.

# Wrong Colors

<img width="401" alt="Image" src="https://github.com/user-attachments/assets/36059bbb-1bb8-41dc-962d-5ad4b84e2c73" />

# Right Colors

<img width="406" alt="Image" src="https://github.com/user-attachments/assets/eb3f2a03-19d1-49f1-b4e8-469502c2ca52" />

namansh70747:
@aaronbrethorst  is this issue still open can i work on this , thanks

PR #196, opened 2025-02-24
fix: update activeClasses for TabItem components in SearchPane

iqbalcodes6602:
added an activeClass to both the tabs to override the existing primary color to brand

fixes: #190


aaronbrethorst (changes requested):
The changes look fine, but the linter is failing. Also, please squash all three resulting commits into a single one.

iqbalcodes6602:
> The changes look fine, but the linter is failing. Also, please squash all three resulting commits into a single one.

hi @aaronbrethorst  done as said.

aaronbrethorst (changes requested):
I'm not seeing the desired behavior here. The active tab here should be green not orange: 
image

iqbalcodes6602:
> I'm not seeing the desired behavior here. The active tab here should be green not orange: image

hi @aaronbrethorst , 
in my local env it shows green only
image
i dont know whats the reason behind this inconsistency

PR #200, opened 2025-02-28
Improve navbar overflow handling and simplify link width measurement

### LINKED ISSUE #189: Prevent x-axis overflow
There are a few ways we could address this. The simplest is to just apply `overflow-x-hidden` to the navigation bar. A more complicated but 'better' approach would be calculate the width of the navigation bar content vs the size of the viewport and show the hamburger menu when there's overflow.

Image

tarunsinghofficial:
Fixes #189 

**Changes**
- Added the `measureLinksWidth` function to use actual link text for accurate width measurement.
- Ensured responsive behavior remains intact by maintaining accurate width calculations.
- Adjusted the layout and positioning of the burger icon to ensure it is correctly aligned.

Screenshot:
PugetSound-GoogleChrome2025-02-2821-14-38-ezgif com-video-to-gif-converter

### DIFF
--- a/src/components/navigation/Header.svelte
+++ b/src/components/navigation/Header.svelte
@@ -12,7 +12,7 @@
 	let isMobileMenuOpen = $state(false);
 	let shouldShowMobile = $state(false);
 	let navContainer;
-	let linksContainer;
+	let linksContainer = $state(null);
 	let linksWidthCache = 0;
 
 	let headerLinks = $state(null);
--- a/src/components/navigation/Header.svelte
+++ b/src/components/navigation/Header.svelte
@@ -5,10 +5,15 @@
 		PUBLIC_NAV_BAR_LINKS
 	} from '$env/static/public';
 
+	import { onMount } from 'svelte';
 	import ThemeSwitcher from '$lib/ThemeSwitch/ThemeSwitcher.svelte';
 	import MobileMenu from './MobileMenu.svelte';
 
 	let isMobileMenuOpen = $state(false);
+	let shouldShowMobile = $state(false);
+	let navContainer;
+	let linksContainer;
+	let linksWidthCache = 0;
 
 	let headerLinks = $state(null);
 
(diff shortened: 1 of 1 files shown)


src/components/navigation/Header.svelte
```diff
@@ -19,13 +24,80 @@
 	function toggleNavbar() {
 		isMobileMenuOpen = !isMobileMenuOpen;
 	}
+
+	function checkOverflow() {
+		if (!navContainer) return;
+
+		const navWidth = navContainer.clientWidth;
+		const logoWidth = navContainer.querySelector('.logo-container')?.clientWidth || 0;
+		const themeWidth = 48;
```
aaronbrethorst, line 33: I don't understand what `themeWidth` is or why it's set to 48, or where the value of `32` comes from on the next line. Ideally, these two values should be self-explanatory, but if not there should be a comment that explains where the 48 and 32 come from, and why they aren't set to a different value.

src/components/navigation/Header.svelte
```diff
@@ -5,10 +5,15 @@
...
 
+	import { onMount } from 'svelte';
 	import ThemeSwitcher from '$lib/ThemeSwitch/ThemeSwitcher.svelte';
 	import MobileMenu from './MobileMenu.svelte';
 
 	let isMobileMenuOpen = $state(false);
+	let shouldShowMobile = $state(false);
+	let navContainer;
+	let linksContainer = $state(null);
+	let linksWidthCache = 0;
```
aaronbrethorst, line 16: rename this to `cachedLinksWidth`

src/components/navigation/Header.svelte
```diff
@@ -19,13 +24,80 @@
...
+
+		const newShouldShowMobileMenu = linksWidth > availableWidth;
+
+		if (shouldShowMobile !== newShouldShowMobileMenu) {
+			shouldShowMobile = newShouldShowMobileMenu;
+		}
+	}
+
+	function measureLinksWidth() {
+		if (headerLinks && !linksContainer) {
```
aaronbrethorst, line 52: Use an early return instead of deep nesting:

```
if (!headerLinks || linksContainer) {
  return;
}
```

Note: this logic ^^ may be incorrect! You'll need to test and validate!

src/components/navigation/Header.svelte
```diff
@@ -19,13 +24,80 @@
...
+			const tempDiv = document.createElement('div');
+			tempDiv.style.position = 'absolute';
+			tempDiv.style.visibility = 'hidden';
+			tempDiv.style.display = 'flex';
+			tempDiv.style.gap = '1rem';
```
aaronbrethorst, line 57: where does the `gap` here and the `padding` on line 62 come from? Are these derived from the current style values on the page? If so, those should be reused somehow in case the styling changes later.

aaronbrethorst (changes requested):
looking good! The behavior feels great. I have some nitpicks about the code I'd like to see addressed before merging, but hopefully they should all be quick.

src/components/navigation/Header.svelte
tarunsinghofficial: The `themeWidth` is the value for our theme toggler switch (now defined as a constant - ), and `32` is the extra padding/margin we need to calculate, when we want to trigger our burger icon for small screens. I have made the following changes based on your suggestions:

- I have now defined THEME_SWITCHER_WIDTH as 56 (the width of toggler) and EXTRA_PADDING (32 as calculated for extra padding) as constants and added comments to explain their purpose.

src/components/navigation/Header.svelte
tarunsinghofficial: Done

src/components/navigation/Header.svelte
tarunsinghofficial: Yes, correct. I have updated the function to use an early return to avoid deep nesting and make the code more readable.

src/components/navigation/Header.svelte
tarunsinghofficial: The `gap` and `padding` are derived from actual style values used in the page here (`<a href={value} class="block px-2 py-1 font-semibold text-gray-900 dark:text-white"
									>{key}</a
								>`).
								
For ease, I've now defined constants in the file with clear comments explaining which Tailwind classes they correspond to.

Note: If the styles change later, we just have to update these constants according to the corresponding Tailwind classes from the template. That's the best & easiest approach to implement, as I think :)

aaronbrethorst (approved):
👍

Issue #193, opened 2025-02-23
Improve the Route Search PopupContent experience


aaronbrethorst:
Repro steps:

1. Click on `Click here for a list of available routes`
2. Click on the first item in the "All Routes" pane
3. Click on the first stop listed in the Route pane

Expected:

The map zooms in on the relevant stop _and_ a popup automatically appears with more information.

Actual:

Only the map zoom occurs. I have to click on the (unlabeled) stop to learn more about it.

# Acceptance Criteria

- [ ] After the map zoom occurs, the popup for the selected stop automatically appears (see OpenStreetMapProvider.svelte.js around line 156-179 to see how the click event is managed today)
- [ ] The PopupContent component has a new "View Stop Information" button that displays the stop information in a side pane (i.e. the normal experience)
- [ ] The feature works identically between both Google Maps and OSM

<img width="948" alt="Image" src="https://github.com/user-attachments/assets/3ce26462-0b45-42fe-9e5e-152494261cd0" />

aaronbrethorst:
@bencodeorg is going to look at this.

tarunsinghofficial:
Can I work on this? @aaronbrethorst

bencodeorg:
I have something close to done for this...busy tonight but should to clean up and put out a PR tomorrow. If it's urgent though feel free to go for it!

tarunsinghofficial:
> I have something close to done for this...busy tonight but should to clean up and put out a PR tomorrow. If it's urgent though feel free to go for it!

No problem @bencodeorg. You can submit the PR :) THanks.

Rani-Boora:
Hello everyone 👋  
I'm Rani, a beginner in Open Source.  
I want to contribute to Wayfinder under GSoC 2025, but all the issues seem assigned already 😔.  

Could anyone please guide me if there are any small tasks or new issues I can take up to start contributing?  
Thank you so much for your time! 😊

batwithhat04:
@aaronbrethorst claim

Rani-Boora:
Thank you @aaronbrethorst for assigning me this issue.
I'm working on it!

Rani-Boora:
Oops! Sorry for the confusion 
I misunderstood the assignment process.
I'm still very much interested in contributing to this issue.
Could you please assign this to me @aaronbrethorst?

bencodeorg:
Alrighty, sorry for the delay -- PR here! https://github.com/OneBusAway/wayfinder/pull/201

[...]

PR #201, opened 2025-03-01
Open stop information on click in route modal, add button to open stop modal

bencodeorg:
Resolves https://github.com/OneBusAway/wayfinder/issues/193 -- makes some improvements to the experience when a stop is selected within the Route Modal, namely:
- shows the popup on the map for the stop you've selected
- adds a button to that popup to allow you to open the stop modal from the popup itself.

### DIFF
--- a/src/components/MapContainer.svelte
+++ b/src/components/MapContainer.svelte
@@ -9,12 +9,12 @@
 	import { onMount } from 'svelte';
 	import { MapSource } from './../config/mapSource.js';
 
-	let apiKey = env.GOOGLE_MAPS_API_KEY;
+	let apiKey = env.PUBLIC_OBA_GOOGLE_MAPS_API_KEY;
 	let { handleStopMarkerSelect, mapProvider = $bindable(), ...restProps } = $props();
 
 	onMount(() => {
 		if (PUBLIC_OBA_MAP_PROVIDER === MapSource.Google) {
-			mapProvider = new GoogleMapProvider(apiKey);
+			mapProvider = new GoogleMapProvider(apiKey, handleStopMarkerSelect);
 		} else if (PUBLIC_OBA_MAP_PROVIDER === MapSource.OpenStreetMap) {
 			mapProvider = new OpenStreetMapProvider(handleStopMarkerSelect);
 		} else {
--- a/src/components/MapContainer.svelte
+++ b/src/components/MapContainer.svelte
@@ -16,7 +16,7 @@
 		if (PUBLIC_OBA_MAP_PROVIDER === MapSource.Google) {
 			mapProvider = new GoogleMapProvider(apiKey);
 		} else if (PUBLIC_OBA_MAP_PROVIDER === MapSource.OpenStreetMap) {
-			mapProvider = new OpenStreetMapProvider(apiKey);
+			mapProvider = new OpenStreetMapProvider(handleStopMarkerSelect);
 		} else {
 			console.error('Unknown map provider:');
 		}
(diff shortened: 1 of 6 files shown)


bencodeorg:
I was say that I didn't understand the third requirement for this ("The feature works identically between both Google Maps and OSM") because I didn't see an equivalent feature in Google Maps, but I'm now realizing it meant when using Google Maps as the base map for OBA 🤦 . Let me fix that...

src/components/MapContainer.svelte
```diff
@@ -9,14 +9,14 @@
 	import { onMount } from 'svelte';
 	import { MapSource } from './../config/mapSource.js';
 
-	let apiKey = env.GOOGLE_MAPS_API_KEY;
+	let apiKey = env.PUBLIC_OBA_GOOGLE_MAPS_API_KEY;
```
bencodeorg, line 12: Is this a good change? Couldn't get Google Maps to work without it, unless I'm putting my key in the wrong place.

src/lib/Provider/GoogleMapProvider.svelte.js
```diff
@@ -122,33 +123,37 @@ export default class GoogleMapProvider {
...
-		marker.addListener('click', () => {
-			if (this.globalInfoWindow) {
-				this.globalInfoWindow.close();
-			}
+		marker.addListener('click', () => this.openStopMarker(stop));
 
-			if (this.popupContentComponent) {
-				unmount(this.popupContentComponent);
-			}
+		this.markersMap.set(stop.id, marker);
```
bencodeorg, line 128: Not 100% sure on this approach / if it's ok to add to this `markersMap` set, but I needed the Marker in order to be able to show it after clicking on the stop in the Route modal. LMK if this seems reasonable.

src/components/map/PopupContent.svelte
```diff
@@ -1,10 +1,11 @@
 <script>
-	let { stopName, arrivalTime } = $props();
+	import { t } from 'svelte-i18n';
+	let { stopName, arrivalTime, handleStopMarkerSelect } = $props();
 </script>
 
 <div class="transform rounded-lg bg-white p-4 shadow-lg">
 	<h3 class="text-xl font-bold text-gray-900">{stopName}</h3>
-	<p class="mt-2 flex items-center text-gray-700">
+	<p class="my-4 flex items-center text-gray-700">
```
bencodeorg, line 8: I think this popup could use a styling pass, but this at least made the Google version a bit less scrunched together.

src/components/MapContainer.svelte
```diff
@@ -9,14 +9,14 @@
...
 	let { handleStopMarkerSelect, mapProvider = $bindable(), ...restProps } = $props();
 
 	onMount(() => {
 		if (PUBLIC_OBA_MAP_PROVIDER === MapSource.Google) {
-			mapProvider = new GoogleMapProvider(apiKey);
+			mapProvider = new GoogleMapProvider(apiKey, handleStopMarkerSelect);
```
bencodeorg, line 17: Felt slightly unholy to be passing this function to the map providers, but not sure how else to get the Stop Modal to show up.

bencodeorg:
Meant to add, this should be working in Google Maps as the base layer and is ready for review! cc @aaronbrethorst

src/components/MapContainer.svelte
aaronbrethorst: Yep, I think that you actually fixed a bug in our google maps integration here. ✅

src/lib/Provider/GoogleMapProvider.svelte.js
aaronbrethorst: I think this looks totally reasonable.

src/components/MapContainer.svelte
aaronbrethorst: This looks pretty idiomatic to me, actually. 👍

aaronbrethorst (changes requested):
looks good! Please run `npm run lint` and `npm run format` over it to satisfy the linter and it'll be ready to merge.

aaronbrethorst (approved):
looks good, nice work! 👍

PR #225, opened 2025-03-26
Manage modals in single bit of state

bencodeorg:
A proposed organizational improvement to managing the modal in the side panel -- namely, using a single bit of state to choose which modal to show, rather than a series of booleans.

### DIFF
--- a/src/components/search/SearchPane.svelte
+++ b/src/components/search/SearchPane.svelte
@@ -108,7 +108,6 @@
 		window.dispatchEvent(event);
 	}
 
-	// only gets called when going trip planner -> stops and stations, not vice versa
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Definitely looks like a major improvement. I'm seeing one blocker preventing merging: trying to plan a trip doesn't work—the modal doesn't pop up.

aaronbrethorst:
@bencodeorg I slacked you the necessary info

bencodeorg:
Alrighty @aaronbrethorst, should be sorted now (was a typo, switched to use a constant to manage the modal states rather than individual strings to hopefully avoid in the future).

Issue #232, opened 2025-03-28
Race Condition in Stop Fetching Due to Rapid Map Zooming/Panning


Sundaresan-Karunakaran:
When users rapidly zoom in and out or pan the map quickly, multiple requests to fetch stops are triggered in quick succession. This can lead to a race condition where older requests complete after newer ones, resulting in outdated or inconsistent stop data being displayed on the map. More importantly, it also causes inefficiencies, as unnecessary requests are processed for data that is no longer relevant or useful.

Steps to Reproduce

1. Open the map interface.
2. Quickly zoom in and out multiple times or pan rapidly.
3. Observe the network requests in the developer tools (many overlapping requests are sent).
4. Sometimes, older data appears instead of the latest fetched stops.

Expected Behavior

- Only the most recent stop data should be displayed after zooming/panning.
- Older requests should be properly canceled when a new request is initiated.

Additional Info
This issue is similar to the issue #220 .

Sanjai-Shaarugesh:
Issue #232 Race Condition in Stop Fetching Due to Rapid Map Zooming/Panning and Enhanced Stop Details with Interactive Map 🗺️ #238

PR #233, opened 2025-03-28
Fix race condition when fetching stops on rapid zoom/pan

Sundaresan-Karunakaran:
This PR fixes the issue #232  where rapidly zooming/panning caused a race condition, leading to incorrect outputs and inefficiency.

Problem

When users rapidly zoom in and out or pan the map, multiple requests to fetch stops are triggered in quick succession. This can lead to:

- A race condition where older requests resolve after newer ones, causing outdated or inconsistent data to be displayed.
- Inefficiencies due to unnecessary network requests fetching data that is no longer needed.

Solution
- Implemented an AbortController to cancel previous fetch requests before making a new one.
- Moved the cancellation logic to loadStopsAndAddMarkers to ensure every request is controlled properly.
- Ensured only the latest valid response updates the stop data.


Sundaresan-Karunakaran:
@aaronbrethorst 
Apologies, I attempted to squash the commits before submitting, but I couldn't find the option here as I did in my forked repo. Next time, I'll handle it from the CLI directly to make it easier.

Sorry for any confusion!

Sundaresan-Karunakaran:
Hi @aaronbrethorst, I saw that the PR was closed, but I didn’t get any feedback before that. I just want to understand if there’s anything I could improve or if it was closed for another reason. Appreciate your time!

PR #238, opened 2025-04-03
Issue 232

Sanjai-Shaarugesh:
#232 

## Race Condition in Stop Fetching Due to Rapid Map Zooming/Panning

- **Store the vehicle data** with the vehicle ID as the key 🏇  
- **Use `requestAnimationFrame`** for smoother animations  
- **Use `requestAnimationFrame`** for smoother zooming  
- **Pre-create container** for InfoWindows to reduce DOM operations  
- **Preload geometry library** for polylines  
- **Use a single listener** with state tracking for better performance  
- **Optimize positioning** with hardware-accelerated transforms  
- **Use closure** to store `stopTime` to avoid recreating functions  
- **Cache SVG icons** for vehicles for better performance  
- **Create cache key** based on orientation and color  
- **Store vehicle data** directly on marker to avoid additional lookups  
- **Create popup on demand** instead of pre-creating for all markers  
- **Smoother animation** using `panTo` first, then zoom  
- **Optimized stop loading** with adaptive loading strategy

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,1237 @@
+{
+  "lockfileVersion": 1,
+  "workspaces": {
+    "": {
+      "name": "onebusaway-webapp",
+      "dependencies": {
+        "@fortawesome/free-regular-svg-icons": "^6.6.0",
+        "@googlemaps/js-api-loader": "^1.16.8",
+        "@maplibre/maplibre-gl-leaflet": "^0.0.22",
+        "@tailwindcss/forms": "^0.5.9",
+        "dotenv": "^16.4.7",
+        "gtfs-realtime-bindings": "^1.1.1",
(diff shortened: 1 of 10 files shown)


Sanjai-Shaarugesh:
@aaronbrethorst 

## Is this bro or not enough bro 🥦 🦕

aaronbrethorst (changes requested):
I like the idea of the embedded map on the standalone stop page, but this pull request is not ready to merge:

* Tests are failing
* linter is failing
* This bun.lock file needs to be deleted
* Get rid of the extraneous whitespace you introduced
* get rid of `src/app.d.ts`
* get rid of the notional performance enhancements
* learn how the project uses a plugin map architecture and _reuse_ that on the standalone stop page

PR #241, opened 2025-08-30
hotfix/converted-string-to-uppercase-in-convertUnixToTime

ANCHITAYUSH:
This PR fixes the issue where the convertUnixToTime function returned lowercase "am" / "pm" instead of uppercase "AM" / "PM".
The test was failing because of case-sensitive string comparison.

- Before: 01:00 pm
- After: 01:00 PM

**Changes Made**
`localDateTime.replace(/am|pm/, time => time.toUpperCase());`

**Fix**
Screenshot 2025-08-31 011844

**ISSUE**: [issue-52](https://github.com/OneBusAway/waystation/issues/52)

### DIFF
--- a/src/lib/formatters.js
+++ b/src/lib/formatters.js
@@ -24,5 +24,7 @@ export function convertUnixToTime(seconds) {
 	if (!seconds) return '';
 	const date = new Date(seconds * 1000);
 	const utcDate = new Date(date.toUTCString().slice(0, -4));
-	return utcDate.toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
The test passes on an `en-US` locale. The correct fix won't involve a regex to the application code, but making the test mock the locale.

ANCHITAYUSH:
I’ve updated the test to mock the locale so that toLocaleTimeString consistently returns results in en-US format with uppercase AM/PM. This ensures the test passes deterministically across environments without modifying the application logic.

Issue #246, opened 2025-09-22
Sort the list of available routes in numerical order by route_id:


aaronbrethorst:
<img width="386" height="599" alt="Image" src="https://github.com/user-attachments/assets/6d513890-51b6-41d4-a637-d2a8e04880e5" />

PR #260, opened 2025-09-25
refactor: sort routes by id

Ahmedhossamdev:
Fix: #246


aaronbrethorst (approved):
looks good, nice work!

I'm creating [a followup task](#268) for someone to do at some point to extract out this function into a helper, but it shouldn't hold up merging the change.

Issue #266, opened 2025-09-27
When we search for a route and click on a stop and click on View stop information button it display the stops on the map with route which is wrong state


Ahmedhossamdev:
<img width="1912" height="932" alt="Image" src="https://github.com/user-attachments/assets/f1119f98-7b04-4bcd-93a7-8a595d418964" />

Ahmedhossamdev:
The root issue is that we're using the same handleStopMarkerSelect for stops in route context and stops in the map and this is wrong, I will add a PR soon

PR #273, opened 2025-10-03
Refactor/map-state-simplification

Ahmedhossamdev:
Fix: #266

- Add Modes enum for cleaner state transitions
- Centralize polyline cleanup in map providers
- Support smooth marker highlighting during mode changes
- Improve route context preservation
- Add automatic transit icon selection by route type


aaronbrethorst (approved):
great work 🔥

PR #324, opened 2026-02-03
fix: prevent race condition in trip planner location selection

### LINKED ISSUE #323: Bug: Race condition when rapidly selecting locations in Trip Planner
## Description
When quickly clicking on multiple location suggestions in the Trip Planner, a race condition occur resulting in duplicate markers, mixed state, or incorrect location being selected.

## Steps to Reproduce
1. Navigate to the app and open the **"Plan Trip"** tab
2. Open **DevTools → Network tab** and throttle to **"Slow 3G"** (makes the bug easier to trigger)
3. Type a location in the "From" field (e.g., "Seattle")
4. When suggestions appear, **rapidly click on two different suggestions** in quick succession
5. Observe the results

## Expected Behavior
- Only the **last clicked** location should be selected
- Only **one marker** should appear on the map
- The input field should show the **correct location name**

## Actual Behavior
- Multiple markers may appear on the map
- The input field may show a different location than the marker
- State becomes inconsistent between the UI and map

## Environment
- Browser: Chrome
- Network: Any (easier to reproduce with slow network)

## Status
Working now :)

tarunsinghofficial:
fixes #323 

Replaced the simple lock boolean with request ID tracking to properly handle rapid location selections. Now only the most recent selection is applied, preventing duplicate markers and mixed state.

### DIFF
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -28,7 +28,8 @@
 	let fromMarker;
 	let toMarker;
 	let loading = $state(false);
-	let lockSelectLocation = false;
+	let fromRequestId = 0;
+	let toRequestId = 0;
 
 	async function fetchAutocompleteResults(query) {
 		const response = await fetch(`/api/oba/place-suggestions?query=${encodeURIComponent(query)}`);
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Tarun! Great catch on this race condition - the request ID tracking pattern you've implemented is exactly the right approach for handling concurrent async operations. The code is well-structured and the logic is sound. Tests pass and lint is clean.

---

## What's Good

The implementation is solid:

- **Correct request ID pattern**: Incrementing and capturing the ID atomically, then comparing after the async call is textbook race condition handling
- **Separate counters for from/to**: Allows independent concurrent requests without cross-field interference
- **Immediate UI feedback**: Clearing results before the async call provides snappy UX
- **Marker cleanup**: Properly removes old markers before adding new ones to prevent duplicates

The fix correctly addresses issue #323.

---

## Suggestions (Non-blocking)

These are improvements to consider, but don't block the merge:

### 1. User feedback on geocoding failure

**File:** `src/components/trip-planner/TripPlan.svelte:119-121`

Currently, if geocoding fails, the user sees nothing happen - the catch block only logs:

```javascript
} catch (error) {
    console.error('Error selecting location:', error);
}
```

Consider clearing the place name to indicate failure:

```javascript
} catch (error) {
    console.error('Error selecting location:', error);
    if (isFrom) {
        fromPlace = '';
        selectedFrom = null;
    } else {
        toPlace = '';
        selectedTo = null;
    }
    // Future: show toast notification
}
```

### 2. Check `addPinMarker` return value

**File:** `src/components/trip-planner/TripPlan.svelte:109, 116`

`addPinMarker` can return `null` if the map isn't initialized (see OpenStreetMapProvider). Currently this isn't checked:

```javascript
fromMarker = mapProvider.addPinMarker(selectedFrom, $t('trip-planner.from'));
// Could be null
```

If you want extra robustness:

```javascript
const newMarker = mapProvider.addPinMarker(selectedFrom, $t('trip-planner.from'));
if (newMarker) {
    if (fromMarker) {
        mapProvider.removePinMarker(fromMarker);
    }
    fromMarker = newMarker;
    // ... rest
}
```

### 3. Consider marker operation order

Currently: remove old marker → add new marker. If `addPinMarker` fails after removal, you're left with no marker.

Alternative: add new marker → remove old marker. This ensures a marker is always visible, though it briefly shows two markers.

This is a minor edge case and the current approach is fine for most scenarios.

---

The race condition fix is well-implemented and the suggestions above are polish items that can be addressed in a follow-up if desired.

Nice work!

PR #337, opened 2026-02-11
feat: add reverse/swap trip button to swap origin and destination

### LINKED ISSUE #325: Feat: Add "Reverse Trip" button to swap origin and destination
## Description
Add a quick "swap/reverse" button to the Trip Planner that swaps the From and To locations, allowing users to easily plan return trips.

## Use Case
- User plans trip from "Home → Work"
- At end of workday, user wants to go "Work → Home"
- Instead of re-typing both locations, user clicks the reverse button

## Acceptance Criteria
- [ ] Add a swap button (↔️ icon) between the From and To fields
- [ ] Clicking the button swaps:
  - From/To text values
  - Selected coordinates
  - Map markers (if present)
- [ ] If a trip was already planned, optionally re-run the search with swapped locations
- [ ] Button is disabled if both fields are empty

## UI Reference
Similar to Google Maps trip planner swap button.

## Working?
Yes

tarunsinghofficial:
Fixes #325

### DIFF
--- a/src/lib/tripPlanUtils.js
+++ b/src/lib/tripPlanUtils.js
@@ -46,7 +46,9 @@ export function swapTripLocations({
 	const newFromMarker = newSelectedFrom
 		? mapProvider.addPinMarker(newSelectedFrom, t('trip-planner.from'))
 		: null;
-	const newToMarker = newSelectedTo ? mapProvider.addPinMarker(newSelectedTo, t('trip-planner.to')) : null;
+	const newToMarker = newSelectedTo
+		? mapProvider.addPinMarker(newSelectedTo, t('trip-planner.to'))
+		: null;
 
 	return {
 		fromPlace: newFromPlace,


aaronbrethorst (changes requested):
Hey Tarun, thanks for adding the swap button to the trip planner — this is a feature users have been asking for, and the core idea of swapping text, coordinates, and markers is exactly right. Before we can merge this, I will need you to make a couple changes:

## Critical

### 1. Branch needs a rebase — the race condition fix from PR #324 has been reverted

Your branch was forked before PR #324 was merged into `develop`, so the `selectLocation()` function has reverted from the `fromRequestId`/`toRequestId` stale-request tracking back to the old `lockSelectLocation` boolean pattern. This was the exact bug that PR #324 fixed — rapid location selections could cause duplicate markers and mixed state.

`git diff origin/develop` confirms that your branch removes `fromRequestId`/`toRequestId` and replaces them with `lockSelectLocation`. Please rebase onto current `develop` to pick up that fix. The `swapLocations()` function and layout changes should rebase cleanly since they touch different parts of the file.

If you haven't done a rebase before or would like a refresher, I wrote a blog post to help demystify the process: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

### 2. Lint failures in two PR files

- **`src/components/trip-planner/__tests__/TripPlanSwap.test.js`** line 5: trailing whitespace in the JSDoc comment (` * ` should be ` *`)
- **`src/locales/en.json`**: missing trailing newline at end of file

Running `npx prettier --write src/components/trip-planner/__tests__/TripPlanSwap.test.js src/locales/en.json` should fix both.

## Important

### 3. Tests don't exercise the actual component — they duplicate the swap logic inline

Every test in `TripPlanSwap.test.js` re-implements the swap logic with local variables rather than testing the actual `swapLocations()` function or the rendered `TripPlan` component. For example:

```javascript
it('should swap text values when both are populated', () => {
    let fromPlace = 'Capitol Hill';
    let toPlace = 'University District';
    const tempPlace = fromPlace;
    fromPlace = toPlace;
    toPlace = tempPlace;
    expect(fromPlace).toBe('University District');
});
```

This tests that JavaScript variable assignment works, not that the component's swap function works. If a bug were introduced in the actual `swapLocations()` — or even if the function were deleted entirely — all 11 tests would still pass.

Please restructure the tests to either:
- **(a)** Render `TripPlan.svelte` with mocked dependencies, click the swap button, and verify the fields swap (following the pattern in the sibling `TripPlanSearchField.test.js`), or
- **(b)** Extract the swap logic into a standalone utility function (e.g., in `$lib/tripPlanUtils.js`) that the tests can import and call directly.

### 4. Marker swap logic has a stale reference bug and dead code

In `swapLocations()` (lines 133-155), the marker section:
1. Removes both markers from the map
2. Swaps the marker references (temp variable pattern)
3. Overwrites both references by creating new markers

The swap in step 2 is dead code when both coordinates exist, because step 3 overwrites both variables. When only one coordinate exists, the other variable retains a stale reference to a removed marker instead of being set to `null`. This won't crash (both map providers' `removePinMarker` is null-safe and idempotent), but it's semantically incorrect and confusing.

Simplify the entire marker section to:

```javascript
// Remove existing markers (both providers are null-safe)
if (fromMarker) mapProvider.removePinMarker(fromMarker);
if (toMarker) mapProvider.removePinMarker(toMarker);

// Re-add with swapped positions (selectedFrom/selectedTo already swapped above)
fromMarker = selectedFrom
    ? mapProvider.addPinMarker(selectedFrom, $t('trip-planner.from'))
    : null;
toMarker = selectedTo
    ? mapProvider.addPinMarker(selectedTo, $t('trip-planner.to'))
    : null;
```

This eliminates the intermediate swap, removes the dead code, and explicitly nulls out stale references.

## Fit and Finish

### 5. Stale HTML comment

The comment on line 245 still reads `<!-- From/To fields: Mobile grid layout (labels left, fields right) | sm+: stacked vertical -->` but the layout has been changed from a CSS grid to flexbox, and the mobile labels now stack above their fields rather than appearing to their left. Please update the comment to match the new layout.

### 6. `sm:space-y-4` on the outer flex container

On line 246, the outer container has `sm:space-y-4` carried over from the old grid layout. In the new flex-row context, this adds `margin-top` to the swap button div on sm+ screens, which may cause a slight vertical misalignment. Consider whether `gap-y-4` on the inner flex-col is sufficient and this class can be removed.

tarunsinghofficial:
Hi @aaronbrethorst 

I have successfully addressed all 6 issues for the trip planner swap button, and also the branch has been rebased onto `upstream/develop`

PR #349, opened 2026-02-23
Add followup improvements from Analytics #332 PR

### LINKED ISSUE #333: Analytics: follow-up improvements from PR #332 review
Follow-up items identified during code review of #332. All are nice-to-have improvements — none are blocking.

### 1. Validation ordering in `forwardEvent`

`isEnabled()` runs before the `name`/`url` validation in `forwardEvent()`, so missing fields are silently swallowed when analytics is disabled. Consider swapping the order or adding a comment noting this is intentional.

**File:** `src/lib/Analytics/PlausibleAnalytics.js`

### 2. Singleton comment in server endpoint

The module-level `new PlausibleAnalytics()` in `+server.js` captures the `dynamicEnv` reference once. This works because SvelteKit's dynamic env is a proxy with live property access, but a brief comment would clarify this for future readers (e.g., `// Singleton: env values are read dynamically on each call`).

### 3. Test for `forwardEvent()` called with no arguments

The test suite covers missing `name` and missing `url` separately but does not test `forwardEvent()` with no argument at all (which would throw a destructuring error). A test would document the expected behavior for this edge case.

**File:** `src/tests/lib/PlausibleAnalytics.test.js`

### 4. Consolidate repeated `consoleSpy` boilerplate

Five tests in `PlausibleAnalytics.test.js` follow an identical pattern of creating a `consoleSpy`, asserting on it, and restoring it. These could be grouped under a `describe('isEnabled guards')` block with shared `beforeEach`/`afterEach` to reduce per-test boilerplate.

**File:** `src/tests/lib/PlausibleAnalytics.test.js`

### 5. Use path alias in `events.test.js` import

The import uses a relative path (`../../routes/api/events/+server.js`) instead of the project's `$src` alias. Could use `$src/routes/api/events/+server.js` for consistency with project conventions documented in `CLAUDE.md`.

**File:** `src/tests/api/events.test.js`

Bikram-ghuku:
Fixes #333 

Added the suggested changes from the issue

### DIFF
--- a/src/lib/Analytics/PlausibleAnalytics.js
+++ b/src/lib/Analytics/PlausibleAnalytics.js
@@ -23,6 +23,8 @@ export class PlausibleAnalytics {
 	}
 
 	async forwardEvent({ name, url, referrer, props }) {
+		// Check if enabled first - when disabled, we skip parameter validation
+		// to avoid throwing errors during development/testing with analytics off
 		if (!this.isEnabled()) {
 			return { status: 'analytics disabled' };
 		}
--- a/src/routes/api/events/+server.js
+++ b/src/routes/api/events/+server.js
@@ -1,5 +1,6 @@
 import { PlausibleAnalytics } from '$lib/Analytics/PlausibleAnalytics.js';
 
+// Singleton: env values are read dynamically on each call via SvelteKit's env proxy
 const analytics = new PlausibleAnalytics();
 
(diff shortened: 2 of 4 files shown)


Bikram-ghuku:
Hi @aaronbrethorst. This PR is ready for review. This isn't a priority since it doesn't fix blocking issues.

PR #359, opened 2026-02-25
Fix/server cache timeouts

suhr25:
## Summary

This PR fixes a critical issue where **every incoming SvelteKit server request was blocked on upstream OBA/OTP API calls** during cache preload.

The root cause was the global `handle()` hook awaiting cache initialization:

```js
export async function handle({ event, resolve }) {
    await Promise.all([preloadRoutesData(), preloadOtpVersion()]);
    return resolve(event);
}
```

Since both `preloadRoutesData()` and `preloadOtpVersion()` performed **network calls without timeouts**, any slow or unresponsive upstream (OBA / OTP) caused **all requests to hang indefinitely**.

Additionally:

* Cache refreshes (TTL expiry) triggered the same blocking behavior
* Failures caused immediate retries with no cooldown
* No stale-cache serving → users were fully blocked even when old data existed

This resulted in **site-wide stalls, connection buildup, and potential memory/FD growth under load**.

---

## Fix

This PR introduces three key improvements:

### 1. Bounded upstream calls with timeouts

All external fetches are now wrapped with timeouts using `AbortController` or a safe promise race:

```js
function withTimeout(promise, ms, label) {
    let timeoutId;
    const timeoutPromise = new Promise((_, reject) => {
        timeoutId = setTimeout(() => reject(new Error(`${label} timed out`)), ms);
    });
    return Promise.race([promise, timeoutPromise])
        .finally(() => clearTimeout(timeoutId));
}
```

And for OTP:

```js
const ac = new AbortController();
const timer = setTimeout(() => ac.abort(), DETECT_TIMEOUT);

try {
    const response = await fetch(PUBLIC_OTP_SERVER_URL, { signal: ac.signal });
} finally {
    clearTimeout(timer);
}
```

This ensures **no request can hang indefinitely**.

---

### 2. Stale-while-revalidate (non-blocking cache refresh)

Cache logic was reworked so that:

* If cache exists but is stale → **serve immediately**
* Refresh runs in background
* Only cold start waits (and now bounded by timeout)

```js
if (routesCache && !forceRefresh) {
    return; // serve stale data immediately
}
```

This removes the **global blocking window during TTL refreshes**.

---

### 3. Failure cooldown to prevent retry storms

A retry guard was added to avoid hammering OBA after failures:

```js
if (!routesCache && lastErrorTime !== null &&
    now - lastErrorTime < ERROR_RETRY_DELAY) {
    return;
}
```

Also:

* `lastErrorTime` is tracked
* `initializationPromise` is cleared on timeout to unblock future requests

---

## Verification

* Ran dev server and verified normal startup behavior
* Simulated slow/unresponsive upstream:

  * Requests no longer hang indefinitely
  * Server responds within timeout window
* Confirmed:

  * No global blocking during cache refresh
  * Stale data is served while background refresh runs
  * Retry cooldown prevents rapid re-fetch loops
* Updated tests to account for `AbortSignal` usage:

```js
expect(mockFetch).toHaveBeenCalledWith(
    'https://otp.test.example.com',
    expect.objectContaining({ signal: expect.any(AbortSignal) })
);
```

---

### Result

* Eliminates **site-wide request blocking**
* Prevents **unbounded connection buildup**
* Makes cache refresh **transparent to users**
* Ensures graceful degradation during upstream outages

### DIFF
--- a/src/lib/otpServerCache.js
+++ b/src/lib/otpServerCache.js
@@ -25,19 +25,7 @@ async function detectOtpVersion() {
 	const ac = new AbortController();
 	const timer = setTimeout(() => ac.abort(), DETECT_TIMEOUT);
 	try {
-		// Race the fetch against the abort signal so callers are never blocked
-		// indefinitely when the server is unreachable (works with real fetch and mocks).
-		const abortPromise = new Promise((_, reject) => {
-			ac.signal.addEventListener(
-				'abort',
-				() => reject(new DOMException('signal aborted', 'AbortError')),
-				{ once: true }
-			);
-		});
-		const response = await Promise.race([
-			fetch(PUBLIC_OTP_SERVER_URL, { signal: ac.signal }),
-			abortPromise
-		]);
+		const response = await fetch(PUBLIC_OTP_SERVER_URL, { signal: ac.signal });
 
 		if (!response.ok) {
 			throw new Error(`OTP server returned HTTP ${response.status}`);
--- a/src/tests/lib/otpServerCache.test.js
+++ b/src/tests/lib/otpServerCache.test.js
@@ -265,7 +265,14 @@ describe('otpServerCache', () => {
 
 		const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
 
-		mockFetch.mockImplementation(() => new Promise(() => {}));
+		mockFetch.mockImplementation(
+			(_url, { signal } = {}) =>
+				new Promise((_, reject) => {
+					signal?.addEventListener('abort', () =>
+						reject(new DOMException('signal aborted', 'AbortError'))
+					);
+				})
+		);
 
 		const { preloadOtpVersion, getOtpApiType } = await import('$lib/otpServerCache.js');
 


aaronbrethorst (changes requested):
Hey Suhrid, you're making great progress on the server cache timeout fix. You addressed every item from my last review — the `timedOut` flag approach is exactly what I was hoping for, the SWR pattern is clean, and the scoped diff (no more analytics changes) makes this much easier to review. Before we can merge this, I need you to make a couple more changes:

## Important

1. **`otpServerCache.js` has no error cooldown on cold-start failure** (`otpServerCache.js:55-98`). You added `ERROR_RETRY_DELAY` and `lastErrorTime` to `serverCache.js` to prevent hammering OBA after a cold-start failure — great. But `otpServerCache.js` doesn't have the same guard. If the OTP server is unreachable at startup, here's what happens: `detectOtpVersion()` aborts after 10 seconds, `otpApiType` stays `null`, `cacheTimestamp` stays `null`, `initPromise` gets cleared in `.finally()`. On the *next* request, all guards pass and a brand-new detection starts — blocking the handle hook for another 10 seconds. Every single request pays this 10-second penalty until OTP comes back.

   The fix is the same pattern you already wrote in `serverCache.js` — add `lastErrorTime` tracking and an `ERROR_RETRY_DELAY` check:

   ```javascript
   const ERROR_RETRY_DELAY = 30_000;
   let lastErrorTime = null;

   // Inside preloadOtpVersion, after the isStale/warm-cache check:
   if (!otpApiType && lastErrorTime !== null && now - lastErrorTime < ERROR_RETRY_DELAY) {
       return;
   }

   // In the .catch():
   .catch((err) => {
       console.error('OTP version detection failed:', err.message);
       lastErrorTime = Date.now();
   })

   // In the .then():
   .then((type) => {
       otpApiType = type;
       cacheTimestamp = Date.now();
       lastErrorTime = null;
   })

   // Don't forget to reset in clearOtpCache():
   lastErrorTime = null;
   ```

2. **SWR refresh failure in `otpServerCache.js` creates a retry storm** (`otpServerCache.js:77-88`). When the OTP cache is stale and a background SWR refresh fails, `cacheTimestamp` is never updated. So every subsequent request sees `isStale === true` and kicks off *another* background refresh attempt. Under sustained OTP downtime, this means one failed fetch per incoming HTTP request — no backoff, no cooldown. Compare with `serverCache.js`, where `lastErrorTime` and the cooldown guard prevent this.

   The simplest fix: when a SWR background refresh fails and we already have a stale value, update `cacheTimestamp` so the TTL restarts and we don't immediately retry:

   ```javascript
   .catch((err) => {
       console.error('OTP version detection failed:', err.message);
       lastErrorTime = Date.now();
       // If we have stale data, push the TTL forward to prevent retry storm
       if (otpApiType) {
           cacheTimestamp = Date.now();
       }
   })
   ```

3. **Missing test: background fetch resolving after timeout populates the cache** (`serverCache.test.js`). The `timedOut` flag design is all about the background fetch eventually self-healing the cache — but no test verifies this. If a regression breaks the `.then()` handler or the `.finally()` reset, the system would stay in degraded mode permanently after a single timeout. Please add a test like:

   ```javascript
   it('background fetch resolving after timeout populates cache', async () => {
       vi.useFakeTimers();
       let resolveFetch;
       mockAgenciesWithCoverageList.mockImplementation(
           () => new Promise((resolve) => { resolveFetch = resolve; })
       );

       const { preloadRoutesData, getRoutesCache, getCacheState } = await import('$lib/serverCache.js');

       // Cold start times out
       const p1 = preloadRoutesData();
       await vi.advanceTimersByTimeAsync(15_000);
       await p1;
       expect(getCacheState()).toBe('error');

       // Background fetch eventually resolves
       resolveFetch({ data: { list: mockAgencies } });
       // Flush microtasks
       await vi.advanceTimersByTimeAsync(0);

       expect(getCacheState()).toBe('loaded');
       expect(getRoutesCache()).not.toBeNull();
   });
   ```

   (You'll need to adapt the mock to match your `fetchRoutesData` chain, but you get the idea.)

4. **Missing test: error cooldown expires and allows a retry** (`serverCache.test.js`). You test that calls *within* the 30-second window hit the cooldown, but not that calls *after* the window succeed. A boundary bug in the comparison (`<=` vs `<`) would go undetected. Add a test that advances past `ERROR_RETRY_DELAY` and confirms a new fetch is triggered.

## Fit and Finish

5. **Log improvement: distinguish timeout from other failures in `otpServerCache.js`** (`otpServerCache.js:82-84`). When the `AbortController` fires, the error message is `"The operation was aborted"`, which gets logged as `"OTP version detection failed: The operation was aborted"`. Compare with `serverCache.js`, which logs `"[serverCache] Routes fetch timed out"` — much clearer for an operator. Consider checking `err.name === 'AbortError'` to log a timeout-specific message:

   ```javascript
   .catch((err) => {
       if (err.name === 'AbortError') {
           console.warn(`[otpServerCache] OTP version detection timed out after ${DETECT_TIMEOUT}ms`);
       } else {
           console.error('[otpServerCache] OTP version detection failed:', err.message);
       }
       lastErrorTime = Date.now();
   })
   ```

aaronbrethorst (changes requested):
Hey Suhrid, you've done solid work across three rounds on this cache reliability fix — the `timedOut` flag approach, the SWR pattern in both modules, the error cooldown, and the `AbortError`-specific logging all landed well. All prior feedback has been addressed. A few more items surfaced on this pass, mainly around error-handling paths that could silently mislead operators or bypass the cooldown you worked hard to build.

## Important

1. **Empty `catch` in the cold-start timeout path catches *all* errors, not just timeouts** (`serverCache.js:209`). The bare `catch {}` will also catch non-timeout rejections from `fetchRoutesData()` — e.g., a `TypeError` from a malformed API response, a DNS failure, or a network error. When that happens, the log says "Routes fetch timed out" even though the real failure is something else entirely, sending operators on a wild goose chase. Please bind the error and distinguish timeout from non-timeout:

   ```javascript
   } catch (err) {
       if (err.message?.includes('timed out')) {
           console.warn(
               '[serverCache] Routes fetch timed out — requests will proceed without cached data'
           );
       } else {
           console.error('[serverCache] Routes fetch failed during cold start:', err);
       }
       cacheState = 'error';
       lastErrorTime = Date.now();
       timedOut = true;
   }
   ```

2. **`lastErrorTime` is never set when `fetchRoutesData()` returns `null` (non-timeout API errors)** (`serverCache.js:181-185`). `fetchRoutesData()` catches errors internally and returns `null`. The `.then()` handler sees `null`, sets `cacheState = 'error'`, but doesn't set `lastErrorTime`. This means the entire error cooldown mechanism is bypassed for non-timeout failures — if the OBA server returns errors (rather than hanging), every incoming request will immediately re-attempt the fetch with no backoff. Please set `lastErrorTime` when routes is `null`:

   ```javascript
   .then((routes) => {
       routesCache = routes;
       cacheTimestamp = routes ? Date.now() : null;
       cacheState = routes ? 'loaded' : 'error';
       if (routes) {
           lastErrorTime = null;
       } else {
           lastErrorTime = Date.now();
       }
       return routes;
   })
   ```

3. **`AbortController` timeout path is never tested in `otpServerCache.test.js`**. The `detectOtpVersion()` function uses `AbortController` with a 10-second timeout and the `.catch()` handler has `AbortError`-specific logging — but no test triggers the abort. If someone refactors the timeout wiring, no test would catch the regression. Please add:

   ```javascript
   it('times out and enters cooldown when OTP server hangs', async () => {
       vi.useFakeTimers();
       const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});

       mockFetch.mockImplementation(() => new Promise(() => {}));

       const { preloadOtpVersion, getOtpApiType } = await import('$lib/otpServerCache.js');

       const p = preloadOtpVersion();
       await vi.advanceTimersByTimeAsync(10_000);
       await p;

       expect(getOtpApiType()).toBeNull();
       expect(warnSpy).toHaveBeenCalledWith(
           expect.stringContaining('timed out')
       );

       // Verify cooldown is active
       await preloadOtpVersion();
       expect(mockFetch).toHaveBeenCalledTimes(1);

       warnSpy.mockRestore();
       vi.useRealTimers();
   });
   ```

4. **The `timedOut` flag path (`serverCache.js:169`) is not directly tested**. The existing "no duplicate fetch" test hits the cooldown check (line 161) before reaching the `timedOut` branch, because `lastErrorTime` is already set. To verify the `timedOut` branch specifically — which prevents callers from blocking on a hung background promise — add a test where a second call arrives immediately after timeout while `initializationPromise` is still live:

   ```javascript
   it('second caller returns immediately via timedOut flag while background fetch is in flight', async () => {
       vi.useFakeTimers();

       let resolveFetch;
       mockAgenciesWithCoverageList.mockImplementation(
           () => new Promise((resolve) => { resolveFetch = resolve; })
       );

       const { preloadRoutesData, getCacheState } = await import('$lib/serverCache.js');

       const p1 = preloadRoutesData();
       await vi.advanceTimersByTimeAsync(15_000);
       await p1;
       expect(getCacheState()).toBe('error');

       // Immediately call again — background fetch is still in flight
       // (initializationPromise is set, timedOut is true)
       await preloadRoutesData();
       expect(mockAgenciesWithCoverageList).toHaveBeenCalledTimes(1);

       // Clean up
       resolveFetch({ data: { list: [] } });
       for (let i = 0; i < 8; i++) await Promise.resolve();

       vi.useRealTimers();
   });
   ```

## Fit and Finish

5. **Error cooldown in `otpServerCache.js` returns completely silently** (`otpServerCache.js:74-76`). Compare with `serverCache.js:162`, which logs `console.debug('[serverCache] Within error cooldown...')`. The OTP cooldown path just does `return;` with no indication. Add a `console.debug` for consistency:

   ```javascript
   if (!otpApiType && lastErrorTime !== null && now - lastErrorTime < ERROR_RETRY_DELAY) {
       console.debug('[otpServerCache] Within error cooldown — serving without cached OTP version');
       return;
   }
   ```

PR #373, opened 2026-03-07
366 formatlastupdated has no input validation

avrame:
- Added `isFiniteNumber()` utility function
- Using `isFiniteNumber()` to validate params that expect numbers in date/time formatting functions

### DIFF
--- a/src/lib/dateTimeFormat.js
+++ b/src/lib/dateTimeFormat.js
@@ -54,7 +54,7 @@ export function msToTimeString(
 	timeZone = getLocalTimeZone(),
 	dateTimeFormat = localTimeFormat
 ) {
-	if (ms === null || ms === undefined || isNaN(ms)) return 'N/A';
+	if (!Number.isFinite(ms)) return 'N/A';
 	const instant = Temporal.Instant.fromEpochMilliseconds(ms);
 	const plainTime = instant.toZonedDateTimeISO(timeZone).toPlainTime();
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Avi — great job addressing all the feedback from the first review! The `convert24HourTo12Hour` string coercion fix, the test deduplication with `it.each`, and dropping the `isFiniteNumber` wrapper are all exactly right. The core date/time validation logic looks solid. This is ready to merge.

## Previous Feedback — All Addressed

1. **`convert24HourTo12Hour` string handling** — Fixed with `typeof` check + `Number()` coercion, plus a test for string input
2. **Duplicate `formatLastUpdated` tests** — Cleaned up with `it.each(INVALID_INPUTS)`
3. **Duplicate `formatSecondsFromMidnight` tests** — Same treatment
4. **`isFiniteNumber` wrapper** — Removed in favor of `Number.isFinite()` directly
5. **`it.each` suggestion** — Adopted with shared `INVALID_INPUTS` constant

## Strengths

- All five items from the prior review have been addressed cleanly
- The `INVALID_INPUTS` constant + `it.each` pattern is a nice improvement — DRY tests with good readability
- String coercion in `convert24HourTo12Hour` handles the `Object.entries` string-key case correctly
- Using `Number.isFinite()` directly is cleaner than the wrapper
- Net reduction of 67 test lines while maintaining the same coverage — less code, same confidence

Issue #367, opened 2026-03-07
formatSecondsFromMidnight doesn't guard against NaN


aaronbrethorst:
## Context

From PR #362 review. `formatSecondsFromMidnight` (`src/lib/dateTimeFormat.js:88`) guards `null` and `undefined` but not `NaN`.

## Problem

If `secondsSinceMidnight` is `NaN`, `Temporal.PlainTime().add({ seconds: NaN })` throws a `RangeError`. This is inconsistent with `msToTimeString` (line 57) which explicitly checks `isNaN(ms)`.

## Fix

Extend the guard:

```javascript
if (secondsSinceMidnight === null || secondsSinceMidnight === undefined || isNaN(secondsSinceMidnight)) return '';
```

tarunsinghofficial:
Hey @aaronbrethorst, I'm closing this issue as it's already fully resolved in the PR #373 with a better approach, including test coverage. 

**What?**
The guard is already in place. formatSecondsFromMidnight uses if (!Number.isFinite(secondsSinceMidnight)) return '';, which covers NaN (the issue's concern) plus null, undefined, Infinity, and -Infinity. This is a cleaner equivalent to the fix proposed in the issue.

And it's also tested.

PR #375, opened 2026-03-08
fix: guard formatSecondsFromMidnight against NaN input

tarunsinghofficial:
Fxes #367

### DIFF
--- a/src/lib/dateTimeFormat.js
+++ b/src/lib/dateTimeFormat.js
@@ -308,6 +308,9 @@ export function formatLastUpdated(timestamp, translations) {
  * @returns {string|null} OffsetDateTime string, or null if time or date format is invalid
  */
 export function convertToISO8601(date, time) {
+	if (!date || typeof date !== 'string') return null;
+	if (!time || typeof time !== 'string') return null;
+
 	const dateMatch = date.match(/^(\d{2})-(\d{2})-(\d{4})$/);
 	if (!dateMatch) return null;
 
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Hey Tarun, nice targeted fix here — clean and consistent with the existing pattern in `msToTimeString`. Thanks for tracking this down from the PR #362 review.

## Critical

No critical issues found.

## Important

1. **Missing test for the new NaN guard** — The `formatSecondsFromMidnight` test suite has tests for `null` and `undefined` inputs, but no test for `NaN`. The sibling function `msToLocalArrivalDepartureTimeString` does have a NaN test (line 92 in `dateTimeFormat.test.js`). Please add a matching test:

```javascript
it('returns a blank string when the input is NaN', () => {
    expect(formatSecondsFromMidnight(NaN)).toBe('');
});
```

## Fit and Finish

No fit-and-finish issues found. The fix is consistent with the existing `isNaN()` usage in `msToTimeString` (line 57), and the formatting matches Prettier's output.

## Verdict

**Request changes** — the fix itself is correct and I'd like to merge this, but please add the NaN test case first. Once that's in, this is good to go.

tarunsinghofficial:
Closing this PR, the NaN guard for `formatSecondsFromMidnight ` was already addressed in PR #373, which replaced the explicit `null/undefined` check with `!Number.isFinite()` (covering NaN, Infinity, and non-number types). 

My changes became redundant after rebasing against develop.

PR #379, opened 2026-03-11
Clear map results if Trip Planner To or From are cleared.

ict-nick:


aaronbrethorst (changes requested):
Hey Nick, thanks for taking the time to make this improvement to clearing trip planner state. The intent here is solid — when a user clears the "to" or "from" field, the map should clean up after itself. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Pin markers are orphaned on the map when clearing trip data** — `clearTripData()` in `+page.svelte` sets `fromMarker = null` and `toMarker = null` without first calling `mapProvider.removePinMarker()` on them. Since `clearAllPolylines()` only removes polylines (not pin markers), this means pin markers can get stranded on the map with no way to remove them short of a page reload.

   The fix is straightforward — call `removePinMarker` before nulling the references. Both map providers already guard against null markers, so this is safe:

   ```javascript
   function clearTripData() {
       tripItineraries = [];
       tripPlanError = null;
       mapProvider.removePinMarker(fromMarker);
       mapProvider.removePinMarker(toMarker);
       fromMarker = null;
       toMarker = null;
       currentModal = null;
       mapProvider.clearAllPolylines();
   }
   ```

2. **Clearing one field orphans the *other* field's marker** — When `clearInput(true)` is called in `TripPlan.svelte`, it removes the "from" marker from the map, then calls `clearTripData()`, which nulls the page-level `toMarker` reference without removing it from the map. The reverse happens when clearing the "to" field. Since `clearInput` already removes the target marker, and with the fix above `clearTripData` would handle both, this should work correctly after the fix — but please verify this flow end-to-end (clear "from" with an active trip plan, and confirm the "to" pin also disappears from the map).

## Important

3. **Whitespace-only changes mixed with functional changes** — The diff includes three hunks that are purely blank-line additions (`+page.svelte` lines 112, 128, and 220). These aren't harmful, but mixing formatting changes with functional changes makes the diff harder to review. Consider separating them into their own commit, or removing them from this PR to keep the diff focused.

aaronbrethorst (approved):
Hey Nick, nice work on this — the approach of centralizing the trip itinerary cleanup while keeping marker management in `TripPlan.svelte` is clean and makes the responsibility boundaries clear.

### Status: Merge ✅

Tests pass, coverage is solid at 80%, and the logic is correct. The lint failure and merge conflicts will be handled by me during rebase.

### Cleanup (for rebase)

**Dead `fromMarker`/`toMarker` state in `+page.svelte`** — The PR removes these props from `TripPlanModal`, but `+page.svelte` still declares the state (lines 54-55), sets them in `handleTripPlan()` (lines 228-229), and passes them to `<TripPlanModal>` (lines 321-322). All three should be removed during rebase since nothing reads them anymore.

### Strengths

- `clearTripItineraries()` is focused — clears results and polylines without touching markers, which correctly separates concerns
- The `isContextMenuTrigger` guard in `SearchPane.svelte` is a nice fix for preventing unintended tab-switch side effects
- Marker lifecycle stays in `TripPlan.svelte` where the from/to fields live, which is the right ownership model

Issue #209, opened 2025-03-10
Improve mathUtils.test.js - Better Coverage & Maintainability


SophiaLi20:
Hey ,
I’d like to propose some improvements to `mathUtils.test.js` to enhance its test coverage, maintainability, and robustness. Before opening a PR, I’d love to get your thoughts on these changes.

Proposed Changes:
Refactor toDirection tests using test.each - This makes test cases cleaner and reduces redundancy.
Add edge case tests for invalid inputs -Ensure toDirection properly handles non-numeric values (null, undefined, NaN, strings).

Enhance calculateMidpoint tests by:

- Handling large number of stops (performance testing).
- Ensuring identical points return the same location correctly.
- Standardize lon usage instead of mixing lng and lon for consistency.

 Implementation Plan:

- Refactor toDirection tests using test.each to reduce duplication.
- Add edge case handling for invalid inputs (null, undefined, NaN).
- Expand calculateMidpoint tests:
    Test with large datasets (1000+ stops).
    Verify identical stops return the same result.
- Ensure consistent variable naming (lon instead of lng).

Relevant File:
[mathUtils.test.js](https://github.com/OneBusAway/wayfinder/blob/main/src/tests/lib/mathUtils.test.js)

PR #383, opened 2026-03-12
209 - Improve mathUtils.test.js - Better Coverage & Maintainability

avrame:
PR for issue: https://github.com/OneBusAway/wayfinder/issues/209
- Refactored toDirection tests to use it.for (an alias for test.each)
- Added edge cases to toDirection tests for non-numeric values
- Added a performance test for calculateMidpoint() with 1000 stops
- Added a test to calculateMidpoint() for identical stops

### DIFF
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -64,7 +64,9 @@
 		// TODO: implement better way to transition to route shape
 		const location = calculateMidpoint(stops);
 
-		mapProvider.flyTo(location.lat, location.lng, 13);
+		if (location) {
+			mapProvider.flyTo(location.lat, location.lon, 13);
+		}
 
 		for (const stopTime of stopTimes) {
 			const stop = stops.find((s) => s.id === stopTime.stopId);
--- a/src/components/search/SearchPane.svelte
+++ b/src/components/search/SearchPane.svelte
@@ -118,7 +118,9 @@
 			}
 
 			const midpoint = calculateMidpoint(orderedStops);
-			mapProvider.flyTo(midpoint.lat, midpoint.lng, 12);
+			if (midpoint) {
+				mapProvider.flyTo(midpoint.lat, midpoint.lon, 12);
+			}
 
 			for (const polylineData of polylinesData) {
 				const shape = polylineData.points;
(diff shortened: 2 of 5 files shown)


aaronbrethorst (changes requested):
Hey Avi, thanks for taking the time to improve the test coverage and maintainability of `mathUtils.test.js`. The refactoring to parameterized tests with `it.for()` is really clean -- it makes adding new test cases trivial while keeping the tests readable. The `lng` to `lon` rename is also a nice consistency fix. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Stale mock in `SearchPane.test.js` not updated for `lng` -> `lon` rename.** The mock at `src/components/search/__tests__/SearchPane.test.js` line 20 still returns `{ lat: 47.6062, lng: -122.3321 }`, but `calculateMidpoint()` now returns `lon`. This means the SearchPane tests silently pass `undefined` as the longitude to `mapProvider.flyTo()`. Update it to:
   ```js
   calculateMidpoint: vi.fn().mockReturnValue({ lat: 47.6062, lon: -122.3321 })
   ```

## Important

2. **`calculateMidpoint([])` with an empty array produces `{ lat: NaN, lon: NaN }`.** This is pre-existing behavior, but since you're already improving this function, consider adding a guard for empty input (return `null` or throw) and a corresponding test. If an API response returns an empty stops array, the map provider would receive `flyTo(NaN, NaN, ...)`, which could cause silent failures.

3. **Consider whether returning `0` for non-finite inputs in `toDirection()` is the right choice.** Returning `0` means "north" -- callers can't distinguish between "the vehicle is actually facing north" and "we have no orientation data." Returning `null` would let callers decide how to handle missing data. If `0` is intentional as a safe default, a brief code comment explaining the rationale would help future maintainers.

## Fit and Finish

4. **The "can handle a large number of stops" test name is misleading.** It's actually a correctness test with a larger input, not a performance test. Consider renaming it to something like "calculates correct midpoint for large input" to set accurate expectations.

5. **Pre-existing: `return direction === 0 ? 0 : direction` on line 20 of `mathUtils.js` is a no-op** -- it returns `direction` in all cases. If you're already touching this function, simplifying it to `return direction;` would be a nice cleanup.

PR #385, opened 2026-03-13
fix: crash when alerts have empty activePeriod

suhr25:
## SUMMARY

Fixes a crash in the alerts API when an alert has an empty `activePeriod`. Since this field can be empty in the GTFS-RT spec, accessing `activePeriod[0]` without checks can throw a runtime error.
Adds simple guards in `isStartDateWithin24Hours()` in `src/routes/api/oba/alerts/+server.js`.

---

## FIX

**Before**

```javascript
const startDate = alert.activePeriod[0].start;
```

**After**

```javascript
if (!alert.activePeriod || alert.activePeriod.length === 0) return false;

const startDate = alert.activePeriod[0].start;
if (!startDate) return false;
```

Adds defensive checks so the function safely skips alerts without a valid `activePeriod`.

---

## VERIFICATION

**Steps**

1. Use an alert with `activePeriod: []`.
2. Call `/api/oba/alerts`.

### DIFF
--- a/src/routes/api/oba/alerts/+server.js
+++ b/src/routes/api/oba/alerts/+server.js
@@ -89,8 +89,11 @@ function getSeverityLevel(alert) {
 	return alert.severityLevel;
 }
 
-function isStartDateWithin24Hours(alert) {
+export function isStartDateWithin24Hours(alert) {
+	if (!alert) return false;
+	if (!alert.activePeriod || alert.activePeriod.length === 0) return false;
 	const startDate = alert.activePeriod[0].start;
+	if (!startDate) return false;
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Suhrid, thanks for addressing all the feedback — the tests cover every guard clause thoroughly, the `alert` null check is in place for consistency, and the branch is cleanly rebased and squashed into a single commit. Really solid follow-through.

There's nothing left to change — this PR is ready to merge. Defensive checks like this against GTFS-RT spec optionality are exactly the kind of small fix that prevents frustrating production crashes down the road. Nice work.

PR #390, opened 2026-03-15
fix: add null safety guards to hexToRgb and getBrightness

ayyagarisujanreddy123:
this fixes a crash in the color utility pipeline when route or agency color data is missing.

hexToRgb() was calling .replace() on null/undefined, and getBrightness() could then crash by accessing .r/.g/.b on a null value. I added small defensive guards so these functions fail safely instead of throwing.

Changes:

hexToRgb() now returns null for missing or non-string input

getBrightness() now returns 0 for falsy input
Testing

Verified that:
valid 3-digit and 6-digit hex values still work

invalid inputs (null, undefined, empty string, bad values, non-strings) return safe fallbacks

getBrightness(hexToRgb(null)) no longer crashes

related helpers like adjustColorForDarkMode() and generatePalette() still handle missing colors safely

All tests are passing.

### DIFF
--- a/src/lib/colorUtils.js
+++ b/src/lib/colorUtils.js
@@ -10,6 +10,7 @@
  * @returns {{r: number, g: number, b: number} | null} RGB object or null if invalid
  */
 export function hexToRgb(hex) {
+	if (!hex || typeof hex !== 'string') return null;
 	hex = hex.replace(/^#/, '');
 	// Expand 3-digit hex to 6-digit
 	if (hex.length === 3) {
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Hey Sujan, thanks for taking the time to tighten up the null safety in our color utilities. Adding early guards to `hexToRgb` and `getBrightness` is a sensible defensive measure — calling `.replace()` on `null` is exactly the kind of thing that crashes in production when API data comes back incomplete. Before we can merge this, I will need you to make a couple changes:

## Important

1. **Missing tests for the new null/undefined guards.** The existing `hexToRgb` test suite covers empty string and invalid strings, but doesn't exercise `null`, `undefined`, or non-string types (e.g. a number). Similarly, `getBrightness` has no test for falsy input returning `0`. Since the whole point of this PR is to handle these edge cases, please add tests that verify the new behavior:
   - `hexToRgb(null)` returns `null`
   - `hexToRgb(undefined)` returns `null`
   - `hexToRgb(123)` returns `null` (non-string type)
   - `getBrightness(null)` returns `0`
   - `getBrightness(undefined)` returns `0`

## Fit and Finish

2. **Whitespace-only reformatting mixed into the functional diff** (`src/lib/colorUtils.js:24-28`). The indentation change to the ternary return object is fine on its own, but mixing formatting changes with functional changes makes the diff harder to review. Not a blocker — just something to be aware of for future PRs. If you'd like, you can split this into a separate commit so the functional change is isolated.

PR #391, opened 2026-03-16
Fix/route schedule table empty times guard

suhr25:
## SUMMARY

This PR fixes a crash in the stop schedule table when an hour contains an empty `times` array. The component assumed `times[0]` always existed and tried to read `arrivalTime`, which caused a runtime error.

A small guard was added in **`RouteScheduleTable.svelte`** to safely handle this case.

---

## FIX

**Before**

```svelte
title="Full Time: {hour}:{extractMinutes(times[0].arrivalTime)}"
```

**After**

```svelte
title="Full Time: {hour}:{times[0]?.arrivalTime ? extractMinutes(times[0].arrivalTime) : ''}"
```

This ensures `extractMinutes()` only runs when a valid `arrivalTime` is present.

---

## VERIFICATION

1. Open `/stops/<stopID>/schedule`
2. Simulate a case where a stopTimes hour returns an empty array (`"3": []`)

**Expected:** The schedule table renders normally and no error occurs.


aaronbrethorst (changes requested):
Hi Suhrid, thanks for tracking down this crash in the schedule table. Defensive guards against unexpected data shapes are definitely good practice. I have one concern about the approach that I'd like to see addressed before merging.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

### 1. The guard is incomplete — empty `times` arrays still render broken rows

The fix guards the `title` attribute on the hour `<td>`, but if `times` is an empty array, the row still renders: you get an hour label (e.g., "3 AM") with no minute pills next to it, because the `{#each times as stopTime}` loop produces nothing.

A more robust fix would filter out empty arrays at the data level so these ghost rows never appear:

```js
// in renderScheduleTable()
const amTimes = stopTimes.filter(([hour, times]) => +hour < 12 && times.length > 0);
const pmTimes = stopTimes.filter(([hour, times]) => +hour >= 12 && times.length > 0);
```

This eliminates the crash **and** the empty-row display bug in one place, and you can revert the `title` attribute back to its original simpler form since `times[0]` is guaranteed to exist.

## Suggestions (0 found)

## Strengths

- Good instinct to add defensive checks — runtime crashes on schedule pages are a poor user experience.
- Clear PR description with before/after code and reproduction steps.

## Recommended Action

1. Filter empty arrays in `renderScheduleTable()` instead of (or in addition to) guarding in the template.
2. With that fix, the `title` attribute can remain simple.

**Verdict: Request Changes**

PR #396, opened 2026-03-19
fix: use brand theme tokens for trip planner From/To markers

### LINKED ISSUE #392: Use brand colors for the "From" and "To" trip planner icons
"From" and "To" bubbles should use `brand-accent` background and `brand-foreground` for text.

The green pin should also use `brand-accent`

Image

ayyagarisujanreddy123:
## Summary
Updated the trip planner “From” and “To” markers to use Wayfinder’s brand theme tokens instead of a hardcoded green color.

This keeps the marker bubbles and pin visually consistent with the configured agency branding and the rest of the app.

## Problem
The trip planner “From” / “To” label bubbles and map pin were using a fixed green color (`#79aa38`) rather than the app’s theme tokens. As a result, these markers did not reflect agency-specific branding and could look inconsistent in branded deployments.

### `src/components/trip-planner/tripPlanPinMarker.svelte`
- Replaced bubble background color with `bg-brand-accent`
- Replaced bubble text color with `text-brand-foreground`
- Updated the pin styling to use the brand accent theme instead of the hardcoded green
- Removed the unused `color` prop since it was not passed by any caller

### `src/components/trip-planner/__tests__/TripPlanPinMarker.test.js`
- Added 5 unit tests to verify:
  - brand theme classes are applied correctly
  - hardcoded green styling is no longer used
  - marker rendering remains stable

## Test Plan
- [x] All 1162 tests pass (53 test files)
- [x] Verified that the marker uses brand theme classes

Closes #392

### DIFF
--- a/src/components/trip-planner/__tests__/TripPlanPinMarker.test.js
+++ b/src/components/trip-planner/__tests__/TripPlanPinMarker.test.js
@@ -1,0 +1,33 @@
+import { describe, it, expect } from 'vitest';
+import { render } from '@testing-library/svelte';
+import TripPlanPinMarker from '../tripPlanPinMarker.svelte';
+
+describe('TripPlanPinMarker', () => {
+	it('renders the label text', () => {
+		const { getByText } = render(TripPlanPinMarker, { props: { text: 'From' } });
+		expect(getByText('From')).toBeInTheDocument();
+	});
+
+	it('uses brand-accent background and brand-foreground text on the bubble', () => {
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Sujan, thanks for taking the time to make this fix to the trip planner markers. Swapping the hardcoded green for brand theme tokens is a clean improvement that will make branded deployments look consistent out of the box. I also appreciate that you verified the `color` prop was unused before removing it and added solid test coverage.

There's nothing left to change — this PR is ready to merge. Nice work keeping it focused and well-scoped.

PR #397, opened 2026-03-19
fix: enable translation of navbar buttons

### LINKED ISSUE #395: fix: translate navigation bar buttons when switching languages
Hello OBA community! My name is Kobi, and I’m new here and excited to contribute!

I noticed that when users switch languages, the navigation bar labels remain in English. This seems inconsistent with the rest of the interface.

**Proposed solution:**
Render the navigation bar labels using `svelte-i18n`. Specifically:

* Import `t` from `svelte-i18n` in `src/components/navigation/Header.svelte`
* Replace `<a>{key}</a>` with the `<a>{$t(key)}</a>` in `src/components/navigation/OverflowMenu.svelte`
* Update the value of `PUBLIC_NAV_BAR_LINKS` in `.env.example` to support translations

I tested this approach locally, and it appears to work as expected.

Before opening a pull request, I wanted to ask whether this change would be welcome, or if there are any preferred approaches for handling translations in this part of the codebase.

Thanks a lot!

kobichiu:
## Summary
This PR fixes an UI issue where navigation bar buttons remain in English after switching languages.

## Related Issue
Closes #395 

## Changes Made
- Imported `t` from `svelte-i18n` in `src/components/navigation/Header.svelte` and `src/components/navigation/OverflowMenu.svelte`
- Updated the original `<a>{key}</a>` to `<a>{$t(key)}</a>` in both files
- Updated the `PUBLIC_NAV_BAR_LINKS` IN `.env.example` to enable translation of the 4 buttons

## Notes
This PR only affects static navigation bar labels

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@ PRIVATE_OBACO_SHOW_TEST_ALERTS=false
 # Example: PRIVATE_OBA_AGENCY_FILTER=19,29
 PRIVATE_OBA_AGENCY_FILTER=
 
-PUBLIC_NAV_BAR_LINKS={"Home": "/","About": "/about","Contact": "/contact","Fares & Tolls": "/fares-and-tolls"}
+PUBLIC_NAV_BAR_LINKS={"header.home": "/","header.about": "/about","header.contact": "/contact","header.fares_and_tolls": "/fares-and-tolls"}
 PUBLIC_OBA_GOOGLE_MAPS_API_KEY=""
 PUBLIC_OBA_LOGO_URL="https://onebusaway.org/wp-content/uploads/oba_logo-1.png"
 OBA_LOGO_URL_DARK=""
--- a/src/components/navigation/Header.svelte
+++ b/src/components/navigation/Header.svelte
@@ -5,6 +5,7 @@
 		PUBLIC_NAV_BAR_LINKS
 	} from '$env/static/public';
 
+	import { t } from 'svelte-i18n';
 	import { onMount } from 'svelte';
 	import OverflowMenu from './OverflowMenu.svelte';
 	import LanguageSwitcher from './LanguageSwitcher/LanguageSwitcher.svelte';
@@ -252,8 +253,9 @@
 				<a
 					href={value}
 					class="block px-2 py-1 font-semibold text-surface-foreground dark:text-surface-foreground-dark"
-					>{key}</a
-				>
+					>
+						{$t(key)}
+				</a>
 			</div>
 		{/each}
 
--- a/src/components/navigation/OverflowMenu.svelte
+++ b/src/components/navigation/OverflowMenu.svelte
@@ -1,4 +1,5 @@
 <script>
+	import { t } from 'svelte-i18n';
 	let { links = [], onClose } = $props();
 	let menuRef = $state(null);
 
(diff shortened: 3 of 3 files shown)


aaronbrethorst (changes requested):
Hey Kobi — nice work tackling the i18n gap in the navigation bar! Getting those hardcoded labels wired up to the translation system is a valuable improvement. I found one issue that needs to be addressed before this can merge.

## Critical Issues (1 found)

### 1. `measureLinkWidths()` measures the i18n key, not the translated text

**File:** `src/components/navigation/Header.svelte:71`

The `measureLinkWidths()` function creates an off-screen element and sets its `textContent` to `key` — which, after this PR, is the i18n key string (e.g., `"header.home"`) rather than the translated label (e.g., `"Home"` or `"Inicio"`).

```js
tempDiv.textContent = key; // "header.home" instead of "Home"
```

This means the overflow calculation will measure the pixel width of `"header.fares_and_tolls"` (20 characters) rather than `"Fares & Tolls"` (13 characters) or its translation in other languages. The result is that the header's responsive overflow behavior — deciding which links are visible vs. tucked into the overflow menu — will be based on incorrect widths. In some languages where translations are longer than the English text, links could overflow or get cut off unexpectedly.

**Fix:** Use the translated string for measurement:

```js
tempDiv.textContent = get(t)(key); // import { get } from 'svelte/store'
```

Or, since `$t` is available in the component's reactive context but not inside a plain function, you may need to pass the translation function in or restructure slightly. The `get()` helper from `svelte/store` is the cleanest approach for reading a store value imperatively.

Additionally, `measureLinkWidths` should ideally be re-called when the locale changes, since translated strings may have different widths. Currently it only runs once in `onMount`.

## Important Issues (1 found)

### 1. Existing tests use raw English strings as keys

**File:** `src/components/navigation/__tests__/Header.test.js:9`

The test mock still uses raw English strings as nav link keys:

```js
PUBLIC_NAV_BAR_LINKS: '{"Home": "/", "About": "/about"}'
```

While the tests don't currently assert on the rendered link text (so they pass), this mock no longer reflects how the component is actually used post-PR. Consider updating the test mock to use i18n keys (e.g., `'{"header.home": "/", "header.about": "/about"}'`) to keep tests aligned with reality. This would also be a good place to add a test that verifies the translated text renders correctly.

## Suggestions (1 found)

### 1. Breaking change for existing deployments

Changing the `.env.example` keys from human-readable strings (`"Home"`, `"About"`) to i18n keys (`"header.home"`, `"header.about"`) is the right call, but anyone with an existing `.env` file will still have the old values. After this PR merges, their nav bar will show the raw i18n keys (e.g., `"header.home"`) if `svelte-i18n` can't find a translation, or worse, it may show the key string literally.

Consider adding a note in the PR description or a brief mention in the README about the migration needed for existing deployments.

## Strengths

- Clean, focused change — only touches the files that need it
- Correctly uses `$t(key)` in both `Header.svelte` and `OverflowMenu.svelte`
- Translation keys already exist in all 25 locale files, so this is ready to go from an i18n content perspective
- Good PR description with clear summary and context

## Recommended Action

1. **Must fix:** Update `measureLinkWidths()` to use the translated string for width measurement
2. **Should fix:** Update test mocks to use i18n keys
3. **Consider:** Add migration note for existing deployments
4. **After fixes:** Re-run review to verify

PR #411, opened 2026-03-21
feat: implement smooth fade-in animation for route polylines #407

tejasva-vardhan:
## Description
This PR implements a smooth fade-in transition for route polylines, addressing the legacy TODO in `RouteMap.svelte`. It replaces the abrupt "popping" of route shapes with a polished, high-performance animation.

Fixes #407

## Key Changes
- **Animation Logic:** Refactored `createPolyline` in `OpenStreetMapProvider` to use `requestAnimationFrame` for a smooth **600ms** opacity transition.
- **Zero-Jank Sync:** Synchronized the line fade-in with the arrow decorators. Arrows now trigger in the same frame as the final opacity lock, ensuring a seamless visual entry.
- **Resource Management:** Implemented `cancelAnimationFrame` cleanup to prevent memory leaks or animation overlaps during rapid route switching.
- **Code Quality:** - Moved arrow decorator logic to a dedicated helper method `addArrowDecorator`.
    - Removed the TODO comment in `src/components/map/RouteMap.svelte`.
    - Added state reset (`shapeData = null`) before new fetches to prevent data ghosting.

## Verification
- [x] Verified smooth 600ms transition on OpenStreetMap provider.
- [x] Confirmed arrow decorators sync perfectly with the animation completion.
- [x] `npx eslint` passes on modified files.
- [x] Verified no jank or performance lag during map movement.

## Demo

https://github.com/user-attachments/assets/2015c220-1a2f-4c58-b8bf-fd498d79bf36

### DIFF
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -51,8 +51,12 @@
 		const routeId = moreTripData?.routeId;
 
 		if (shapeId && isMounted) {
+			shapeData = null;
 			const shapeResponse = await fetch(`/api/oba/shape/${shapeId}`);
-			shapeData = await shapeResponse.json();
+			if (shapeResponse.ok) {
+				shapeData = await shapeResponse.json();
+			}
+
 			const shapePoints = shapeData?.data?.entry?.points;
 			if (shapePoints && isMounted) {
 				await mapProvider.createPolyline(shapePoints);
@@ -61,7 +65,6 @@
 
 		const stopTimes = tripData?.data?.entry?.schedule?.stopTimes ?? [];
 		const stops = tripData?.data?.references?.stops ?? [];
-		// TODO: implement better way to transition to route shape
 		const location = calculateMidpoint(stops);
 
 		mapProvider.flyTo(location.lat, location.lng, 13);
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -55,6 +55,8 @@
 			const shapeResponse = await fetch(`/api/oba/shape/${shapeId}`);
 			if (shapeResponse.ok) {
 				shapeData = await shapeResponse.json();
+			} else {
+				console.error(`Failed to fetch shape: shapeId=${shapeId}, status=${shapeResponse.status}`);
 			}
 
 			const shapePoints = shapeData?.data?.entry?.points;
--- a/src/lib/Provider/OpenStreetMapProvider.svelte.js
+++ b/src/lib/Provider/OpenStreetMapProvider.svelte.js
@@ -464,11 +464,15 @@ export default class OpenStreetMapProvider {
 		}
 
 		const withArrow = options.withArrow ?? true;
+		const targetOpacity = options.opacity ?? 1;
+
+		// Used to avoid running fade/arrow logic after the polyline was removed.
+		const fadeInToken = Symbol('fadeInToken');
 
 		const polylineOpts = {
 			color: options.color || COLORS.POLYLINE,
 			weight: options.weight || 4,
-			opacity: options.opacity ?? 1
+			opacity: 0
 		};
 		if (options.dashArray) {
 			polylineOpts.dashArray = options.dashArray;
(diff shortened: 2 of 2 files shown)


aaronbrethorst (changes requested):
Hey Tejasva, the animation work here is genuinely impressive -- the Symbol-based token invalidation pattern and the deferred viewport-aware fade-in show real thought about edge cases. The demo video looks smooth. I have a few structural issues I'd like addressed before merging.

## Critical Issues (0 found)

None.

## Important Issues (3 found)

### 1. Cleanup logic duplicated between `removePolyline` and `clearAllPolylines`

The fade-in cleanup (cancel RAF, remove moveend listener, null out token) is copy-pasted across both methods (~10 identical lines each). This is the time to fix it, before it ships.

Extract a helper:

```javascript
_cancelFadeIn(polyline) {
    if (polyline.__fadeDeferredMoveEnd) {
        this.map?.off('moveend', polyline.__fadeDeferredMoveEnd);
        polyline.__fadeDeferredMoveEnd = null;
    }
    if (polyline.__fadeInRafId != null) {
        cancelAnimationFrame(polyline.__fadeInRafId);
        polyline.__fadeInRafId = null;
    }
    polyline.__fadeInToken = null;
}
```

Then both `removePolyline()` and `clearAllPolylines()` call `this._cancelFadeIn(polyline)`. You could go further and extract `_cleanupPolyline()` that also handles arrow decorator removal, since that's duplicated too.

Note the `this.map?.off(...)` -- this also fixes a subtle issue where the current code guards with `polyline.__fadeDeferredMoveEnd && this.map`, which means if `this.map` is nulled during teardown, the listener is never removed. Using optional chaining is safer.

`OpenStreetMapProvider.svelte.js:566-579, 598-618`

### 2. Use Leaflet's `map.once()` instead of manual `off()` inside the moveend handler

The `onMoveEnd` handler manually calls `this.map.off('moveend', onMoveEnd)` to self-remove. Leaflet has `map.once()` for exactly this pattern:

```javascript
if (overlapsViewport) {
    beginFadeIn();
} else {
    const onMoveEnd = () => {
        polyline.__fadeDeferredMoveEnd = null;
        beginFadeIn();
    };
    polyline.__fadeDeferredMoveEnd = onMoveEnd;
    this.map.once('moveend', onMoveEnd);
}
```

The cleanup in `_cancelFadeIn` would still call `this.map?.off('moveend', ...)` for the case where the polyline is removed before moveend fires, but the happy path becomes cleaner.

`OpenStreetMapProvider.svelte.js:529-535`

### 3. Shape fetch failure is silently swallowed

The new `.ok` check on the shape response is a good addition, but the `else` case does nothing -- no logging, no feedback. When the route line silently fails to appear, debugging will be painful.

```javascript
if (shapeResponse.ok) {
    shapeData = await shapeResponse.json();
} else {
    console.error(`Failed to fetch shape: shapeId=${shapeId}, status=${shapeResponse.status}`);
}
```

`RouteMap.svelte:56-58`

## Suggestions (1 found)

- [code-simplifier]: Consider using a `WeakMap` keyed by polyline instead of dunder properties (`__fadeInToken`, `__fadeInRafId`, `__fadeDeferredMoveEnd`) to keep animation state separate from Leaflet's object. This avoids potential namespace collisions with future Leaflet internals. Not a blocker -- dunder props are common in Leaflet plugin code -- but worth considering. `OpenStreetMapProvider.svelte.js:484-486`

## Strengths

- The `Symbol`-based token pattern for invalidating stale animations is clean and race-condition-resistant.
- Deferring the fade-in until the polyline is in the viewport (or the next `moveend`) prevents wasted animation frames during `flyTo`.
- Arrow decorators synced to animation completion avoids visual jank.
- `shapeData = null` before the fetch and the `.ok` guard are defensive improvements in `RouteMap.svelte`.
- The cubic ease-out (`1 - Math.pow(1 - progress, 3)`) is a nice touch for natural-feeling motion.

## Recommended Action

1. Extract the duplicated cleanup into a `_cancelFadeIn()` helper (or `_cleanupPolyline()` if you want to also deduplicate arrow decorator cleanup).
2. Switch to `this.map.once('moveend', ...)` for the deferred fade-in.
3. Add `console.error` in the shape fetch failure branch.

**Verdict: Request Changes**

aaronbrethorst (changes requested):
Hey Tejasva, nicely done on this revision — all four items from the last round are addressed cleanly. The WeakMap-based state, the `map.once('moveend', ...)` switch, the `_cleanupPolyline` helper, and the shape-fetch logging all landed well. A couple of small things worth tightening before this goes in.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

### 1. `_cleanupPolyline` triggers a Leaflet warning on every in-viewport polyline cleanup

`src/lib/Provider/OpenStreetMapProvider.svelte.js:582`

```js
this.map?.off('moveend', state?.fadeDeferredMoveEnd);
```

For any polyline that began fading in immediately (the common case — the polyline overlaps the viewport, so we never schedule a deferred `moveend`), `state.fadeDeferredMoveEnd` is `null`. That call becomes `this.map.off('moveend', null)`, which inside Leaflet's `_off()` hits this branch:

```js
if (typeof fn !== 'function') {
    console.warn('wrong listener type: ' + typeof fn);
    return;
}
```

So every route switch / map clear will spam `wrong listener type: object` (or `undefined`) once per polyline. It's not a correctness bug — no unintended listener removal happens — but it's noisy enough to obscure real warnings during dev. Guard it:

```js
if (state?.fadeDeferredMoveEnd) {
    this.map?.off('moveend', state.fadeDeferredMoveEnd);
}
if (state) {
    state.fadeDeferredMoveEnd = null;
}
```

## Suggestions (2 found)

### 1. `state.fadeInToken` is dead code

`OpenStreetMapProvider.svelte.js:484`

The Symbol is written into `state` once and never read anywhere. The WeakMap lookup already handles invalidation: when `_cleanupPolyline` does `polylineAnimationState.delete(polyline)`, subsequent `step()` ticks early-return because `stepState` is undefined. The token has no remaining purpose — safe to drop the field.

### 2. Hoist the fade duration to a named constant

`OpenStreetMapProvider.svelte.js:498`

`const durationMs = 600` is buried inside the closure. A module-level constant next to `polylineAnimationState` would mirror how `COLORS` is treated elsewhere in this file and make the timing easier to find/tune:

```js
const FADE_IN_DURATION_MS = 600;
```

## Strengths

- All four prior-review items addressed correctly — particularly nice that you took the optional `WeakMap` suggestion all the way through.
- Centralizing animation state in the module-level `polylineAnimationState` keeps the `OpenStreetMapProvider` instance clean and removes the dunder-property namespace pollution.
- The `step()` callback's defensive checks (`!stepState || !this.map?.hasLayer(polyline)`) make the RAF lifecycle robust to mid-flight removal.
- `addArrowDecorator` extraction is clean and the new helper boundary makes the fade-then-arrows sequencing obvious.
- `shapeData = null` + `.ok` guard + `console.error` in `RouteMap.svelte` are exactly what was asked for.
- Prettier passes, eslint passes.

## Recommended Action

1. Guard the `off('moveend', ...)` call so it only fires when there's actually a handler to remove
2. Delete the unused `fadeInToken` field
3. Hoist `600` to a named module-level constant

### Verdict: **Request Changes**

aaronbrethorst:

PR #412, opened 2026-03-21
Task/402 unit tests stabilization

### LINKED ISSUE #402: Add unit tests for small, self-contained components
## What needs to change

Several small components don't have any tests yet. Writing tests for these is a great way to get familiar with the codebase and the test setup.

## Good components to start with (pick one or more)

- `src/components/FullPageLoadingSpinner.svelte` — renders a loading spinner
- `src/components/controls/CompassArrow.svelte` — renders a direction indicator
- `src/components/trip-planner/OptionsPill.svelte` — renders a small pill/chip UI element
- `src/components/tabs/TabLink.svelte` — renders a tab navigation link

These are small, have few dependencies, and are good candidates for focused tests.

## How tests work in this project

- Tests live in `src/tests/` and use the naming pattern `ComponentName.test.js`
- The framework is **Vitest** with a jsdom environment
- `vitest-setup.js` already mocks common dependencies (env vars, i18n, SvelteKit stores, etc.)
- Look at existing tests in `src/tests/` for patterns to follow
- Run tests with `npm run test` or `npm run test:watch` during development

## What to test

For each component, consider:
- Does it render without crashing?
- Does it display the expected text/content given certain props?
- If it has conditional rendering (e.g., loading vs. loaded), do both states work?

## Acceptance criteria

- [ ] At least one new test file is added to `src/tests/`
- [ ] Tests pass with `npm run test`
- [ ] Tests cover basic rendering and at least one meaningful behavior

tejasva-vardhan:
### Description
This PR adds unit tests for the small, self-contained components as requested in #402. These tests cover basic rendering, prop-driven behavior, and conditional states using Vitest and @testing-library/svelte.

### Components Added:
- **FullPageLoadingSpinner**: Tests visibility and rendering.
- **CompassArrow**: Validates rotation based on the `heading` prop.
- **OptionsPill**: Tests labels and click interactions.
- **TabLink**: Verifies ARIA roles and navigation attributes.

### Additional Fixes:
While implementing these, I noticed some flakiness in the existing `StopModal` and `dateTimeFormat` tests. I've included minor stubs and normalization to ensure the full suite passes consistently in CI.

### Verification:
- `npx vitest run` -> 1167 tests passed.
- ESLint passed on all modified files.

Closes #402


aaronbrethorst (changes requested):
Hey Tejasva, thanks for tackling the test coverage for these four components -- the TabLink and OptionsPill tests in particular are nicely structured. I can see you put real thought into the StopModal flakiness fix and the dateTimeFormat meridiem normalization, and those are welcome improvements. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Split the polyline fade-in animation into a separate PR.** The 82-line animation system in `OpenStreetMapProvider.svelte.js` (fade-in with `requestAnimationFrame`, viewport intersection detection, `moveend` deferral, Symbol-based cancellation tokens) is a feature for #407, not test stabilization for #402. It also introduces a behavioral inconsistency -- OSM users see a smooth fade-in while Google Maps users see an instant polyline. This needs its own PR with its own review, tests, and ideally a matching implementation in `GoogleMapProvider`. Please revert this change from this PR and submit it separately.

2. **Fix or remove `role="tab"` on `TabLink.svelte`.** Per the WAI-ARIA spec, `role="tab"` requires the element to be owned by an element with `role="tablist"`. The parent `<div class="tab-container__item">` has no such role, and the grandparent container doesn't either. This will be flagged by accessibility audit tools (axe, Lighthouse). Two options:
   - Add `role="tablist"` to the parent container that wraps TabLink instances, making it a proper ARIA tab pattern.
   - Remove `role="tab"` and keep these as standard links (their current semantic role).

   If you choose to add the tablist pattern, please update the StopPageHeader tests accordingly.

3. **Run Prettier.** Two files fail formatting:
   - `src/components/FullPageLoadingSpinner.svelte` -- the content inside `{#if visible}` needs indentation
   - `src/components/stops/__tests__/StopModal.test.js` -- minor formatting differences

   Run `npm run format` to fix both.

## Important

4. **Add error logging for the shape fetch failure in `RouteMap.svelte`.** The `if (shapeResponse.ok)` guard you added (line 56) is the right idea, but when it fails, the polyline silently doesn't render with no logging. Every other fetch in the codebase logs errors on failure. Add a `console.error` in the else branch:
   ```javascript
   if (shapeResponse.ok) {
       shapeData = await shapeResponse.json();
   } else {
       console.error(`Failed to fetch shape ${shapeId}: ${shapeResponse.status}`);
   }
   ```

5. **Reduce the 15-second test timeout in StopModal.test.js.** The `stubArrivalsFetch` + `userEvent.setup({ delay: null })` should resolve the flakiness, making the 15s/10s timeouts unnecessary. A 5-second test timeout with a 2-second `waitFor` would be a more reasonable safety margin. As-is, a genuinely broken test will waste 15 seconds before failing in CI.

## Fit and Finish

6. **Replace `options-pill--active` with a `data-testid`.** The class on `OptionsPill.svelte` line 14 isn't defined in any CSS -- it only exists as a test hook. The `@testing-library` convention is to use `data-testid="options-pill-active"` for test selectors instead of empty CSS classes.

7. **Remove `class=""` from CompassArrow.svelte.** In the `heading` branch (line 51 in the diff), `<FontAwesomeIcon icon={faArrowRight} class="" />` passes an empty class string. Either omit the `class` prop entirely or leave it unset.

8. **Move new test files to co-located `__tests__` directories.** The project convention is `src/components/<feature>/__tests__/`. The new tests in `src/tests/` should be:
   - `CompassArrow.test.js` -> `src/components/controls/__tests__/`
   - `OptionsPill.test.js` -> `src/components/trip-planner/__tests__/`
   - `TabLink.test.js` -> `src/components/tabs/__tests__/`
   - `FullPageLoadingSpinner.test.js` -> `src/components/__tests__/`

aaronbrethorst (changes requested):
Hi Tejasva, thanks for the quick turnaround on the prior feedback. The polyline animation extraction is clean, the test relocations landed in the right spots, and the `StopModal` stub + timeout reduction is exactly what I was after. The `dateTimeFormat` ICU meridiem normalization is a nice touch — it's the right way to handle that runtime variance.

There's one blocker, though, and a couple of smaller things worth tightening before this merges.

## Critical

1. **The test suite is failing.** Running `npx vitest run` produces **5 failures** in `src/components/stops/__tests__/StopPageHeader.test.js`. The `TabLink.svelte` change in commit `bb0a47f` removed `role="tab"` and `aria-selected` from the anchor (good — that's option 2 from my last review), but the `StopPageHeader.test.js` changes still query `getByRole('tab', ...)` and assert `aria-selected="true"`. The two edits contradict each other.

   Concretely, these need to revert to `role: 'link'`:
   - `src/components/stops/__tests__/StopPageHeader.test.js:132` — `getByRole('tab', { name: /arrivals & departures/i })`
   - `src/components/stops/__tests__/StopPageHeader.test.js:136` — `getByRole('tab', { name: /route schedules/i })`
   - `src/components/stops/__tests__/StopPageHeader.test.js:149` — same
   - `src/components/stops/__tests__/StopPageHeader.test.js:164` — same
   - `src/components/stops/__tests__/StopPageHeader.test.js:193` / `:196` — same

   And these need to be removed/reverted:
   - Line ~152: `expect(arrivalsTab).toHaveAttribute('aria-selected', 'true')` — that attribute is no longer set.
   - Lines ~234–243: the rewrite around `allLinks.toHaveLength(1)` and `getAllByRole('tab').toHaveLength(2)` needs to go back to expecting 3 links total (back-to-map + 2 tab links).

   Please run `npx vitest run` locally before pushing — a green local run would have caught this.

## Important

2. **`OptionsPill` is keyboard-inaccessible when clickable.** In `src/components/trip-planner/OptionsPill.svelte:5–17`, you've added `role="button"` + `onclick` to a `<span>`, but there's no `tabindex="0"` and no `onkeydown` for Enter/Space. A keyboard-only user can't focus or activate it. Svelte's a11y compiler usually warns on this — if it didn't here, it's only because `onclick` is conditional. Either:
   - Add `tabindex={onclick ? 0 : undefined}` and an `onkeydown` handler that fires `onclick` on Enter/Space, **or**
   - Render a real `<button>` when `onclick` is provided (much simpler, and gets keyboard support for free).

## Fit and Finish

3. **The new props on `CompassArrow`, `FullPageLoadingSpinner`, and `OptionsPill` have no production consumers.** I should have caught this in the first review — apologies for the late flag. Specifically:
   - `CompassArrow.svelte` adds a `heading` prop (degrees-based rotation); the only callsite, `StopPageHeader.svelte:37`, still uses `stopDirection`.
   - `FullPageLoadingSpinner.svelte` adds a `visible` prop; the only callsite, `MapContainer.svelte:35`, controls visibility via an outer `{#if}` and doesn't pass `visible`.
   - `OptionsPill.svelte` adds `active` and `onclick`; all four callsites in `TripPlan.svelte:380–392` pass only `icon` and `label`.

   These props exist solely to make the tests do more, which inverts the relationship (tests should exercise real behavior, not motivate new API surface). Please either wire them up to a real consumer, or drop them and shrink the tests to cover what the components actually do today. My preference is the latter — keep this PR focused on #402.

4. **The `svelte-i18n` mock in `FullPageLoadingSpinner.test.js:7–46` is heavier than it needs to be.** `vitest-setup.js` already mocks `svelte-i18n`; the 40-line custom mock with subscriber tracking and locale-switching is only there so the last test can re-render on `locale.set('es')`. If you keep that test, a much smaller mock that exposes a writable `t` derived from a locale store would do the job. If you drop the unused `visible` prop (per item 3) and the locale-switch test, you can probably delete the custom mock entirely and rely on the global one.

5. **The new comment in `StopPageHeader.test.js:166–167`** ("Note: $app/stores is statically mocked...") points at a test that doesn't actually verify the thing its name claims to verify. Better to make the page mock dynamic so `highlights schedule tab when on schedule page` actually asserts `aria-selected`/`current` — or delete the test, since right now it just asserts `toBeInTheDocument()` on an element you already know is there.

## Strengths

- Polyline animation cleanly excised — `git diff develop` confirms `OpenStreetMapProvider.svelte.js` and `RouteMap.svelte` are untouched. 
- Test files are in their co-located `__tests__/` directories per project convention.
- `StopModal` integration test stabilization (fetch stub + `userEvent.setup({ delay: null })` + `waitFor`) is the right pattern, and the 5s/2s timeouts are appropriately tight.
- `FontAwesomeIconMock.svelte` is a nice small primitive — passing `class` through so callers can assert it is exactly the right shape.
- The `expectTimeEqual` / `normalizeMeridiem` helper in `dateTimeFormat.test.js` is a clean way to absorb ICU's `am`/`AM` variance without weakening the assertions.

## Recommended Action

1. Fix the failing `StopPageHeader.test.js` assertions (item 1) and re-run `npx vitest run` to confirm green.
2. Address keyboard a11y on `OptionsPill` (item 2).
3. Decide on item 3 — I'd vote to remove the unused props.
4. Sweep up the smaller polish items (4, 5).
5. Push and I'll re-review.

Verdict: **Request Changes**

Thanks again — this is close, just needs that last test run to be honest about what changed.

aaronbrethorst:

PR #414, opened 2026-03-21
Bookmarks

Bruce-Mahagwa:
- Added a bookmarks tab that allows a user to browse all saved routes. They are saved in local storage.
- The user can delete any bookmark.
- The user can also view a route. When a saved bookmark is viewed it automatically redirects to the home page and the route's name is populated in the search field. The results are then fetched.
- The user can save routes from various pages in the app.

https://github.com/user-attachments/assets/fcbc966a-94b7-461f-89c3-7038ee998454

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,5 +70,3 @@ PUBLIC_LANGUAGE_SWITCHER_ENABLED=true
 PUBLIC_LANGUAGE_SWITCHER_BUTTON_FORMAT=native
 # Menu format (default: "native-english")
 PUBLIC_LANGUAGE_SWITCHER_MENU_FORMAT=native-english
-#Storage key for the user's bookmarks in localStorage
-STORAGE_KEY_BOOKMARKS="bookmarks"
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@ PRIVATE_OBACO_SHOW_TEST_ALERTS=false
 # Example: PRIVATE_OBA_AGENCY_FILTER=19,29
 PRIVATE_OBA_AGENCY_FILTER=
 
-PUBLIC_NAV_BAR_LINKS={"Home": "/","About": "/about","Contact": "/contact","Fares & Tolls": "/fares-and-tolls"}
+PUBLIC_NAV_BAR_LINKS={"Home": "/","About": "/about","Contact": "/contact","Fares & Tolls": "/fares-and-tolls", "Bookmarks": "/bookmarks"}
 PUBLIC_OBA_GOOGLE_MAPS_API_KEY=""
 PUBLIC_OBA_LOGO_URL="https://onebusaway.org/wp-content/uploads/oba_logo-1.png"
 OBA_LOGO_URL_DARK=""
@@ -70,3 +70,5 @@ PUBLIC_LANGUAGE_SWITCHER_ENABLED=true
 PUBLIC_LANGUAGE_SWITCHER_BUTTON_FORMAT=native
 # Menu format (default: "native-english")
 PUBLIC_LANGUAGE_SWITCHER_MENU_FORMAT=native-english
+#Storage key for the user's bookmarks in localStorage
+STORAGE_KEY_BOOKMARKS="bookmarks"
--- a/src/components/ArrivalDeparture.svelte
+++ b/src/components/ArrivalDeparture.svelte
@@ -1,6 +1,7 @@
 <script>
 	import { t } from 'svelte-i18n';
 	import { msToLocalArrivalDepartureTimeString } from '$lib/dateTimeFormat';
+	import { bookmarks } from '$src/stores/bookmarksStore';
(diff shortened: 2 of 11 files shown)


aaronbrethorst (changes requested):
Hey Bruce, great initiative building out a bookmarks feature — the overall flow of saving a route and navigating back to search from a bookmark is a solid concept. There are a number of issues to work through before this is ready to merge, though. Let's dig in.

### Critical Issues (4 found)

1. **Dead code file: `src/lib/utils/bookmarks.js`** — This file imports `STORAGE_KEY_BOOKMARKS` from `$env/static/public`, but that env var doesn't exist anywhere (not in `.env.example`, not in `env-schema.json`). Nothing in the PR imports from this file — all bookmark logic goes through `bookmarksStore.js`. This file will cause a build failure if anything ever imports it. It should be deleted entirely.

2. **No internationalization (i18n)** — This project uses `svelte-i18n` for all user-facing strings (see `src/locales/en.json`). Every hardcoded English string in this PR needs to use `$_()` or `$t()` translation calls:
   - `"Save Route"`, `"Saved"` (`ArrivalDeparture.svelte:247`)
   - `"Save Route"` (`RouteModal.svelte:75`)
   - `"Bookmarks"`, `"Loading bookmarks..."`, `"No bookmarks saved yet."`, `"Delete Route"`, `"View Route"` (`bookmarks/+page.svelte`)

   Add corresponding keys to `src/locales/en.json`.

3. **No dark mode support on bookmarks page** — The bookmarks page uses `bg-gray-100` without dark variants, and text elements lack `dark:text-white` classes. In dark mode, this page will look broken. Follow existing patterns like `text-black dark:text-white` and `bg-gray-100 dark:bg-gray-800`.

4. **Formatting failures** — 5 of the changed files fail Prettier checks (`npm run lint`). Run `npm run format` before pushing. The PR also introduces trailing whitespace changes in `TripDetailsPane.svelte`, `SearchPane.svelte`, and `+page.svelte` that add noise to the diff without changing behavior — these should be reverted.

### Important Issues (5 found)

1. **`isSaved` state doesn't reflect actual store contents** — `ArrivalDeparture.svelte:214` — The `isSaved` flag is component-local state initialized to `false`. If the route is already bookmarked, the button still shows "Save Route". It should be derived from the store:
   ```js
   let isSaved = $derived(bookmarks.getOne(arrivalDeparture.routeShortName) !== null);
   ```
   (Though `getOne` also needs a `browser` guard — see below.)

2. **`getOne()` has no SSR guard** — `bookmarksStore.js:41-44` — Every other method checks `if (browser)` before accessing `localStorage`, but `getOne` calls `localStorage.getItem()` directly. This will throw during server-side rendering.

3. **Bookmark key is route short name — not unique** — Bookmarks are keyed by `routeName` (which is the route's `shortName`). Route short names aren't guaranteed unique across agencies in a multi-agency region. Two different routes named "1" from different agencies would collide. Consider using the route ID instead.

4. **`SearchField.svelte` — `$page.state` read at module top level isn't reactive** — `SearchField.svelte:13-14` — `const routeNameFromBookmarks = $page.state?.name` is evaluated once when the component mounts. If `$page.state` changes later (e.g., navigating back from bookmarks a second time), the value won't update. Use `$derived` instead:
   ```js
   let routeNameFromBookmarks = $derived($page.state?.name);
   let previousPage = $derived($page.state?.previousPage);
   ```

5. **No associated issue** — This feature doesn't reference an open issue. For a feature of this scope (new page route, new store, UI across multiple components), it would be good to have an issue first to align on the design, UX, and scope before implementation.

### Suggestions (4 found)

1. **Button styling** — The `bg-black` buttons don't use the project's brand color system (`text-brand`, `bg-brand`, etc.). Consider using brand colors for consistency, or at minimum add dark mode variants.

2. **Missing `<svelte:head>` on bookmarks page** — Other pages set page titles via `<svelte:head><title>`. The bookmarks page should too.

3. **No confirmation on delete** — Clicking "Delete Route" immediately removes the bookmark with no undo or confirmation. A brief toast or undo option would improve UX.

4. **Missing newline at end of file** — Both `bookmarksStore.js` and `bookmarks.js` are missing a trailing newline.

### Strengths

- The `createBookmarkStore()` pattern in `bookmarksStore.js` is well-structured and follows the project's existing store patterns nicely
- Good use of `$app/environment`'s `browser` check for SSR safety (in most methods)
- The "View Route" flow using `goto()` with state to trigger a search is a creative approach

## Recommended Action
1. Delete the dead `src/lib/utils/bookmarks.js` file
2. Add i18n support for all user-facing strings
3. Add dark mode classes to the bookmarks page
4. Run `npm run format` and revert whitespace-only changes
5. Fix the `isSaved` state, `getOne` SSR guard, and `$page.state` reactivity issues
6. Consider using route ID as the bookmark key instead of short name

### Verdict: **Request Changes**

aaronbrethorst:
Closing: please feel free to reopen after you've updated the PR!

PR #417, opened 2026-03-22
fix/use correct type declaration for props with any types - phase 1

### LINKED ISSUE #404: Replace `any` types with proper interfaces in component props
## What needs to change

Several components use \`@property {any}\` for their props, which means we lose type safety and editor autocomplete. These should be replaced with descriptive types or interfaces.

## Example files

- \`src/components/map/StopMarker.svelte\` — \`stop\`, \`onClick\`, \`icon\` are all \`any\`
- \`src/components/stops/StopPane.svelte\` — \`stop\` and \`arrivalsAndDeparturesResponse\` are \`any\`
- \`src/components/tabs/TabLink.svelte\` — \`href\` is \`any\` (should be \`string\`)
- \`src/components/containers/Accordion.svelte\` — \`items\` is \`any\`

## What to do

1. Look at how the props are actually used in each component to understand their shape
2. Look at the API responses in \`src/lib/\` or \`src/routes/api/\` to see what data structures look like
3. Replace \`{any}\` with the correct type (e.g., \`{string}\`, \`{Function}\`, \`{Object}\`) or create a JSDoc \`@typedef\` for more complex shapes

You don't need to fix all of them in one PR — even replacing a few is valuable!

## Acceptance criteria

- [ ] At least 2-3 components have their \`any\` props replaced with meaningful types
- [ ] The types accurately reflect how the props are used
- [ ] \`npm run lint\` passes
- [ ] \`npm run test\` passes

Bruce-Mahagwa:
Closes #404 
- Corrected 'any' type declarations for props in the following components:
1. src/components/containers/Accordion.svelte
2. src/components/map/StopMarker.svelte
3. src/components/stops/StopPane.svelte
4. src/components/tabs/TabLink.svelte

### DIFF
--- a/src/components/containers/Accordion.svelte
+++ b/src/components/containers/Accordion.svelte
@@ -10,7 +10,7 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {string} [items] - Props to track all possible item IDs
+	 * @property {string[]} [items] - Props to track all possible item IDs
 	 * @property {import('svelte').Snippet} [children]
 	 */
 
--- a/src/components/containers/Accordion.svelte
+++ b/src/components/containers/Accordion.svelte
@@ -10,13 +10,12 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {any} [items] - Props to track all possible item IDs
+	 * @property {string} [items] - Props to track all possible item IDs
 	 * @property {import('svelte').Snippet} [children]
 	 */
 
 	/** @type {Props} */
 	let { items = $bindable([]), children } = $props();
-
 	// Methods to open/close all items
 	export const openAll = (animate = true) => {
 		skipAnimation.set(!animate);
--- a/src/components/map/StopMarker.svelte
+++ b/src/components/map/StopMarker.svelte
@@ -4,23 +4,23 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {{
-	 *   id: string,
-	 *   name: string,
-	 *   lat: number,
-	 *   lon: number,
-	 *   direction: string,
-	 *   routes: {
-	 *     id: string,
-	 *     shortName: string,
-	 *     type: number
-	 *   }[]
-	 * }} stop
+	 * @property {(
+	 *   import('onebusaway-sdk/resources/stop').StopRetrieveResponse & {
+	 *     data: import('onebusaway-sdk/resources/stop').StopRetrieveResponse.Data & {
+	 *       entry: import('onebusaway-sdk/resources/stop').StopRetrieveResponse.Data.Entry & {
+	 *         routes: (import('onebusaway-sdk/resources/shared').References.Route & {
+	 *           code?: string
+	 *         })[]
+	 *       }
+	 *     }
+	 *   }
+	 * )} stop
 	 * @property {() => void} onClick
 	 * @property {import('@fortawesome/fontawesome-svg-core').IconDefinition} icon
 	 * @property {boolean} [isHighlighted]
 	 * @property {boolean} [showRoutesLabel]
 	 */
+
 	/** @type {Props} */
 	let { stop, onClick, icon, isHighlighted = false, showRoutesLabel = false } = $props();
 
--- a/src/components/map/StopMarker.svelte
+++ b/src/components/map/StopMarker.svelte
@@ -4,13 +4,23 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {any} stop
-	 * @property {any} onClick
-	 * @property {any} icon
+	 * @property {{
+	 *   id: string,
+	 *   name: string,
+	 *   lat: number,
+	 *   lon: number,
+	 *   direction: string,
+	 *   routes: {
+	 *     id: string,
+	 *     shortName: string,
+	 *     type: number
+	 *   }[]
+	 * }} stop
+	 * @property {() => void} onClick
+	 * @property {import('@fortawesome/fontawesome-svg-core').IconDefinition} icon
 	 * @property {boolean} [isHighlighted]
 	 * @property {boolean} [showRoutesLabel]
 	 */
-
 	/** @type {Props} */
 	let { stop, onClick, icon, isHighlighted = false, showRoutesLabel = false } = $props();
 
--- a/src/components/stops/StopPane.svelte
+++ b/src/components/stops/StopPane.svelte
@@ -19,8 +19,77 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {any} stop
-	 * @property {any} [arrivalsAndDeparturesResponse]
+	 * @property {{
+	 *   id: string,
+	 *   name: string,
+	 *   lat: number,
+	 *   lon: number,
+	 *   direction: string,
+	 *   routeIds: string[],
+	 *   routes: {
+	 *     id: string,
+	 *     shortName: string,
+	 *     type: number
+	 *   }[]
+	 * }} stop
+	 * @property {{
+	 *   entry: {
+	 *     stopId: string,
+	 *     arrivalsAndDepartures: {
+	 *       routeId: string,
+	 *       routeShortName: string,
+	 *       routeLongName: string,
+	 *       tripId: string,
+	 *       tripHeadsign: string,
+	 *       serviceDate: number,
+	 *       predicted: boolean,
+	 *       scheduleDeviation: number,
+	 *       distanceFromStop: number,
+	 *       numberOfStopsAway: number,
+	 *       tripStatus: {
+	 *         activeTripId: string,
+	 *         vehicleId: string,
+	 *         position: { lat: number, lon: number },
+	 *         orientation: number,
+	 *         status: string
+	 *       } | null,
+	 *       frequency: any | null,
+	 *       predictedArrivalTime: number | null,
+	 *       scheduledArrivalTime: number,
+	 *       predictedDepartureTime: number | null,
+	 *       scheduledDepartureTime: number,
+	 *       status: string,
+	 *       lastUpdateTime: number,
+	 *       blockTripSequence: number,
+	 *       vehicleId: string | null
+	 *     }[],
+	 *   },
+	 *   references: {
+	 *     routes: {
+	 *       id: string,
+	 *       nullSafeShortName: string,
+	 *       shortName: string,
+	 *       longName: string,
+	 *       description: string,
+	 *       type: number,
+	 *       url: string,
+	 *       color: string,
+	 *       textColor: string
+	 *     }[],
+	 *     situations: {
+	 *       id: string,
+	 *       summary: string,
+	 *       description: string,
+	 *       severity: string,
+	 *       reason: string,
+	 *       effect: string,
+	 *       cause: string,
+	 *       url: string,
+	 *       activeWindows: { from: number, to: number }[],
+	 *       informedEntities: { agencyId: string, routeId: string | null, stopId: string | null, tripId: string | null }[]
+	 *     }[]
+	 *   }
+	 * }} [arrivalsAndDeparturesResponse]
 	 */
 
 	/** @type {Props} */
--- a/src/components/stops/StopPane.svelte
+++ b/src/components/stops/StopPane.svelte
@@ -19,77 +19,10 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {{
-	 *   id: string,
-	 *   name: string,
-	 *   lat: number,
-	 *   lon: number,
-	 *   direction: string,
-	 *   routeIds: string[],
-	 *   routes: {
-	 *     id: string,
-	 *     shortName: string,
-	 *     type: number
-	 *   }[]
-	 * }} stop
-	 * @property {{
-	 *   entry: {
-	 *     stopId: string,
-	 *     arrivalsAndDepartures: {
-	 *       routeId: string,
-	 *       routeShortName: string,
-	 *       routeLongName: string,
-	 *       tripId: string,
-	 *       tripHeadsign: string,
-	 *       serviceDate: number,
-	 *       predicted: boolean,
-	 *       scheduleDeviation: number,
-	 *       distanceFromStop: number,
-	 *       numberOfStopsAway: number,
-	 *       tripStatus: {
-	 *         activeTripId: string,
-	 *         vehicleId: string,
-	 *         position: { lat: number, lon: number },
-	 *         orientation: number,
-	 *         status: string
-	 *       } | null,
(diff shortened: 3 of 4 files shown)


aaronbrethorst (commented):
## PR Review: Replace `any` type declarations with proper JSDoc types

Thanks for working on eliminating `any` types -- this is a valuable effort. I have found several issues that need to be addressed, ranging from a clear bug to architectural concerns.

---

### 1. Bug: `Accordion.svelte` -- `items` typed as `{string}` but used as an array

**File:** `src/components/containers/Accordion.svelte`, line 13

The type was changed from `{any}` to `{string}`, but `items` is unambiguously an array:
- Initialized as `$bindable([])`
- Called with `items.includes(id)` (line 38)
- Spread with `[...items, id]` (implicit in `registerItem`)
- Passed to `new Set(items)` (line 22)

The correct type is `{string[]}`.

---

### 2. Major concern: `StopPane.svelte` -- ~70-line inline type is the wrong approach

**File:** `src/components/stops/StopPane.svelte`, lines 22-92

The `onebusaway-sdk` package already ships comprehensive TypeScript definitions for every API response shape. The types you are hand-writing already exist at:

- `onebusaway-sdk/resources/arrival-and-departure.d.ts` → `ArrivalAndDepartureListResponse`
- `onebusaway-sdk/resources/stop.d.ts` → `StopRetrieveResponse`
- `onebusaway-sdk/resources/shared.d.ts` → `References`, `References.Route`, `References.Situation`, etc.

Hand-duplicating these creates several problems:
- **Drift risk:** The inline types will silently diverge from the SDK as it updates. For example, the SDK's `Situation` type has `summary` and `description` as nested objects (`{ lang?: string, value?: string }`), not plain strings as declared in this PR.
- **Incomplete coverage:** The inline type omits many fields the SDK defines (e.g., `arrivalEnabled`, `departureEnabled`, `stopSequence`, `totalStopsInTrip` on arrivals; `code`, `parent`, `staticRouteIds`, `wheelchairBoarding` on stops).
- **Maintenance burden:** 70 lines of inline JSDoc in a component's `<script>` block is difficult to read and maintain.

**However**, there is a structural mismatch that makes direct SDK type reuse non-trivial: the component fetches from `/api/oba/arrivals-and-departures-for-stop/`, which returns the *entire* SDK response (including the `code`, `currentTime`, `text`, `version`, `data` wrapper). The prop type in this PR starts at the `data` level, but the component actually accesses `arrivalsAndDeparturesResponse.data.references.routes` on line 165-166, meaning the prop contains the full response wrapper, not just `data`. The type as written would be wrong even structurally.

**Recommendation:** Rather than inlining these types, consider one of:
1. Import and reference the SDK types via JSDoc: `@property {import('onebusaway-sdk/resources/arrival-and-departure').ArrivalAndDepartureListResponse} [arrivalsAndDeparturesResponse]`
2. Or, if the SDK import paths don't resolve cleanly in JSDoc, create a single shared `src/lib/types.js` (or `.d.ts`) file that re-exports or aliases the SDK types, similar to the pattern already used in `src/lib/serverCache.js` with its `@typedef` declarations.

---

### 3. `StopPane.svelte` -- `frequency: any | null` still uses `any`

**File:** `src/components/stops/StopPane.svelte`, line 56

This defeats the purpose of the PR. The SDK types `frequency` as `string | undefined`. The correct JSDoc type is `{string} [frequency]` (optional string).

---

### 4. `StopPane.svelte` -- Props typedef is incomplete

**File:** `src/components/stops/StopPane.svelte`, lines 96-101

The `Props` typedef only declares `stop` and `arrivalsAndDeparturesResponse`, but the destructuring on lines 96-101 also extracts `handleUpdateRouteMap` and `tripSelected`. This means TypeScript/JSDoc checking would flag those as invalid properties. While this is a pre-existing issue, a PR specifically about correcting types should either fix it or at minimum note it as out-of-scope.

The types for those would be:
```
@property {((event: {detail: any}) => void) | null} [handleUpdateRouteMap]
@property {((event: {detail: any}) => void) | null} [tripSelected]
```

---

### 5. `StopMarker.svelte` -- `stop` type is missing fields the component actually uses

**File:** `src/components/map/StopMarker.svelte`, line 32

The component accesses `r?.code` on line 32 in the route name derivation:
```js
.map((r) => r?.shortName || r?.code || (r?.id ? String(r.id).split('_').pop() : null))
```

But the route type in the PR only declares `{ id: string, shortName: string, type: number }` -- it omits `code`. The SDK's `References.Route` type does not include `code` either, so this is also a pre-existing data concern, but the new type should at minimum not be *more* restrictive than what the code actually touches. Consider adding `code?: string` to the route sub-type.

Similarly, the `stop` type in the SDK includes `code?: string`, `parent: string`, `routeIds: string[]`, and `staticRouteIds: string[]` -- the inline type only captures a subset.

---

### 6. Minor: Deleted blank lines

The diff removes blank lines between the typedef and the `let` declaration in Accordion.svelte (line 18-19) and StopMarker.svelte (line 23-24). These are cosmetic but should be consistent with the rest of the codebase. Most other components have a blank line separating the typedef from the destructuring.

---

### Summary of requested changes

| File | Severity | Issue |
|------|----------|-------|
| `Accordion.svelte` | **Bug** | `{string}` must be `{string[]}` |
| `StopPane.svelte` | **Architecture** | 70-line inline type should use SDK types or a shared typedef file |
| `StopPane.svelte` | **Bug** | Type starts at `data` level but prop holds full response (with `code`, `data`, etc.) |
| `StopPane.svelte` | **Incomplete** | `frequency: any \| null` still uses `any` |
| `StopPane.svelte` | **Incomplete** | `handleUpdateRouteMap` and `tripSelected` missing from Props |
| `StopMarker.svelte` | **Incomplete** | Route sub-type missing `code` field that the component accesses |

The `TabLink.svelte` change (`any` to `string`) is correct and looks good.

I'd recommend fixing the Accordion bug and reconsidering the approach for StopPane before merging. The direction of this PR is right -- eliminating `any` is important work -- but the replacement types need to be accurate and maintainable.

aaronbrethorst (changes requested):
Hey Bruce, glad to see work on tightening up the types in this codebase -- it's the kind of unglamorous work that pays dividends down the road. I found a few issues with the types as written that need to be addressed.

## Critical Issues (2 found)

### 1. `items` in Accordion.svelte is typed as `{string}` but used as an array

The prop is initialized with `$bindable([])`, iterated with `.includes()`, spread with `[...items, id]`, and passed to `new Set(items)`. The type must be `{string[]}`, not `{string}`.

```diff
- * @property {string} [items] - Props to track all possible item IDs
+ * @property {string[]} [items] - Props to track all possible item IDs
```

`Accordion.svelte:13`

### 2. `arrivalsAndDeparturesResponse` type doesn't match actual usage in StopPane.svelte

The inline type starts at the `data` level (with `entry` and `references` as top-level keys), but the component accesses `arrivalsAndDeparturesResponse.data.references.routes` on line 165. The actual prop holds the full OBA response wrapper which includes `code`, `currentTime`, `text`, `version`, and `data`. As written, the type would cause type errors at every call site.

`StopPane.svelte:35-92`

## Important Issues (3 found)

### 3. 70-line inline JSDoc duplicates (and diverges from) existing SDK types

The `onebusaway-sdk` package already ships TypeScript declarations for these response shapes in `node_modules/onebusaway-sdk/resources/arrival-and-departure.d.ts` and `shared.d.ts`. The hand-written types are already drifting -- for example, the SDK defines `Situation.summary` and `Situation.description` as `{ lang?: string, value?: string }` objects, not plain strings.

Rather than hand-maintaining 70 lines of inline JSDoc, consider either:
- Importing the SDK types via JSDoc (`@type {import('onebusaway-sdk/resources/...').SomeType}`)
- Or creating a shared `src/lib/types.js` with `@typedef` definitions that can be reused across components

This would also make the types more maintainable as the API evolves.

`StopPane.svelte:35-92`

### 4. `frequency: any | null` still uses `any`

The whole purpose of this PR is to replace `any` types, but one slipped through. The SDK types `frequency` as `string | undefined`.

`StopPane.svelte:56`

### 5. Missing `code` field in StopMarker route type

The component accesses `r?.code` on line 32 in the route name derivation, but the new type for routes omits the `code` property:

```javascript
.map((r) => r?.shortName || r?.code || ...)
```

Add `code?: string` to the route sub-type.

`StopMarker.svelte:7-17, 32`

## Suggestions (1 found)

- The Props typedef in `StopPane.svelte` only declares `stop` and `arrivalsAndDeparturesResponse`, but the destructuring on lines 98-99 also pulls `handleUpdateRouteMap` and `tripSelected`. While pre-existing, if you're already touching this typedef it would be good to add those too. `StopPane.svelte:96-101`

## Strengths

- Correct and clean types for `TabLink.svelte` (`href: string`) and `StopMarker.svelte` (`onClick: () => void`, icon using the FontAwesome import type).
- Good initiative tackling the `any` type cleanup across the codebase.

## Recommended Action

1. Fix the `string` → `string[]` bug in Accordion.
2. Fix the response wrapper mismatch in StopPane (the type needs a `data` wrapper, or the access pattern needs to change).
3. Address the leftover `any` on `frequency`.
4. Add the missing `code` field to StopMarker's route type.
5. Consider whether the 70-line inline type is the right approach vs. importing from the SDK or extracting shared types.

**Verdict: Request Changes**

aaronbrethorst (changes requested):
Bruce -- thanks for the follow-up commit. The Accordion fix, the `frequency: any | null` cleanup, the new `code?: string` on the route sub-type, and the addition of `handleUpdateRouteMap` / `tripSelected` to the StopPane Props are all on the money. The pivot away from a hand-rolled 70-line inline type to `import('onebusaway-sdk/...')` is exactly the direction I was hoping for.

That said, in the process of switching to SDK type imports, the same structural mismatch I flagged before on `arrivalsAndDeparturesResponse` has now been reintroduced in two new places: the `stop` props in both `StopPane.svelte` and `StopMarker.svelte` are typed as the full SDK *response* shape, but the actual values held by those props are nested *entries* inside that response. The type would have the consumer reaching for `stop.data.entry.name`, while the code uniformly accesses `stop.name`.

## Critical Issues (2 found)

### 1. `StopPane.svelte` -- `stop` prop is typed as the full response, but holds the entry only

**File:** `src/components/stops/StopPane.svelte:22`

The type is:
```js
@property {import('onebusaway-sdk/resources/stop').StopRetrieveResponse} stop
```

But `StopRetrieveResponse` is defined in the SDK as:

```ts
interface StopRetrieveResponse extends Shared.ResponseWrapper {
  data: { entry: { id, lat, lon, name, routeIds, ... }, references: ... }
}
```

The actual value passed in is the **entry**, not the full response. From `src/routes/stops/[stopID]/+page.svelte:15`:

```js
const stop = data.stopData.entry;
// ...
<StopPane {stop} ... />
```

And the component accesses fields directly off `stop`:
- `stop?.id` (line 87)
- `stop.routeIds.includes(...)` (line 100)
- `stop.id`, `stop.lat`, `stop.lon` (lines 140-142)
- `stop.name`, `stop.id` (lines 187-189)

Under the declared type, every one of these accesses would be a type error -- `id`/`name`/`lat`/`lon`/`routeIds` are not on `StopRetrieveResponse`, they're on `StopRetrieveResponse.Data.Entry`.

**Fix:**
```js
@property {import('onebusaway-sdk/resources/stop').StopRetrieveResponse.Data.Entry} stop
```

### 2. `StopMarker.svelte` -- same problem, plus the prop comes from a *different* endpoint

**File:** `src/components/map/StopMarker.svelte:7-17`

The type is:
```js
@property {(
  import('onebusaway-sdk/resources/stop').StopRetrieveResponse & {
    data: ... & {
      entry: ... & {
        routes: (References.Route & { code?: string })[]
      }
    }
  }
)} stop
```

Two problems here:

**(a) The wrapper level is wrong.** Same as #1 -- the component does `stop?.routes`, `stop.name`, `stop.direction`, `stop.id` directly. None of these would resolve under a type rooted at `StopRetrieveResponse`.

**(b) The SDK type chosen is for the wrong endpoint.** This marker's `stop` does not come from `/api/oba/stop/{id}` -- it comes from `/api/oba/stops-for-location`, then has a `routes` field grafted on client-side. Look at `MapView.svelte:185-199`:

```js
const newStops = stopsData.data.list;       // an array of StopsForLocationListResponse.Data.List items
// ...
newStops.forEach((stop) => {
  stop.routes = stop.routeIds?.map((routeId) => routeLookup.get(routeId)).filter(Boolean) || [];
  // ...
});
// stops then flow through `addMarker({ stop: s, ... })`
```

So the actual shape is `StopsForLocationListResponse.Data.List & { routes: References.Route[] }`, not `StopRetrieveResponse`. (Both endpoints happen to return entries with the same fields, but conceptually -- and for SDK-version drift safety -- the type should reflect where the data actually came from.)

**Fix:**
```js
@property {(
  import('onebusaway-sdk/resources/stops-for-location').StopsForLocationListResponse.Data.List & {
    routes: (import('onebusaway-sdk/resources/shared').References.Route & { code?: string })[]
  }
)} stop
```

## Suggestions (1 found)

- `StopPane.svelte:24-25` still uses `any` inside the event detail types: `{detail: any}`. I'd previously suggested this form as a quick-and-dirty fix, but now that you're properly importing SDK types, it'd be cleaner to type these too. `handleUpdateRouteMap`'s detail is `{ show: boolean }`; `tripSelected`'s detail is the accordion's `activeData`, which holds a `Data.Entry.ArrivalsAndDeparture` from the arrivals SDK type. Worth a follow-up if you'd rather keep this PR tight.

## Strengths

- Accordion `string` -> `string[]` fix is exactly right.
- `frequency: any | null` is gone -- nice.
- `code?: string` on the route sub-type covers the actual access at `StopMarker.svelte:32`.
- Adding `handleUpdateRouteMap` and `tripSelected` to the StopPane Props closes the destructuring gap.
- Moving from 70 lines of inline JSDoc to SDK imports is the right architectural call -- you've made the types track upstream automatically.
- `TabLink.svelte` -- still clean.

## Recommended Action

1. Fix the `stop` prop type in `StopPane.svelte:22` -- use `StopRetrieveResponse.Data.Entry`.
2. Fix the `stop` prop type in `StopMarker.svelte:7-17` -- use `StopsForLocationListResponse.Data.List` (with the route augmentation).
3. Optionally tighten the `{detail: any}` types on the StopPane handlers, or leave that for a phase-2 PR.

The principle to remember -- and one I should have stated more plainly the first time around -- is: **the JSDoc type should describe what the prop actually holds at the call site, which may be a deeply nested slice of the SDK's response type, not the response type itself.** When in doubt, trace upward from the field accesses in the component (`stop.name`, `stop.id`) and pick the narrowest SDK type that has those fields.

**Verdict: Request Changes**

aaronbrethorst:

PR #418, opened 2026-03-22
fix(vehicle): correct CANCELED check and guard trips-for-route fetch

mukesharyal:
OBA exposes the trip state string on status.status, not the status object. Return empty references/list when the HTTP response fails or the payload is incomplete so the 30s poller does not throw on data.references.

### DIFF
--- a/src/lib/vehicleUtils.js
+++ b/src/lib/vehicleUtils.js
@@ -15,13 +15,14 @@ const vehicleMarkersMap = new Map();
 
 export async function fetchVehicles(routeId) {
 	const response = await fetch(`/api/oba/trips-for-route/${routeId}`);
-	const responseBody = await response.json();
 	if (!response.ok) {
 		console.warn('fetchVehicles: request failed', routeId, response.status);
 		return { references: { trips: [] }, list: [] };
 	}
+	const responseBody = await response.json();
 	const data = responseBody.data;
 	if (!data?.references?.trips || !Array.isArray(data.list)) {
+		console.warn('fetchVehicles: unexpected response structure for route', routeId);
 		return { references: { trips: [] }, list: [] };
 	}
 	return data;
@@ -43,7 +44,7 @@ export async function updateVehicleMarkers(routeId, mapProvider, routeType) {
 		const activeTrip = activeTripMap.get(activeTripId);
 
 		// OBA puts the trip state string on status.status (e.g. SCHEDULED, CANCELED), not on status itself
-		if (activeTrip && activeTrip?.routeId === routeId && tripStatus.status?.status !== 'CANCELED') {
+		if (activeTrip && activeTrip.routeId === routeId && tripStatus.status?.status !== 'CANCELED') {
 			const vehicleStatus = tripStatus.status;
 
 			activeTripIds.add(activeTripId);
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Hey Mukesh, thanks for digging into this bug — the `tripStatus.status !== 'CANCELED'` fix is a great catch. Comparing an object to a string meant canceled trips were never filtered, and the comment you added explaining the double `.status` pattern is exactly the kind of documentation that prevents future regressions. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Move `response.json()` after the `response.ok` check.** Right now (line 18-19), `response.json()` is called *before* `response.ok` is checked. If the server returns a non-JSON error body (HTML 502 page, empty body, plain-text error), `response.json()` will throw a `SyntaxError`, and your `response.ok` guard on line 19 is never reached. The fix is simply to reorder:

   ```javascript
   export async function fetchVehicles(routeId) {
       const response = await fetch(`/api/oba/trips-for-route/${routeId}`);
       if (!response.ok) {
           console.warn('fetchVehicles: request failed', routeId, response.status);
           return { references: { trips: [] }, list: [] };
       }
       const responseBody = await response.json();
       const data = responseBody.data;
       if (!data?.references?.trips || !Array.isArray(data.list)) {
           return { references: { trips: [] }, list: [] };
       }
       return data;
   }
   ```

## Important

2. **Add a `console.warn` to the structural validation fallback (lines 24-26).** When `data?.references?.trips` is missing or `data.list` isn't an array, the function silently returns empty defaults with no logging. This makes it very hard to debug when the API response shape changes unexpectedly. Add a warning:

   ```javascript
   if (!data?.references?.trips || !Array.isArray(data.list)) {
       console.warn('fetchVehicles: unexpected response structure for route', routeId);
       return { references: { trips: [] }, list: [] };
   }
   ```

3. **Wrap the polling callback in a try-catch.** If `fetchVehicles` throws (network error, unexpected response), the unhandled rejection propagates from the `setInterval` callback on line 77. This means the poller keeps firing and failing every 30 seconds with no diagnostic output. Consider wrapping both calls in `fetchAndUpdateVehicles`:

   ```javascript
   export async function fetchAndUpdateVehicles(routeId, mapProvider, routeType) {
       try {
           await updateVehicleMarkers(routeId, mapProvider, routeType);
       } catch (error) {
           console.error('fetchAndUpdateVehicles: initial fetch failed', routeId, error);
       }

       return setInterval(async () => {
           try {
               await updateVehicleMarkers(routeId, mapProvider, routeType);
           } catch (error) {
               console.error('fetchAndUpdateVehicles: polling update failed', routeId, error);
           }
       }, 30000);
   }
   ```

## Fit and Finish

4. **Clean up redundant optional chaining on line 46.** The condition `activeTrip && activeTrip?.routeId === routeId` uses `?.` after already confirming `activeTrip` is truthy. Simplify to `activeTrip && activeTrip.routeId === routeId`.

aaronbrethorst (approved):
Mukesh, this is exactly the kind of clean follow-up I hoped to see. Every item from the previous round is addressed precisely as discussed, and the code reads well — the explanatory comment on line 46 clearly flags the OBA-quirk that caused the original bug, so the next person to touch this won't make the same mistake.

## Verification of Previous Feedback

1. ✅ **Critical — `response.json()` ordering fixed** (`src/lib/vehicleUtils.js:17-22`). The `response.ok` guard now correctly precedes the JSON parse, so a non-JSON error body won't blow up the poller.
2. ✅ **Important — structural validation logs a warning** (`src/lib/vehicleUtils.js:25`). When the response shape is unexpected, we now get a diagnostic rather than a silent empty return.
3. ✅ **Important — polling callback wrapped in try/catch** (`src/lib/vehicleUtils.js:75-89`). Both the initial fetch and the 30-second tick are protected, with distinct log messages so it's easy to tell them apart in the console.
4. ✅ **Fit and finish — redundant optional chaining removed** (`src/lib/vehicleUtils.js:47`). The `activeTrip.routeId === routeId` check reads cleanly now.

## Strengths

- The diagnostic messages are well-differentiated (`'request failed'`, `'unexpected response structure'`, `'initial fetch failed'`, `'polling update failed'`), which will make production debugging much easier.
- The explanatory comment on line 46 is short, accurate, and points future readers at the right thing.
- The shape returned from the early-return paths (`{ references: { trips: [] }, list: [] }`) matches the structure consumers expect, so `updateVehicleMarkers` can iterate without additional guards.

## Recommended Action

Merge.

PR #422, opened 2026-03-23
feat: localize hardcoded aria-labels in search and survey components

### LINKED ISSUE #406: Internationalize hardcoded aria-label strings
## What needs to change

A couple of components have \`aria-label\` attributes with hardcoded English text instead of using the translation system. This means screen reader users in other languages hear English labels.

## Files to update

- **\`src/components/trip-planner/TripPlanSearchField.svelte\`** (line ~54) — has \`aria-label="Clear"\`
- **\`src/components/surveys/SurveyQuestion.svelte\`** (line ~41) — has \`aria-label="Required question"\`

## What to do

1. Add translation keys to \`src/locales/en.json\` (e.g., \`"search.clear"\` and \`"survey.required_question"\`)
2. Replace the hardcoded strings with \`\$_()\` calls:

\`\`\`diff
- aria-label="Clear"
+ aria-label={$_('search.clear')}
\`\`\`

## Acceptance criteria

- [ ] Both hardcoded aria-labels are replaced with \`\$_()\` calls
- [ ] Translation keys are added to \`src/locales/en.json\`
- [ ] \`npm run lint\` passes

tijaruS:
## Summary
Fixes #406

This PR replaces hardcoded English `aria-label` strings with localized translations using the `svelte-i18n` framework. This ensures that screen reader users in other languages hear correctly translated labels instead of English defaults.

## Key Changes

- **`src/locales/en.json`**: Added new translation keys:
  - `search.clear`: `"Clear"`
  - `survey.required_question`: `"Required question"`
- **`src/components/trip-planner/TripPlanSearchField.svelte`**: Updated the clear button to use the `$t('search.clear')` translation.
- **`src/components/surveys/SurveyQuestion.svelte`**:
  - Imported the `t` store from `svelte-i18n`.

## Verification Results

- Ran `npm run format` to ensure consistent code style in JSON and Svelte files.
- Ran `npm run lint` to verify that all imports are correct and there are no illegal variable names (fixed the `$t` definition error).
- Verified that the components render correctly with the new translation keys.


aaronbrethorst (changes requested):
Hey Surajit, I appreciate you working on localizing these hardcoded aria-labels — accessibility and i18n are both important, and this PR addresses them together. I found a few things that need to be fixed before we can merge:

## Critical

1. **Typo in `en.json`: `"servey"` should be `"survey"`.** The JSON key on line 75 of `src/locales/en.json` is `"servey"`, but the component uses `$t('survey.required_question')`. This means the translation lookup will fail at runtime, and users will see the raw key string `"survey.required_question"` instead of "Required question". Fix the key to `"survey"`.

2. **3 test failures in `TripPlanSearchField.test.js`.** The tests query for `getByLabelText('Clear')`, but the clear button now uses `$t('search.clear')`. Since the i18n mock returns the key as a string, the rendered aria-label becomes `"search.clear"`. Update the 3 failing test assertions to match the new translated label. Run `npm run test` to verify all tests pass.

## Important

3. **Run Prettier.** `src/components/surveys/SurveyQuestion.svelte` fails the formatting check. The import on line 35 uses 2-space indentation instead of the project's tab indentation. Run `npm run format` to fix this.

4. **Move the import to the top of the `<script>` block.** In `SurveyQuestion.svelte`, the `import { t } from 'svelte-i18n'` is at line 35, after all the component logic. It should be at the top with the other imports (after line 2). The project convention is imports first, then component logic.

aaronbrethorst (approved):
Thanks Surajit — you've cleanly addressed every item from the last round. The `survey` typo is fixed, the import sits with its siblings at the top of the script block, formatting is clean, and the 24 tests in `TripPlanSearchField.test.js` all pass locally. Happy to merge this.

## What I verified

- `src/locales/en.json:75` — `survey.required_question` key is correctly spelled and nested at the top level alongside `search`, `header`, etc.
- `src/components/surveys/SurveyQuestion.svelte:3` — `import { t } from 'svelte-i18n'` is grouped with the other imports.
- `npm run lint` passes.
- `npx vitest run src/components/trip-planner/__tests__/TripPlanSearchField.test.js` — 24/24 passing.

## Strengths

- Tight, focused PR — exactly the scope that issue #406 asked for.
- Clean response to review: you fixed each item without scope creep.
- Sensible top-level `survey` namespace in `en.json` rather than burying the key somewhere arbitrary.

## Follow-up (tracked separately, not blocking this PR)

The updated tests assert against the raw i18n key (`'search.clear'`) rather than the translated label (`'Clear'`). The same file already shows the better pattern at lines 13-16, where the local svelte-i18n mock has a `translations` map so assertions can target user-facing strings. I've opened [issue #472](https://github.com/OneBusAway/wayfinder/issues/472) to clean this up as a separate fit-and-finish pass — not something to fix here.

## Verdict

**Merge.** Nice work cleaning this up.

PR #434, opened 2026-03-24
Add ARIA attributes to FullPageLoadingSpinner for accessibility

### LINKED ISSUE #405: Add aria-live announcements to loading spinners
## What needs to change

The loading spinner component (\`src/components/FullPageLoadingSpinner.svelte\`) doesn't announce itself to screen readers. When content is loading, screen reader users get no feedback that something is happening.

## What to do

Add ARIA attributes to the loading spinner so screen readers announce the loading state:

\`\`\`svelte
<div role="status" aria-live="polite" aria-label="Loading">
  
</div>
\`\`\`

- \`role="status"\` tells the screen reader this is a status message
- \`aria-live="polite"\` makes the screen reader announce it without interrupting
- The label text should use the \`\$_()\` translation function

## How to verify

You can test with a screen reader (VoiceOver on Mac: Cmd+F5) or just inspect the DOM to confirm the attributes are present.

## Acceptance criteria

- [ ] Loading spinner has \`role="status"\` and \`aria-live="polite"\`
- [ ] An \`aria-label\` is set using the translation function
- [ ] Translation key is added to \`src/locales/en.json\`
- [ ] \`npm run lint\` passes

Hitanshi7556:
**Accessibility improvements to the loading spinner**
Adds ARIA attributes to FullPageLoadingSpinner.svelte to make it accessible to screen reader users:

role="status" on the outer div so screen readers recognize it as a live status region
aria-live="polite" so screen readers announce the loading state without interrupting the user
aria-label={$t('loading_page')} provides a localized, human-readable label for the spinner using the existing i18n system
aria-hidden="true" on the decorative SVG so screen readers skip it and avoid reading redundant or meaningless content

**Bug fix: i18n locale normalization**
Also fixes a bug in getInitialLocale() in i18n.js where getLocaleFromNavigator() could return region-specific locale codes like en-US or zh-CN that weren't registered in svelte-i18n. When this happened, svelte-i18n would silently fail to load any locale, causing $isLoading to stay true indefinitely and the app to be permanently stuck on the loading spinner.
The fix normalizes the browser locale by first checking for an exact match, then falling back to just the language part (e.g. en-US → en), and finally defaulting to en if nothing matches.

fixes  #405

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,16 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"label": "Build",
+			"type": "shell",
+			"command": "npm",
+			"args": [
+				"run",
+				"build"
+			],
+			"problemMatcher": [],
+			"group": "build"
+		}
+	]
+}
--- a/PHASE_1_AUDIT.md
+++ b/PHASE_1_AUDIT.md
@@ -1,0 +1,183 @@
+# PHASE 1: DISCOVERY & ARCHITECTURE AUDIT
+
+**Date:** 22 March 2026  
+**Scope:** Baseline architecture, delivery-gap analysis, and readiness for implementation
+
+---
+
+## 1) EXECUTIVE SUMMARY
+
+Phase 1 is complete. The current codebase is stronger than expected and already ships key parts of the proposed GSOC scope.
+
+### Current status against target outcomes
+
+- **Favorites / bookmarks (localStorage):** **Not implemented**
+- **Service alerts display:** **Partially implemented / present in stop flow**
+- **Dynamic arrivals auto-refresh:** **Implemented** (30s polling in stop pane)
+- **Extended departures beyond default window:** **Implemented** (date-based stop schedule)
+- **Accessibility audit + fixes:** **Pending formal audit; multiple known issues**
+- **Expanded tests:** **Already strong baseline**
+- **Bug fixes (trip details / past departures):** **Needs targeted verification**
+
+---
+
+## 2) ARCHITECTURE BASELINE
+
+### Stores and state
+
+Observed store files:
+
+- `src/stores/mapStore.js`
+- `src/stores/modalOpen.js`
+- `src/stores/recentTripsStore.js` *(persisted to localStorage)*
+- `src/stores/surveyStore.js`
+- `src/stores/tripOptionsStore.js` *(persisted fields + derived `effectiveDistanceUnit`)*
+- `src/stores/userLocationStore.js`
+
+### localStorage patterns already used
+
+- `src/stores/recentTripsStore.js`
+- `src/stores/tripOptionsStore.js`
+
+These patterns are reusable for `favoritesStore` implementation.
+
+### API and data integration
+
+- OBA SDK wrapper exists in `src/lib/obaSdk.js` with central response guard (`handleOBAResponse`).
+- API routes are organized under `src/routes/api/**`.
+- Alerts data paths already exist:
+  - GTFS-RT alerts endpoint: `src/routes/api/oba/alerts/+server.js`
+  - Stop-level situations consumed in stop pane.
+
+---
+
+## 3) FEATURE GAP FINDINGS (WHAT IS ALREADY DONE VS MISSING)
+
+### A) Favorites / bookmarks
+
+**Status:** Missing (no favorites/bookmark symbols found in `src/**`).
+
+**Phase 2 target files to add:**
+
+- `src/stores/favoritesStore.js`
+- `src/lib/favoritesUtils.js`
+- `src/components/favorites/FavoriteButton.svelte`
+- `src/components/favorites/FavoritesPanel.svelte`
+
+**Likely files to modify:**
+
+- `src/components/StopItem.svelte`
+- `src/components/RouteItem.svelte`
+- `src/routes/+layout.svelte` (or another global nav host)
+
+### B) Service alerts
+
+**Status:** Present in stop experience.
+
+Evidence:
+
+- `src/components/stops/StopPane.svelte` pulls situations from arrivals response and renders `ServiceAlerts`.
+- `src/components/service-alerts/**` already exists.
+
+**Remaining work:** Improve filtering/scoping UX (route/stop relevance), strengthen a11y semantics, add focused tests.
+
+### C) Dynamic arrivals auto-refresh
+
+**Status:** Implemented.
+
+Evidence:
+
+- `src/components/stops/StopPane.svelte` uses `setInterval(..., 30000)` to re-fetch arrivals.
+- Request cancelation via `AbortController` is already in place.
+
+**Remaining work:** refine stale-state messaging and verify no timing regressions in edge cases.
+
+### D) View departures beyond default range
+
+**Status:** Implemented via date picker schedule view.
+
+Evidence:
+
+- `src/routes/stops/[stopID]/schedule/+page.svelte` fetches `/api/oba/schedule-for-stop/[id]?date=...`.
+
+**Remaining work:** evaluate adding explicit time-range controls (if required by mentors).
+
+---
+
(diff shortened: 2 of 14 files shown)


Hitanshi7556:
Done!

aaronbrethorst (changes requested):
Hi Hitanshi,

Thanks for tackling the loading spinner accessibility work — the screen-reader story here was definitely under-served and your i18n locale normalization is a nice catch that I'd want to ship on its own. Before we can land any of this, though, the PR has some structural problems that need to be sorted out first.

The biggest issue is **scope**. The title and description describe an ARIA fix plus a small i18n bug, but the diff actually bundles in a brand-new favorites feature (`favoritesStore.js`, `FavoriteButton.svelte`, integrations across four components), a 183-line `PHASE_1_AUDIT.md` planning document, and a `.vscode/tasks.json` file. That's three or four independent PRs masquerading as one, and it makes it very hard for reviewers to gate the changes. Please split this up.

## Critical Issues (4 found)

- **Duplicate attributes break the build.** `src/components/FullPageLoadingSpinner.svelte:5-12` declares `role`, `aria-live`, and `aria-label` twice each, with two different `aria-label` values (`$t('loading_page')` and `$t('loading')`). `npm run lint` fails with `attribute_duplicate: Attributes need to be unique`. Pick one of each. The cleanest fix is to keep only `role="status"`, `aria-live="polite"`, and `aria-label={$t('loading_page')}`.

- **Invalid HTML: `<button>` inside `<button>`.** `src/components/RouteItem.svelte:24-36` and `src/components/StopItem.svelte:7-23` wrap the entire row in a `<button>` and then render `<FavoriteButton>` (also a `<button>`) inside it. Nested interactive elements are invalid HTML and produce unpredictable focus/click behavior across browsers and screen readers — a particularly bad outcome on a PR whose stated goal is accessibility. The `e.stopPropagation()` in `handleToggle` only papers over the mouse-click symptom; the markup still breaks tab order and AT semantics. The row needs to stop being a `<button>` (use a `<div>` plus a separate clickable hit area), or the favorite button needs to live as a sibling outside it.

- **Hardcoded English strings throughout the favorites UI.** The whole app routes through `svelte-i18n` with 25 locales, but:
  - `src/components/favorites/FavoriteButton.svelte:5` defaults `ariaLabel = 'Add to favorites'` (hardcoded English).
  - `src/components/favorites/FavoriteButton.svelte:27` uses `title={isFavorited ? 'Remove from favorites' : 'Add to favorites'}` (hardcoded English).
  - `src/components/RouteItem.svelte:35` and `src/components/StopItem.svelte:22` pass `` `Add ${name} to favorites` `` — again hardcoded English from the parent.
  - `src/locales/en.json` adds `loading_page` but the other 24 locale files don't get a translation, so non-English users get the English string for the spinner.
  
  Add proper keys (e.g. `favorites.add`, `favorites.remove`, `favorites.add_named`) and use `$t(...)` everywhere.

- **PR scope.** Please split into three PRs: (a) the ARIA fix for `FullPageLoadingSpinner`, (b) the `getInitialLocale()` normalization (this one I'd happily land tomorrow — it's correct and self-contained), (c) the favorites feature. Also drop `PHASE_1_AUDIT.md` and `.vscode/tasks.json` from the diff (see Important below).

## Important Issues (6 found)

- **`PHASE_1_AUDIT.md` doesn't belong in the repo.** This is a 183-line project-management document with dated headings and GSOC scope notes. It's useful for you, but it's not source code and it'll rot the moment Phase 2 lands. Move it to a wiki page, an issue comment, or your own notes — but not the repo root.

- **`.vscode/tasks.json` doesn't belong in the repo either.** Editor-local config shouldn't be checked in. Either delete the file or add `.vscode/` to `.gitignore` (we already ignore `.cursor/settings.json` — same idea).

- **`FavoriteButton` uses Svelte 4 store patterns instead of Svelte 5 runes — and it has a real bug.** `src/components/favorites/FavoriteButton.svelte:7-14` does `let isFavorited = $state(false)` and then subscribes inside `onMount`. CLAUDE.md mandates Svelte 5 runes, and the idiomatic version is one line:
  ```js
  let isFavorited = $derived($favorites.some(f => f.id === id && f.type === type));
  ```
  More importantly, the current implementation captures `id` and `type` in the `onMount` closure, so if the same `FavoriteButton` instance ever gets reused with different props, it'll silently report the wrong favorite state. `$derived` re-runs reactively when props change.

- **`isFavorite` uses a subscribe-then-immediately-unsubscribe anti-pattern.** `src/stores/favoritesStore.js:81-87` does `subscribe(cb)()` to read the value synchronously. It happens to work because `writable` emits synchronously on subscribe, but it's brittle and fights the store contract. Use `import { get } from 'svelte/store'; return get({ subscribe }).some(...)` instead. Or, better, delete the method — nothing in the diff actually calls it, and reactive consumers should use `$derived`.

- **localStorage failures silently lose user data.** All five mutation paths in `favoritesStore.js` (`addFavorite`, `removeFavorite`, `toggleFavorite`, `clearAll`, and the load block) `try/catch` and `console.warn`, but the in-memory store updates regardless. Practical scenario: user is in Safari Private Browsing or has hit the storage quota, stars a stop, sees the yellow star, navigates away, comes back — favorite is gone with no feedback. At minimum the failed write should roll back the in-memory `update()` so the UI matches actual persistence; ideally there's a `persistenceError` flag the UI can surface via an `aria-live` region (which fits this branch's a11y theme nicely). I know `recentTripsStore.js` set this precedent — but this PR is a good place to introduce a better pattern, not to copy the existing one.

- **No tests for `favoritesStore.js`.** Component tests assert that `FavoriteButton` renders, but the store itself — add/remove/toggle/persistence/SSR-when-`browser`-is-false/corrupt-localStorage-data — is untested. CLAUDE.md sets a 70% coverage threshold and this is net-new logic.

## Suggestions (3 found)

- **Duplication in `favoritesStore.js`.** The `try { localStorage.setItem(...) } catch { console.warn(...) }` block appears three times (lines 34-40, 48-54, 70-76). Extract a `persist(favorites)` helper. While you're at it: `addFavorite` and `removeFavorite` aren't called from anywhere in this PR — only `toggleFavorite` is used by `FavoriteButton`. Consider trimming the API surface to just `toggle` + `clear` (plus the store contract) until something needs the others.

- **No schema validation on load.** `src/stores/favoritesStore.js:14-15` accepts anything `Array.isArray(...)` returns. If a future code change alters the favorite shape, everyone's stored list silently breaks. `recentTripsStore.js` validates entries (`.filter(t => t && t.fromCoords && ...)`) — worth copying.

- **`StopItem` markup restructuring is needlessly verbose.** `src/components/StopItem.svelte:12-22` wraps the existing two `<div>`s in an extra nested flex container just to make room for the favorite button. Once you fix the nested-button issue (Critical #2), most of this restructure goes away.

## Strengths

- The i18n locale normalization in `src/lib/i18n.js:90-110` is correct, well-commented, and addresses a real silent-failure bug. The exact-match → language-part → `'en'` cascade is the right shape. I want to merge this on its own.
- The `aria-hidden="true"` on the decorative SVG is exactly right — that's the highest-value piece of the spinner change.
- Component-level test additions for the favorite button's presence and event isolation are good practice.

## Recommended Action

Request Changes. Please:

1. Split this PR into three: (a) spinner ARIA fix, (b) i18n locale normalization, (c) favorites feature.
2. For the spinner PR: remove the duplicate attributes and add translations for `loading_page` to the other 24 locales (or scope the key to a single canonical attribute).
3. For the i18n PR: this is essentially done. Ship it.
4. For the favorites PR: fix the nested `<button>` issue, localize all user-facing strings, convert to Svelte 5 runes (`$derived`), add tests for the store, decide what to do about silent localStorage failures, and drop `PHASE_1_AUDIT.md` + `.vscode/tasks.json`.

Thanks again for the work, and apologies for the long list — splitting the PR will make each piece much faster to land.

aaronbrethorst:
closing per previous feedback. please reopen as smaller PRs.

PR #436, opened 2026-03-25
feat: add favorites/bookmarks system for stops and routes

### LINKED ISSUE #315: Favorites / Saved Stops and Routes
Add a Favorites feature that allows riders to save commonly used stops and routes for faster access.
Users can add an item to favorites using the + button and remove it at any time from the Favorites tab.
This feature improves usability by reducing repeated searches and providing quicker navigation for frequent trips.

## Screenshots:

**Adding to Favorites** :
Image

**Favorites list** :
Image

**Advantages** :
Faster access to frequently used routes and stops
Improved overall user experience
Reduces repeated searching for the same locations

Amartuvshins0404:
## Summary

- Add favorites feature allowing riders to save frequently-used stops and routes for one-tap access
- localStorage-backed store following the existing `recentTripsStore.js` pattern
- Star toggle button on stop pages and modals, dedicated Favorites tab in the search pane
- Closes #315

### New files
- **`src/stores/favoritesStore.js`** — Store with `addFavorite`, `removeFavorite`, `isFavorite`, `clearAll`. Deduplicates by type+entityId. Validates entries on load. Browser-guarded localStorage with try-catch on every operation.
- **`src/components/favorites/FavoriteToggle.svelte`** — Star icon button (filled yellow when favorited, gray outline when not). Keyboard accessible (Enter + Space), ARIA labels via i18n.
- **`src/components/favorites/FavoritesList.svelte`** — Renders saved favorites with name, direction, type icon. Per-item trash button to remove. Empty state message. "Clear all favorites" link.

### Modified files
- **`src/components/search/SearchPane.svelte`** — Added "Favorites" tab between "Stops and Stations" and "Plan a Trip". Renders FavoritesList with click handlers that navigate to the stop/route on the map.
- **`src/components/stops/StopPageHeader.svelte`** — Added FavoriteToggle next to stop name in the heading. Added optional `stopLat`/`stopLon` props for coordinates.
- **`src/components/stops/StopModal.svelte`** — Added FavoriteToggle above StopPane inside the modal.
- **`src/locales/en.json`** — Added `tabs.favorites` and `favorites.*` translation keys.

## Design decisions

- **localStorage only** — mirrors `recentTripsStore.js`. No backend persistence needed. Can be extended later.
- **No max limit** — unlike recent trips (capped at 3), favorites are user-curated and should not be pruned.
- **Dedup by type+entityId** — saving the same stop twice updates the timestamp instead of creating a duplicate. A stop and route with the same entityId are kept as separate favorites.
- **`get()` from svelte/store for one-shot reads** — used in toggle() instead of subscribe/unsubscribe to avoid double-fire inside callbacks.

## Tests

- 14 store tests: add stop/route, LIFO order, no max limit, dedup, remove by ID, isFavorite true/false, clearAll, missing optional fields, malformed data on load, corrupted JSON, localStorage quota exceeded
- 7 FavoriteToggle tests: renders button, aria-label add/remove, calls addFavorite/removeFavorite, Enter key, Space key, missing optional props
- 9 FavoritesList tests: empty state, renders list, shows/hides direction, onStopClick/onRouteClick callbacks, clear all visible/hidden, keyboard navigation
- All 1187 tests pass (1157 existing + 30 new)

## Verification

1. Open the app → "Favorites" tab visible next to "Stops and Stations"
2. Click Favorites tab → shows "No favorites yet" empty state
3. Click any bus stop on map → modal opens with ☆ outline star
4. Click the star → turns ★ yellow, stop appears in Favorites tab instantly
5. Click star again → unfavorites, removed from list
6. Navigate to `/stops/[stopID]` → star appears next to stop name in header
7. Refresh browser → favorites persist from localStorage
8. `npm run lint` passes
9. `npm run test` — 55 files, 1187 tests, all passing

### Favorites tab (empty state)
Favorites tab showing empty state with helper text

### Star toggle on stop page

| Before (unfavorited) | After (favorited) |
|---|---|
| Stop page with unfavorited star outline | Stop page with filled yellow star after favoriting |

### Dark mode support
Favorites tab in dark mode

### DIFF
--- a/src/components/favorites/FavoriteToggle.svelte
+++ b/src/components/favorites/FavoriteToggle.svelte
@@ -1,0 +1,51 @@
+<script>
+	import { FontAwesomeIcon } from '@fortawesome/svelte-fontawesome';
+	import { faStar as faStarSolid } from '@fortawesome/free-solid-svg-icons';
+	import { faStar as faStarOutline } from '@fortawesome/free-regular-svg-icons';
+	import { get } from 'svelte/store';
+	import { favorites } from '$stores/favoritesStore';
+	import { t } from 'svelte-i18n';
+
+	let { type, entityId, name, direction = null, coords = null } = $props();
+
+	let isFav = $state(false);
+
+	$effect(() => {
+		const unsubscribe = favorites.subscribe((list) => {
+			isFav = list.some((f) => f.type === type && f.entityId === entityId);
+		});
+		return unsubscribe;
+	});
+
+	function toggle() {
+		if (isFav) {
+			const entry = get(favorites).find((f) => f.type === type && f.entityId === entityId);
+			if (entry) {
+				favorites.removeFavorite(entry.id);
+			}
+		} else {
+			favorites.addFavorite({ type, entityId, name, direction, coords });
+		}
+	}
+
+	function handleKeydown(e) {
+		if (e.key === 'Enter' || e.key === ' ') {
+			e.preventDefault();
+			toggle();
+		}
+	}
+</script>
+
+<button
+	type="button"
+	onclick={toggle}
+	onkeydown={handleKeydown}
+	aria-label={isFav ? $t('favorites.remove') : $t('favorites.add')}
+	class="inline-flex items-center justify-center rounded-md p-1 transition-colors hover:bg-gray-100 focus:outline-none focus:ring-2 focus:ring-brand-accent dark:hover:bg-gray-700"
+>
+	{#if isFav}
+		<FontAwesomeIcon icon={faStarSolid} class="text-xl text-yellow-400" />
+	{:else}
+		<FontAwesomeIcon icon={faStarOutline} class="text-xl text-gray-400" />
+	{/if}
+</button>
--- a/src/components/favorites/FavoriteToggle.svelte
+++ b/src/components/favorites/FavoriteToggle.svelte
@@ -2,24 +2,16 @@
 	import { FontAwesomeIcon } from '@fortawesome/svelte-fontawesome';
 	import { faStar as faStarSolid } from '@fortawesome/free-solid-svg-icons';
 	import { faStar as faStarOutline } from '@fortawesome/free-regular-svg-icons';
-	import { get } from 'svelte/store';
 	import { favorites } from '$stores/favoritesStore';
 	import { t } from 'svelte-i18n';
 
 	let { type, entityId, name, direction = null, coords = null } = $props();
 
-	let isFav = $state(false);
-
-	$effect(() => {
-		const unsubscribe = favorites.subscribe((list) => {
-			isFav = list.some((f) => f.type === type && f.entityId === entityId);
-		});
-		return unsubscribe;
-	});
+	let isFav = $derived($favorites.some((f) => f.type === type && f.entityId === entityId));
 
 	function toggle() {
 		if (isFav) {
-			const entry = get(favorites).find((f) => f.type === type && f.entityId === entityId);
+			const entry = $favorites.find((f) => f.type === type && f.entityId === entityId);
 			if (entry) {
 				favorites.removeFavorite(entry.id);
 			}
@@ -27,19 +19,11 @@
 			favorites.addFavorite({ type, entityId, name, direction, coords });
 		}
 	}
-
-	function handleKeydown(e) {
-		if (e.key === 'Enter' || e.key === ' ') {
-			e.preventDefault();
-			toggle();
-		}
-	}
 </script>
 
 <button
 	type="button"
 	onclick={toggle}
-	onkeydown={handleKeydown}
 	aria-label={isFav ? $t('favorites.remove') : $t('favorites.add')}
 	class="inline-flex items-center justify-center rounded-md p-1 transition-colors hover:bg-gray-100 focus:outline-none focus:ring-2 focus:ring-brand-accent dark:hover:bg-gray-700"
 >
--- a/src/components/favorites/FavoritesList.svelte
+++ b/src/components/favorites/FavoritesList.svelte
@@ -6,14 +6,7 @@
 
 	let { onStopClick = () => {}, onRouteClick = () => {} } = $props();
 
-	let items = $state([]);
-
-	$effect(() => {
-		const unsubscribe = favorites.subscribe((list) => {
-			items = list;
-		});
-		return unsubscribe;
-	});
+	let items = $derived($favorites);
 
 	function handleItemClick(item) {
 		if (item.type === 'stop') {
--- a/src/components/favorites/FavoritesList.svelte
+++ b/src/components/favorites/FavoritesList.svelte
@@ -1,0 +1,90 @@
+<script>
+	import { favorites } from '$stores/favoritesStore';
+	import { FontAwesomeIcon } from '@fortawesome/svelte-fontawesome';
+	import { faSignsPost, faRoute, faTrash } from '@fortawesome/free-solid-svg-icons';
+	import { t } from 'svelte-i18n';
+
+	let { onStopClick = () => {}, onRouteClick = () => {} } = $props();
+
+	let items = $state([]);
+
+	$effect(() => {
+		const unsubscribe = favorites.subscribe((list) => {
+			items = list;
+		});
+		return unsubscribe;
+	});
+
+	function handleItemClick(item) {
+		if (item.type === 'stop') {
+			onStopClick(item);
+		} else {
+			onRouteClick(item);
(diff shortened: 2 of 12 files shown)


aaronbrethorst (changes requested):
Hey Amartuvshin, this is a really well-put-together feature — the store follows the `recentTripsStore.js` pattern nicely, the test coverage at 30 tests is thorough (especially the localStorage edge cases), and the PR description with screenshots and verification steps is excellent. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Pass `stopLat`/`stopLon` to `StopPageHeader` in both standalone stop pages.** Neither `/stops/[stopID]/+page.svelte` nor `/stops/[stopID]/schedule/+page.svelte` passes the new coordinate props, even though the `stop` object has `lat` and `lon` available. This means any favorite added from these pages saves `coords: null`. When the user clicks that favorite from the Favorites tab, `handleStopClick` calls `mapProvider.flyTo(undefined, undefined, 20)` and `mapProvider.addMarker` with undefined coordinates — which will crash or produce broken map behavior.

   Fix in `src/routes/stops/[stopID]/+page.svelte` line 46:
   ```svelte
   <StopPageHeader stopName={stop.name} stopId={stop.id} stopDirection={stop.direction} stopLat={stop.lat} stopLon={stop.lon} />
   ```
   And similarly in `src/routes/stops/[stopID]/schedule/+page.svelte` line 151.

## Important

2. **Use `$derived` with store auto-subscription instead of manual `$effect` + `subscribe`.** Both `FavoriteToggle.svelte` (lines 13-19) and `FavoritesList.svelte` (lines 12-17) use a `$state` variable updated inside an `$effect` with manual `subscribe`/`unsubscribe`. This project uses Svelte 5 with runes and the store auto-subscription `$storeName` syntax is cleaner and less error-prone. For example, in `FavoriteToggle`:

   ```js
   // Replace the $state + $effect + subscribe pattern with:
   let isFav = $derived($favorites.some((f) => f.type === type && f.entityId === entityId));
   ```

   And in `FavoritesList`:
   ```js
   let items = $derived($favorites);
   ```

   This eliminates the risk of subscription leaks and is more idiomatic.

3. **Remove the redundant `handleKeydown` on `FavoriteToggle`'s `<button>`.** Native HTML `<button>` elements already fire `click` events on Enter and Space keypress — the custom `handleKeydown` handler (lines 33-37) duplicates built-in browser behavior and can be safely removed along with the `onkeydown` binding on line 44.

4. **Add tests for the remove button and "Clear all" button in `FavoritesList`.** Both user actions are implemented but untested:
   - Clicking the trash icon should call `removeFavorite` with the correct ID *and* should NOT trigger `onStopClick`/`onRouteClick` (the `stopPropagation` behavior).
   - Clicking "Clear all favorites" should call `clearAll`.

## Fit and Finish

5. **Use `??` instead of `||` for `direction` and `coords` defaults in `favoritesStore.js` line 61-62.** `item.direction || null` would coerce an empty string `""` to `null`. While unlikely to matter in practice, `??` is the correct operator for "nullish, not falsy" semantics.

aaronbrethorst (changes requested):
Hi Amartuvshin — thanks for the quick turnaround on the previous round. The store refactor to `$derived`, the new tests for the remove and "Clear all" buttons, the `??` change, and the propagated `stopLat`/`stopLon` props all look great. The 33 tests pass cleanly. Almost there — just a few things to clean up before we can merge.

## Critical

1. **`npm run lint` fails.** Prettier reports formatting issues in two files. From a clean checkout of `2e1a5c3`:

   ```
   [warn] src/components/favorites/__tests__/FavoritesList.test.js
   [warn] src/routes/stops/[stopID]/+page.svelte
   ```

   In `FavoritesList.test.js` lines 104–110 and 121–127, the `const item = { ... }` literals need to be wrapped across multiple lines. In `src/routes/stops/[stopID]/+page.svelte` line 46, the `<StopPageHeader ... />` line is now too long and needs to be wrapped across multiple lines.

   Easy fix: `npm run format` then commit. Please confirm `npm run lint` passes locally before pushing.

## Important

2. **No UI for favoriting routes — the PR claims `Closes #315` but only covers stops.** Issue #315 explicitly says "save commonly used stops **and routes**" and that "users can add an item to favorites using the + button." The store handles `type: 'route'`, `FavoritesList` renders route entries with the right icon, and `SearchPane`'s `onRouteClick` is wired up — but nothing in the UI ever creates a route favorite. `FavoriteToggle` is only used in `StopModal.svelte` and `StopPageHeader.svelte`, and `grep -rn "type=\"route\"" src/components/favorites src/components/stops src/components/search` returns nothing. Today the route code paths in the store, list, and search-pane click handler are dead.

   Two acceptable resolutions:

   a) Add a `FavoriteToggle` (with `type="route"`) next to the route name on the route-results panel and/or vehicle popup so users can actually favorite routes. This is what #315 asks for.

   b) Narrow the scope of this PR to stops-only — update the PR title and description, change `Closes #315` to `Refs #315`, and either remove the route handling code or leave a tracking issue for the route side.

   Either is fine, but the PR as written falsely claims to close the issue.

3. **Stop with `coords: null` will crash the map when re-opened from Favorites.** In both `StopModal.svelte:29` and `StopPageHeader.svelte:35`, you pass `coords={stop.lat && stop.lon ? { lat: stop.lat, lng: stop.lon } : null}`. Two failure paths:

   - If `stop.lat` or `stop.lon` is ever missing (during fetch races, stale cached data, etc.), the favorite gets stored with `coords: null`. The user later clicks it from the Favorites tab and `SearchPane`'s `onStopClick` runs `mapProvider.flyTo(undefined, undefined, 20)` and `mapProvider.addMarker({ position: { lat: undefined, lng: undefined } })` — broken behavior.
   - `stop.lat && stop.lon` will also coerce `lat === 0` (equator) to falsy. Unlikely for OBA stops but the right operator is `stop.lat != null && stop.lon != null`.

   Two-part fix: tighten `isValidFavorite` in `favoritesStore.js` to reject stop entries without numeric `coords.lat`/`coords.lng`, and use `!= null` instead of `&&` for the coords guard in both call sites.

## Fit and Finish

4. **`isFavorite()` in the store is dead code outside of tests.** Both `FavoriteToggle.svelte:10` and the previous review feedback moved component logic to `$derived($favorites.some(...))`, which is the right call. Now `favorites.isFavorite()` is only exercised by `favoritesStore.test.js:131-141`. If you want to keep the method for future callers, simplify the implementation using `get()` from `svelte/store`:

   ```js
   import { get } from 'svelte/store';
   // ...
   isFavorite: (type, entityId) =>
       get({ subscribe }).some((f) => f.type === type && f.entityId === entityId),
   ```

   Otherwise consider deleting `isFavorite` and the two tests for it. No need to maintain unused API surface.

5. **Nested interactive elements in `FavoritesList.svelte`.** The outer `<div role="button" tabindex="0">` (line 39) contains a real `<button>` for trash (line 64). Nesting interactive controls inside other interactive controls is an accessibility anti-pattern — screen readers and keyboard navigation can behave unexpectedly even with `stopPropagation`. Consider restructuring so the item label and the trash button are siblings inside a flex container rather than nested. A reasonable shape is: an outer `<div class="flex">` containing a `<button class="flex-1">` for the item and a separate `<button>` for trash.

## Strengths

- Every item from the previous review is addressed cleanly — `$derived` instead of manual subscribe, propagated `stopLat`/`stopLon` on both standalone pages, redundant `handleKeydown` removed, new tests for trash + Clear All, `??` instead of `||`.
- 33 tests in three files, including the hard-to-think-of edge cases: corrupted JSON, malformed array entries, `setItem` quota exceeded.
- Following the `recentTripsStore.js` pattern is the right call. Consistent with how the rest of the project handles localStorage-backed stores.
- PR description with screenshots, verification steps, and design rationale is excellent and made the review much easier.

## Recommended action

1. Run `npm run format` and re-verify `npm run lint` passes.
2. Decide on the route-favoriting question (#2) — either add the UI or rescope the PR.
3. Tighten the coord validation and switch to `!= null` (#3).
4. Either wire `isFavorite()` to a caller or remove it (#4).
5. Restructure the nested buttons in `FavoritesList` (#5).

Thanks again — once the lint failure and the scope/coords issues are sorted, this should be ready to land.

tarunsinghofficial:
Hey @Amartuvshins0404, just checking in on this one. @aaronbrethorst's second review has some clear action items for you. So, are you still planning to address the remaining feedback? 

Happy to help if you're stuck on any of it, and if not, let us know, and we can figure out next steps.

aaronbrethorst:
Please address PR feedback, and then reopen.

PR #438, opened 2026-03-25
feat: add dynamic arrival updates with diffing and fade transitions

### LINKED ISSUE #331: Feat: Add/Remove Arrivals as Time Progresses
If a stop is kept open, only the initial arrivals that were shown when opening the stop details appear.  Older departures should eventually "fall off" after a set amount of time, and new arrivals should be shown within a typical window of time.

Image

Amartuvshins0404:
## Summary

- Arrivals auto-refresh every 30s: departed rows drop out of the list, new ones fade in
- Pure `arrivalFiltering` helpers (`makeKey`, `filterDeparted`) with 15 unit tests
- Keyed `{#each}` + Svelte fade (300ms in after first paint, 200ms out)
- Closes #331

### New files
- **`src/lib/arrivalFiltering.js`** — `makeKey(arrival)` (`tripId` + `serviceDate` + `stopSequence`) and `filterDeparted(arrivals, now)`. ETA uses predicted arrival time when available and > 0, otherwise scheduled arrival time — same rule `ArrivalDeparture` actually renders (no `stopSequence === 0` departure-time branch).
- **`src/lib/__tests__/arrivalFiltering.test.js`** — 15 tests covering predicted vs scheduled, first-stop layover (arrival past / departure future is removed), empty lists, and two visits of the same trip distinguished by `stopSequence`.

### Modified files
- **`src/components/stops/StopPane.svelte`** — Filters each poll through `filterDeparted`. `isFirstLoad` + `await tick()` skips the initial fade-in; reset when the stop changes (alongside the survey hero-question reset). Keyed `{#each}` with `makeKey(arrival)`. Fade wrapper around each accordion item.

## Design decisions

- **0-minute threshold** — drop a row as soon as arrival ETA goes negative
- **Arrival-time ETA only** — matches `ArrivalDeparture.svelte` (`stopSequence || 1` makes its departure branch unreachable)
- **`tripId` + `serviceDate` + `stopSequence` key** — stable across polls, unique for loop/circulator visits
- **Separate in/out durations** — 300ms fade-in (0 on first paint), 200ms fade-out
- **Wrapper div** — Svelte transitions must be on DOM elements, not component tags
- **No `ArrivalDeparture.svelte` changes** — parent owns list membership

## Tests

- 15 unit tests for `arrivalFiltering`
- Full suite: 1791 passing locally

## Verification

1. Open a stop → arrivals list shows only future (or arriving-now) buses
2. Wait 30 seconds → departed buses fade out, new ones fade in
3. Persistent buses do not animate between polls
4. First-stop layover (arrived, not yet departed) disappears instead of showing "arrived N min ago"
5. `npm run lint` passes
6. `npm run test` — 1791 tests passing
7. `npm run build` succeeds

### Stop page arrivals
| Before | After |
|--------|-------|
| Before: static arrivals list | After: dynamic arrivals with diffing |

### Dark mode
| Before | After |
|--------|-------|
| Before: dark mode static | After: dark mode dynamic |


aaronbrethorst (changes requested):
Amartuvshin — impressive work here. The architecture is well thought out: extracting the diffing logic into a pure utility module with thorough tests, keeping the Svelte component integration minimal. The PR description is one of the best I've seen — the design decisions section shows real engineering judgment. A couple of things to address before this can merge.

## Critical Issues (0 found)

None.

## Important Issues (2 found)

### 1. All arrivals fade in on initial page load

**File:** `src/components/stops/StopPane.svelte:252`

On the first load, `previousArrivals` is `null`, so `diffArrivals(null, rawArrivals, Date.now())` tags every arrival as `_isNew = true`. More importantly, since every arrival is new to the DOM on the first render, the `in:fade={{ duration: 300 }}` transition fires on **all** of them. The user opens a stop page and instead of arrivals appearing instantly, they all fade in over 300ms.

This is a subtle UX regression — arrivals should appear immediately on first load, and only animate on subsequent polls when new arrivals appear or departed ones are removed.

**Fix:** Skip the transition on initial render. One approach: track whether this is the first load and conditionally apply the transition:

```svelte
{#each arrivalsAndDepartures.arrivalsAndDepartures as arrival (makeKey(arrival))}
    <div
        in:fade={{ duration: previousArrivals ? 300 : 0 }}
        out:fade={{ duration: 200 }}
    >
```

But there's a catch: `previousArrivals` is set to `rawArrivals` on line 69 *before* the template re-renders, so by the time Svelte processes the transition, `previousArrivals` is no longer `null`. You'd need a separate flag (e.g., `let isFirstLoad = $state(true)`) that gets set to `false` after the first successful render.

### 2. `data-new={arrival._isNew}` is silently dropped

**File:** `src/components/stops/StopPane.svelte:253`

`AccordionItem` only accepts `data`, `header`, and `children` via `$props()`. The `data-new` attribute is not destructured and Svelte 5 doesn't spread unrecognized props to the DOM by default, so this prop is silently ignored.

If `_isNew` is intended to drive future styling or behavior, it needs to be wired up properly (either via the `data` prop or by adding `data-new` support to AccordionItem). If it's not needed, remove it to avoid dead code that suggests functionality that doesn't exist.

## Suggestions (3 found)

### 1. Missing test: `predicted: true` with `predictedArrivalTime === 0`

**File:** `src/lib/arrivalDiffing.js:60`, `src/lib/__tests__/arrivalDiffing.test.js`

The OBA API uses `0` as a sentinel value for "no prediction available," and the code correctly guards against it with `bestPredictedArrival > 0` (falling back to the scheduled time). However, there's no test exercising the `predicted: true, predictedArrivalTime: 0` case. If someone changed `> 0` to `>= 0`, arrivals would be filtered against a zero timestamp and everything would look departed. A single test pinning this behavior would protect a subtle but critical branch.

Similarly, the `stopSequence === 0` branch with `predicted: true` and a nonzero `predictedDepartureTime` is completely untested (lines 50-51).

### 2. `_isNew` flag is unused — consider removing it

The `_isNew` property is computed by `diffArrivals` and attached to every arrival object, but nothing in the template consumes it (the `data-new` attribute is dropped as noted above). The fade transition applies based on DOM entry/exit, not the `_isNew` flag.

If the intent is to use `_isNew` for future conditional styling (e.g., a highlight effect), that's fine — but add a comment indicating it's reserved for future use. Otherwise, simplify `diffArrivals` to just filter departed arrivals without the tagging step.

### 3. `filterDeparted` duplicates ETA logic from `ArrivalDeparture.svelte`

**File:** `src/lib/arrivalDiffing.js:42-68` vs `src/components/ArrivalDeparture.svelte:55+`

The predicted-vs-scheduled time selection logic and the stopSequence === 0 departure-time handling exists in both places. This isn't a blocker, but it's worth noting for future maintainability — if the ETA calculation logic changes in one place, it could drift in the other. The PR description's follow-up suggestion about a shared helper would address this.

## Strengths

- Excellent separation of concerns: pure utility functions with no Svelte/DOM dependencies
- 18 well-structured tests covering edge cases (empty lists, all departed, duplicate tripId with different serviceDate, predicted vs scheduled, stopSequence === 0)
- Non-mutation guarantee (tested!) — `diffArrivals` creates new objects via spread
- Keyed `{#each}` with `makeKey()` ensures stable DOM identity across polls
- `previousArrivals = null` reset on stop change prevents stale diffing across navigation
- Smart use of `tripId + serviceDate` as composite key to handle midnight rollover
- Clean integration: only 10 lines added to StopPane.svelte

## Recommended Action

1. Fix the initial load fade-in — arrivals should appear instantly on first render
2. Either wire up `data-new` properly or remove it along with the `_isNew` tagging
3. Consider the suggestions about shared ETA logic for future cleanup

aaronbrethorst (changes requested):
Amartuvshin — nice follow-through on the prior round. The `isFirstLoad` flag with `tick()` is exactly the right pattern, the `_isNew` tagging is cleanly removed, and the two missing test branches (`predicted: true` with `predictedArrivalTime === 0`, and `stopSequence === 0` predicted/scheduled departures) are now covered. One small thing standing between this and a merge.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

### 1. Lint check is failing

**File:** `src/lib/arrivalDiffing.js:71`

`npm run lint` reports a Prettier violation — there's a trailing blank line at the end of `arrivalDiffing.js`:

```
[warn] src/lib/arrivalDiffing.js
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

The fix is `npm run format`, which removes the extra newline. CI will block the merge until this is clean.

## Suggestions (2 found)

### 1. PR description is stale

The description mentions `diffArrivals` and claims "18 unit tests: makeKey (2), filterDeparted (8), diffArrivals (8)" — but `diffArrivals` was removed in the second commit, and the actual count is 13 tests (2 + 11). Please refresh the summary so future readers (and reviewers) aren't tripped up by the diff between description and code.

### 2. The module name is slightly misleading now

**File:** `src/lib/arrivalDiffing.js`

With `diffArrivals` removed, the file no longer performs "diffing" in the JS sense — it just filters departed arrivals. The actual diffing now happens implicitly via Svelte's keyed `{#each}` block using `makeKey()`. Not a blocker, but `arrivalFiltering.js` would be a more accurate name. Up to you whether to rename.

## Strengths

- Initial-load fade fix uses exactly the right primitive — `await tick()` between the data assignment and the flag flip ensures the first render sees `isFirstLoad = true` (duration 0) and subsequent polls see `false` (duration 300)
- `isFirstLoad = true` reset inside the stop-change `$effect` correctly handles navigation between stops
- `_isNew` removal kept the diffing logic lean — no dead code left behind
- New tests pin two subtle branches: the OBA sentinel `predictedArrivalTime === 0` fallback (would silently regress if someone changed `> 0` to `>= 0`), and the `stopSequence === 0` departure-time path in both predicted and scheduled variants
- All 1170 tests still pass
- `grep` confirms no stale references to `diffArrivals` or `_isNew` anywhere in the codebase

## Recommended Action

1. Run `npm run format` to fix the trailing newline
2. Update the PR description to reflect the actual test count and the removal of `diffArrivals`
3. (Optional) Consider renaming the module to `arrivalFiltering.js`
4. Re-request review and this should be ready to merge

tarunsinghofficial:
Hi @Amartuvshins0404, I picked up the outstanding lint fix so this doesn't get lost and pushed a one-line formatting fix to unblock the merge. @aaronbrethorst this should now be ready to review again.

tarunsinghofficial:
@aaronbrethorst this one is ready to be merged, as we're done resolving the last feedback :)

aaronbrethorst:
Skipping this one in the current review sweep because it has merge conflicts against `develop` — GitHub currently reports the branch as conflicting, so I can't get a clean diff to review or land it as-is.

Could you merge or rebase `develop` into `feat/dynamic-arrivals` and resolve the conflicts? `develop` has moved a fair bit since this branch was opened, so it's worth double-checking that the arrival-diffing logic still lines up with what's there now.

Once it's mergeable again, ping me here and I'll give it a full review.

tarunsinghofficial:
> Skipping this one in the current review sweep because it has merge conflicts against `develop` — GitHub currently reports the branch as conflicting, so I can't get a clean diff to review or land it as-is.
> 
> Could you merge or rebase `develop` into `feat/dynamic-arrivals` and resolve the conflicts? `develop` has moved a fair bit since this branch was opened, so it's worth double-checking that the arrival-diffing logic still lines up with what's there now.
> 
> Once it's mergeable again, ping me here and I'll give it a full review.

Ready to be merged @aaronbrethorst !

aaronbrethorst:
### Code review

Found 1 issue:

1. `filterDeparted` uses departure times for `stopSequence === 0`, but the row it filters never does — so the "remove departed buses" behavior silently doesn't apply to first-stop rows. `ArrivalDeparture.svelte` derives `stopSequence = arrivalDeparture.stopSequence || 1`, which turns `0` into `1`, making its `stopSequence !== 0` check always true and its departure-time branch unreachable; the displayed ETA is always arrival-based. At a terminal with a layover (or a bus that pulls in early and waits), `predictedArrivalTime` is already in the past while `predictedDepartureTime` is not, so the row renders as "arrived N min ago" but `filterDeparted` keeps it in the list. Either drop the `stopSequence === 0` branch so the filter matches what's rendered, or fix the `|| 1` default so both agree. This is the ETA-logic duplication flagged as suggestion #3 in the first review — it has now actually drifted. (bug due to `src/lib/arrivalFiltering.js` vs `src/components/ArrivalDeparture.svelte:29`)

https://github.com/OneBusAway/wayfinder/blob/eb976c05c4825f971930cdef12be126eeac90005/src/lib/arrivalFiltering.js#L44-L55

aaronbrethorst (changes requested):
Thanks for sticking with this one — and for coming back to it after the branch went stale. Everything I asked for in the previous two rounds is genuinely done: `isFirstLoad` with the `await tick()` is exactly the right pattern and it's correctly reset in the stop-change `$effect`, the `_isNew` tagging and the orphaned `data-new` attribute are fully gone, the module rename to `arrivalFiltering` landed cleanly with the import updated, the lint failure is fixed, and the three test branches I asked for are covered. I also checked the `develop` merge resolution and it's clean — nothing from `develop` was dropped.

Two things to fix, and one process item, before this can land.

## 1. The filter and the row disagree about first stops

`filterDeparted` branches on the raw `stopSequence === 0` and uses departure times for that case. But `ArrivalDeparture.svelte:29` does:

```js
let stopSequence = $derived(arrivalDeparture.stopSequence || 1); // Default to 1 if not provided
```

That `|| 1` coerces `0` to `1`, so the `stopSequence !== 0` test on line 59 is always true and the departure-time branch on lines 64-65 is dead code. The rendered ETA is *always* arrival-time-based.

So at a terminal with a layover — or any bus that arrives early and sits — `predictedArrivalTime` is already in the past while `predictedDepartureTime` isn't. The row displays "arrived N min ago" and your filter keeps it anyway, because the filter is looking at departure time. The feature silently doesn't apply to first-stop rows, which are exactly the rows most likely to sit at a stop long enough for this to show.

This is the drift I flagged as suggestion 3 back in March — two copies of the ETA rule going out of sync. The cheapest correct fix is to make `filterDeparted` match what's actually rendered and drop the `stopSequence === 0` branch entirely. If you'd rather keep the departure-time behavior, then `ArrivalDeparture`'s `|| 1` needs to go at the same time so both sides agree — but that's a bigger behavior change and I'd rather not fold it into this PR.

## 2. `makeKey` isn't unique enough for a keyed `{#each}`

```js
return `${arrival.tripId}_${arrival.serviceDate}`;
```

A loop or circulator route that serves the same stop twice within one trip returns two entries with identical `tripId` and `serviceDate`, differing only in `stopSequence`. Svelte 5 requires keys to uniquely identify each item: in dev that's a hard `each_key_duplicate` throw, and in production the keyed reconciler silently mis-handles the duplicate and drops a row. The `{#each}` was unkeyed before this PR, so those lists rendered fine — this is a regression, not a pre-existing issue.

Adding `stopSequence` fixes it and doesn't weaken the cross-poll identity you need, since `stopSequence` is stable for a given trip's visit to a stop:

```js
return `${arrival.tripId}_${arrival.serviceDate}_${arrival.stopSequence}`;
```

A test with two arrivals sharing tripId and serviceDate but differing in stopSequence would lock it in.

## 3. CI hasn't actually run on this branch

The Test and Lint workflows are sitting at `action_required` and have never run against `eb976c05` — only the CLA and CodeRabbit checks reported. So the "all tests pass" claim isn't verified by anything I can see. The branch is also ~40 commits behind `develop` now; `556edd4` added survey-reset lines to the same `$effect` block you're adding `isFirstLoad = true` to, so that merge deserves a careful look. Please merge `develop` in when you push the fixes, and I'll approve the workflow run so we get a real green build.

## Not blocking

The fade transitions run through the Web Animations API, so no CSS `prefers-reduced-motion` block suppresses them. I'm not holding this up because `AccordionItem` already does the same thing with `transition:slide` — it's the existing house pattern. Worth a follow-up that covers both.

Also, the PR description is still stale from two renames ago: it describes `diffArrivals`, `_isNew: true`, `arrivalDiffing.js`, and "18 unit tests" (it's 13 now, in `arrivalFiltering.js`). Please refresh it so the merge commit doesn't preserve a description of code that isn't here.

The shape of this is right and the tests on `filterDeparted` are genuinely good. Fix the two above, get a green CI run, and I'll land it.

tarunsinghofficial:
Addressed the latest review:

1. **filter vs row** — Dropped the `stopSequence === 0` departure-time branch in `filterDeparted`. It now always uses arrival times, matching `ArrivalDeparture` (`stopSequence || 1`). A first-stop layover (arrival past, departure still future) is removed instead of lingering as “arrived N min ago”. Tests cover that case.

2. **`makeKey` uniqueness** — Key is now `tripId_serviceDate_stopSequence`, with a test for two visits of the same trip at the same stop.

3. **develop merge** — Merged current `develop` (includes the survey-reset `$effect` from 556edd4 alongside `isFirstLoad = true`). Local suite is 1791 passing. Workflows still need a maintainer approval to run.

PR description updated to match `arrivalFiltering` / 15 tests (no more `diffArrivals` / `_isNew` / `arrivalDiffing.js`).

aaronbrethorst (approved):
All three blocking items from my last pass are genuinely fixed, and I verified each rather than trusting the summary.

The filter/row alignment is the one I cared most about, and it's right. `ArrivalDeparture` derives `stopSequence` as `arrivalDeparture.stopSequence || 1`, so its `stopSequence !== 0` branch is unreachable and the row always renders an arrival-time ETA. Making `filterDeparted` unconditionally arrival-time based is therefore the matching rule, not an approximation of it — and the two first-stop tests (layover dropped, future arrival kept) pin exactly the case that would drift. The comment explaining why the special case must not come back is the right thing to leave behind.

`makeKey` as `tripId_serviceDate_stopSequence` is sufficient for a keyed `{#each}`, and the "distinguishes two visits of the same trip" test covers the loop/circulator case that motivated it. I also checked the develop merge: `556edd4`'s survey-reset lines are intact in the `untrack` block, with `isFirstLoad = true` added alongside `noMoreArrivals = false` rather than displacing anything. CI is green across build, lint, test, and coverage.

Two things I checked that could have broken and didn't: `noMoreArrivals` and `loadMoreArrivals` both compare *furthest* arrival times, and filtering only removes *nearest* ones, so widening the window still behaves; and `count` is only used as a completed-vs-aborted sentinel, so the filtered count doesn't mislead it.

Merging. One follow-up I'm filing separately rather than holding this on, since the cause lives elsewhere: `SingleSelectAccordion` never clears `activeItem`/`activeData` when an item unmounts, and `handleAccordionSelectionChanged` drives the trip polyline. So if you expand an arrival and it then departs and gets filtered out, the row fades away but its route line stays drawn on the map with nothing expanded. That mechanism predates this PR, but filtering makes it routine instead of rare — it's precisely the bus the rider was watching. The fix belongs in the accordion, not here.

The `prefers-reduced-motion` and shared-ETA-helper items stay open as the non-blocking follow-ups they always were.

Thanks for staying with this one — it's been a long haul and it landed in good shape.

Issue #251, opened 2025-09-22
When viewing the Route Schedule, short-lines aren't handled properly


aaronbrethorst:
From Devin:

> For example, if you look at Route 120 at stop 12434, half of the trips end at Fashion Valley and half end at Kearny Mesa, but this is not signified in the schedule in any way. 

<img width="978" height="45" alt="Image" src="https://github.com/user-attachments/assets/36c1df79-dc41-4f52-9b4e-0703be86df43" />

> In the current OBA schedule view you can highlight shortlines, which isn't the best option either...maybe there's a better solution.

<img width="420" height="700" alt="Image" src="https://github.com/user-attachments/assets/84a60904-02d9-4872-a90f-9e003c569914" />

PR #441, opened 2026-03-25
Feat/shortline fix 251

Hitanshi7556:
Overview
This PR implements short-line detection for route schedules to address issue #251.

Problem
When viewing a route schedule, if some trips end at an intermediate stop (short-line) instead of the final destination, this wasn't indicated in any way. For example, Route 120 at stop 12434 has trips ending at both Fashion Valley and Kearny Mesa, but there was no visual distinction.

Solution
- Detect short-lines by comparing each trip's `stopHeadsign` with the main route headsign
- Highlight short-line times with amber background and † marker
- Show tooltip on hover: "Ends at [destination] — does not continue to [final destination]"
- Display legend only when route has short-lines

 Changes
- `src/routes/stops/[stopID]/schedule/+page.svelte`: Added short-line detection logic in `groupStopTimesByHour()`
- `src/components/schedule-for-stop/RouteScheduleTable.svelte`: Updated UI with amber styling and hover tooltips
- `src/locales/en.json`: Added i18n translations
- `src/routes/stops/__tests__/schedule-shortline.test.js`: Added unit tests (4 tests, all passing)

Testing
- Unit tests pass for short-line detection logic
- Tested with mock data showing mixed destinations
- Tooltips appear instantly on hover

i used mock data to test the feature : 
**AFTER**
image

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,16 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"label": "Build",
+			"type": "shell",
+			"command": "npm",
+			"args": [
+				"run",
+				"build"
+			],
+			"problemMatcher": [],
+			"group": "build"
+		}
+	]
+}
--- a/PHASE_1_AUDIT.md
+++ b/PHASE_1_AUDIT.md
@@ -1,0 +1,183 @@
+# PHASE 1: DISCOVERY & ARCHITECTURE AUDIT
+
+**Date:** 22 March 2026  
+**Scope:** Baseline architecture, delivery-gap analysis, and readiness for implementation
+
+---
+
+## 1) EXECUTIVE SUMMARY
+
+Phase 1 is complete. The current codebase is stronger than expected and already ships key parts of the proposed GSOC scope.
+
+### Current status against target outcomes
+
+- **Favorites / bookmarks (localStorage):** **Not implemented**
+- **Service alerts display:** **Partially implemented / present in stop flow**
+- **Dynamic arrivals auto-refresh:** **Implemented** (30s polling in stop pane)
+- **Extended departures beyond default window:** **Implemented** (date-based stop schedule)
+- **Accessibility audit + fixes:** **Pending formal audit; multiple known issues**
+- **Expanded tests:** **Already strong baseline**
+- **Bug fixes (trip details / past departures):** **Needs targeted verification**
+
+---
+
+## 2) ARCHITECTURE BASELINE
+
+### Stores and state
+
+Observed store files:
+
+- `src/stores/mapStore.js`
+- `src/stores/modalOpen.js`
+- `src/stores/recentTripsStore.js` *(persisted to localStorage)*
+- `src/stores/surveyStore.js`
+- `src/stores/tripOptionsStore.js` *(persisted fields + derived `effectiveDistanceUnit`)*
+- `src/stores/userLocationStore.js`
+
+### localStorage patterns already used
+
+- `src/stores/recentTripsStore.js`
+- `src/stores/tripOptionsStore.js`
+
+These patterns are reusable for `favoritesStore` implementation.
+
+### API and data integration
+
+- OBA SDK wrapper exists in `src/lib/obaSdk.js` with central response guard (`handleOBAResponse`).
+- API routes are organized under `src/routes/api/**`.
+- Alerts data paths already exist:
+  - GTFS-RT alerts endpoint: `src/routes/api/oba/alerts/+server.js`
+  - Stop-level situations consumed in stop pane.
+
+---
+
+## 3) FEATURE GAP FINDINGS (WHAT IS ALREADY DONE VS MISSING)
+
+### A) Favorites / bookmarks
+
+**Status:** Missing (no favorites/bookmark symbols found in `src/**`).
+
+**Phase 2 target files to add:**
+
+- `src/stores/favoritesStore.js`
+- `src/lib/favoritesUtils.js`
+- `src/components/favorites/FavoriteButton.svelte`
+- `src/components/favorites/FavoritesPanel.svelte`
+
+**Likely files to modify:**
+
+- `src/components/StopItem.svelte`
+- `src/components/RouteItem.svelte`
+- `src/routes/+layout.svelte` (or another global nav host)
+
+### B) Service alerts
+
+**Status:** Present in stop experience.
+
+Evidence:
+
+- `src/components/stops/StopPane.svelte` pulls situations from arrivals response and renders `ServiceAlerts`.
+- `src/components/service-alerts/**` already exists.
+
+**Remaining work:** Improve filtering/scoping UX (route/stop relevance), strengthen a11y semantics, add focused tests.
+
+### C) Dynamic arrivals auto-refresh
+
+**Status:** Implemented.
+
+Evidence:
+
+- `src/components/stops/StopPane.svelte` uses `setInterval(..., 30000)` to re-fetch arrivals.
+- Request cancelation via `AbortController` is already in place.
+
+**Remaining work:** refine stale-state messaging and verify no timing regressions in edge cases.
+
+### D) View departures beyond default range
+
+**Status:** Implemented via date picker schedule view.
+
+Evidence:
+
+- `src/routes/stops/[stopID]/schedule/+page.svelte` fetches `/api/oba/schedule-for-stop/[id]?date=...`.
+
+**Remaining work:** evaluate adding explicit time-range controls (if required by mentors).
+
+---
+
+## 4) ACCESSIBILITY BASELINE (PHASE 1 PRIORITIES)
+
+Initial code scan found interactive patterns that need a formal audit pass:
+
+- Click handlers without full keyboard parity in several components.
+- Missing/limited ARIA labels in action-heavy UI.
+- Focus states and modal focus management need review.
+
+### High-priority candidates
+
+- `src/components/StopItem.svelte`
+- `src/components/RouteItem.svelte`
+- `src/components/map/PopupContent.svelte`
+- `src/components/trip-planner/TripOptionsModal.svelte`
+- `src/components/navigation/OverflowMenu.svelte`
+- `src/components/containers/Accordion.svelte`
+
+---
+
+## 5) TEST BASELINE (ACTUAL RUN)
+
+Command executed: `npm run test:coverage`
+
+### Result snapshot
+
+- **Test files:** 52 passed
+- **Tests:** 1157 passed
+- **Coverage:**
+  - Statements: **78.59%**
+  - Branches: **90.30%**
+  - Functions: **70.50%**
+  - Lines: **78.59%**
+
+### Key uncovered zones for this project scope
+
+- map stack (`src/components/map/**`)
+- route/stop server endpoints with 0% files
+- some trip-planner modal/detail components
+- provider and system theme infrastructure
+
+This means testing is not a greenfield task; it is now a targeted expansion task.
+
+---
+
+## 6) RISKS & DEPENDENCIES
+
+- Favorites UX needs clear interaction model (stop vs route vs trip).
+- Accessibility fixes may touch shared primitives, causing broad test updates.
+- Service alerts can vary by regional backend configuration and feed quality.
+
+---
+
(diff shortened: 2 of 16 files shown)


Hitanshi7556:
Resolved all comments, PTAL @aaronbrethorst 
Thank you!

aaronbrethorst (changes requested):
Hitanshi — thanks for tackling #251. The pure-function approach to short-line detection (extracting `groupStopTimesByHour` into `$lib/scheduleUtils.js` so it can be unit-tested) is exactly the right shape, and the amber-with-dagger styling reads well in the screenshot.

That said, I can't merge this in its current state. Three things need to happen before this is ready:

1. The PR has to be split — the short-line fix and the favorites feature are unrelated work that need separate reviews.
2. The short-line tooltip is broken: the i18n placeholder names don't match, so users will see literal `{mainDestination}` text on hover.
3. `npm run lint` exits non-zero — the new Svelte code mixes Svelte 4 `on:event` directives into a Svelte 5 codebase, and 9 files need Prettier.

Details below.

## Critical Issues (5 found)

### 1. The tooltip — the whole point of this PR — is broken
`src/locales/en.json:152` defines the template as:

```
"ends_at_short_line": "Ends at {destination} — does not continue to {mainDestination}"
```

But `src/components/schedule-for-stop/RouteScheduleTable.svelte:21-26` passes:

```js
values: {
    destination: stopTime.stopHeadsign,
    mainHeadsign: schedule.mainHeadsign   // ← key name mismatch
}
```

The placeholder is `{mainDestination}` in the locale file but the value object passes `mainHeadsign`. Every user who hovers a short-line marker sees the literal text `{mainDestination}` in the tooltip. The existing tests don't catch this because they only exercise `groupStopTimesByHour`, never the rendered tooltip.

Fix: either rename the placeholder to `{mainHeadsign}` in `en.json`, or pass `mainDestination: schedule.mainHeadsign` in the values object. Then add a render assertion against the tooltip text so this can't regress.

### 2. Scope creep — three independent bundles in one PR
The PR description says this addresses #251 (short-line detection), but the diff also includes:

- A complete favorites feature: new `src/stores/favoritesStore.js`, new `src/components/favorites/FavoriteButton.svelte`, integration into `RouteItem.svelte`, `StopItem.svelte`, `ModalPane.svelte`, `StopModal.svelte`, plus new tests in those files.
- `PHASE_1_AUDIT.md` (183 lines) — a process/planning artifact that doesn't belong in product code. The contents look like notes for a GSOC scoping exercise.
- `.vscode/tasks.json` — editor-local config. This repo has no `.vscode/` directory today and no gitignore for it.

Please cherry-pick the short-line commits onto a clean branch and open this PR with just those. Open a separate PR against a separate issue for favorites — it touches shared UI primitives (`ModalPane`, `RouteItem`, `StopItem`) and introduces a new persisted public surface that deserves its own review thread. Drop the audit doc and the vscode file entirely (or move the audit content into the issue/PR description if you'd like to share the findings).

### 3. Favorites feature is half-built
Even setting aside the scope-creep argument, the favorites feature isn't ready to ship: there's no surface anywhere in the app for *viewing* favorites. Users can tap the star on a stop or route, but there's no FavoritesPanel, no nav entry, no list. Whatever they favorite vanishes into localStorage with no way to navigate back to it.

Either build the viewing surface so the feature is end-to-end useful, or hold the star buttons until the surface lands. Shipping "add to favorites" without "view favorites" is worse than not shipping at all — users will use it, then come back later and find their saved stops are nowhere.

### 4. Lint fails — `npm run prepush` would block this branch
Two separate issues:

**ESLint (8 errors)** in `src/components/schedule-for-stop/RouteScheduleTable.svelte:74-77, 128-131` — all on the newly added short-line span. The diff uses Svelte 4 directive syntax (`on:mouseenter`, `on:mouseleave`, `on:click`, `on:keydown`) inside a Svelte 5 runes project. CLAUDE.md is explicit: "All components use modern Svelte 5 syntax." Convert to attribute form: `onmouseenter={...}`, `onclick={...}`, etc. (Note that `FavoriteButton.svelte` uses the new form correctly — so this is inconsistent within the same PR.)

**Prettier (9 files)** — every changed `.svelte`, the audit doc, and `.vscode/tasks.json`. Run `npm run format` before pushing.

**Unused import** — `src/routes/stops/[stopID]/schedule/+page.svelte:6` still imports `msToTimeString` even though `groupStopTimesByHour` (the only caller) moved to `scheduleUtils.js`. Remove the import.

Please make sure `npm run prepush` exits 0 before requesting another review.

### 5. Silent localStorage write failures lose user data
`src/stores/favoritesStore.js` wraps every `localStorage.setItem`/`removeItem` in `try { ... } catch (e) { console.warn(...) }`. The in-memory store updates regardless of whether persistence succeeded. The user sees the star fill in, walks away happy, then on next page load their "favorite" is gone.

Hidden errors that produce this scenario: `QuotaExceededError` (mobile Safari, full storage), `SecurityError` (cookies/storage disabled, sandboxed iframes), Safari private browsing. None of these surface to the user — just a `console.warn` no real user will see.

This is the silent-failure anti-pattern. At minimum: write to localStorage *first*, only update the in-memory state on success, and surface a toast or inline message when persistence fails. (Note: this is conditional on keeping the favorites feature in *some* PR — but it has to be addressed before any version of this code ships.)

## Important Issues (6 found)

### 6. `mainHeadsign` undefined would flag every trip as short-line
`src/lib/scheduleUtils.js:13`:

```js
isShortLine: !!(stopTime.stopHeadsign && stopTime.stopHeadsign !== mainHeadsign)
```

If `directionSchedule.tripHeadsign` is ever falsy at the caller (a malformed API response, a regional feed quirk), every trip with a `stopHeadsign` gets flagged as a short-line and the entire schedule fills with daggers. Tighten the guard:

```js
isShortLine: !!(mainHeadsign && stopTime.stopHeadsign && stopTime.stopHeadsign !== mainHeadsign)
```

…and add a test case covering an undefined `mainHeadsign`.

### 7. `FavoriteButton.svelte` doesn't follow project conventions
Two convention violations:

**Manual `onMount` + `subscribe`** — In Svelte 5 with a writable store, this whole block:

```svelte
let isFavorited = $state(false);
onMount(() => {
    const unsubscribe = favorites.subscribe((favs) => {
        isFavorited = favs.some((fav) => fav.id === id && fav.type === type);
    });
    return unsubscribe;
});
```

…collapses to:

```svelte
let isFavorited = $derived(
    $favorites.some((fav) => fav.id === id && fav.type === type)
);
```

…which also correctly tracks `id`/`type` prop changes (the current code only re-checks on store mutation, so changing the props on a mounted component leaks stale state).

**Hand-rolled inline SVG instead of FontAwesome** — Every other icon in the project uses `@fortawesome/svelte-fontawesome` (see `ModalPane.svelte`, `StopPageHeader.svelte`, etc.). Use `faStar` / `faStarRegular`. The inline polygon coordinates are unreadable and inconsistent.

### 8. The favorites store ships a public API mostly nobody calls
`addFavorite`, `removeFavorite`, `isFavorite`, and `clearAll` have zero callers outside the file itself — only `toggleFavorite` is used by `FavoriteButton`. Delete the unused methods until a real consumer appears (YAGNI). This also cuts ~40 lines of repeated try/catch boilerplate.

If you do keep some of these: pull the localStorage persistence into a private `persist(value)` helper so the body of each public method is one line, not six. And `isFavorite` uses `get({ subscribe })` which is wrong — that constructs an anonymous store-shaped object. Reference the actual store: keep `const store = writable(initialFavorites)` and call `get(store)`.

### 9. RouteScheduleTable AM/PM blocks are byte-for-byte duplicated
The short-line span (lines 68-97 and 122-151) is identical for AM and PM. Extract a Svelte 5 snippet:

```svelte
{#snippet timeCell(hour, stopTime, index)}
    {@const key = `${hour}_${index}`}
    {#if stopTime.isShortLine}
        ...
    {:else}
        ...
    {/if}
{/snippet}
```

…then `{@render timeCell(hour, stopTime, index)}` in each `{#each}`. Cuts ~30 lines and prevents the two copies from drifting. While you're in there, `renderScheduleTable(schedule)` is called four times — once per branch — each time re-filtering. Convert to two `$derived` values (`amTimes`, `pmTimes`).

### 10. Tests miss critical surfaces
- **`favoritesStore` has zero tests.** A persisted store that survives across sessions deserves at least the coverage `recentTripsStore.test.js` provides — add/remove/toggle, persistence round-trip, parse-failure fallback. Use `recentTripsStore.test.js` as the template.
- **No tooltip render assertion.** The `{mainDestination}` bug above could have been caught by one test that renders `RouteScheduleTable` with a short-line entry and asserts the rendered tooltip contains the expected text.
- **No keyboard activation test** for the short-line span — the keydown handler is the kind of thing that silently breaks during refactors.

### 11. Fragile test selectors
The `RouteItem` and `StopItem` tests were rewritten to use `screen.getAllByRole('button')[0]` after the wrapper-div restructure. This is brittle — it breaks the moment another button shows up in the row. Prefer `screen.getByRole('button', { name: /route name pattern/i })`, which the favorite-button tests at the bottom of each file already use correctly. The deduplication chain in `ViewAllRoutesModal.test.js:548` (`.filter((btn, idx, arr) => !arr.slice(0, idx).some(...))`) is a particular smell — just use `document.querySelectorAll('.route-item')`, as a few other lines in the same file already do.

## Suggestions (5 found)

- **Keydown handler is verbose and missing `preventDefault`** — `RouteScheduleTable.svelte:77-79, 131-133`. Pressing Space scrolls the page because there's no `e.preventDefault()`. Pull out a named `handleTooltipKey(e, key)` that early-returns on non-Enter/Space and calls `preventDefault`. Avoids the nested ternary and fixes the scroll bug.
- **`e.stopPropagation()` in `FavoriteButton.handleToggle` is now a no-op** — after the row-restructure, `FavoriteButton` is a sibling of the main button, not nested. No ancestor click handler to stop. Either remove it or leave a comment explaining the intent. (The test "favorite click doesn't trigger row click" passes trivially because they're siblings, not because of the stopPropagation call.)
- **`scheduleUtils.groupStopTimesByHour` should use `const`** — `let stopTime of stopTimes` is never reassigned.
- **Empty tooltip box on edge case** — `getTooltip` returns `null` if `stopHeadsign` is missing, but the tooltip `<div>` wrapper renders regardless. Since `isShortLine` already implies `stopHeadsign` is truthy, you can drop the null guard inside `getTooltip` *or* gate the `<div>` on `getTooltip(stopTime) != null`. Either works; pick one.
- **Pre-existing**: `+page.svelte` fetch errors are swallowed with `console.error` only — no user-facing error state when the schedule API fails. Not introduced by this PR (git blames to last year), but worth filing as a follow-up issue.

## Strengths
- Extracting `groupStopTimesByHour` to `$lib/scheduleUtils.js` is the right call — it follows the existing `$lib/*Utils.js` convention and unlocks unit testing.
- The four short-line unit tests (`schedule-shortline.test.js`) cover the happy path, missing headsign, matching headsign, and hour grouping cleanly. Good pure-function tests.
- The amber background + dagger marker + tooltip + footer legend is a thoughtful UX treatment for what's actually a subtle distinction. Reads clearly in the screenshot.
- The click-isolation tests on `RouteItem`/`StopItem` are valuable — that's a real regression class.

## Recommended Action

1. **Split the PR.** Cherry-pick the short-line commits onto a fresh branch off develop. Open this PR with just those changes, plus the fixes below. Open a separate PR (against its own issue) for favorites, *after* the viewing surface exists.
2. **Drop the noise files.** Remove `PHASE_1_AUDIT.md` and `.vscode/tasks.json` from the diff.
3. **Fix the i18n placeholder bug** (#1) so the tooltip actually works.
4. **Make lint pass** — convert `on:event` to `onevent`, remove the unused `msToTimeString` import, run `npm run format`. Confirm `npm run prepush` exits 0.
5. **Tighten `isShortLine`** to also guard against a falsy `mainHeadsign` (#6).
6. **Add a tooltip render test** so the i18n bug can't come back, plus at least one keyboard test for the short-line span.
7. Re-request review when those are done.

For the eventual favorites PR (separate work): build the viewing surface, address the silent-failure pattern in the store, switch FavoriteButton to `$derived` + auto-subscription + FontAwesome, delete the unused store methods, and de-duplicate the AM/PM blocks via snippet.

aaronbrethorst:
Closing this PR per feedback above. Please split into smaller PRs and resubmit.

PR #445, opened 2026-03-26
feat: add dark mode support to stop marker popup and fix related dark mode issues

sannidhyaroy:
## Summary
- Add dark mode support to `PopupContent.svelte` (stop marker popup) [closes #443]
- Add dark mode overrides for Leaflet popup wrapper and tip (`.leaflet-popup-content-wrapper`, `.leaflet-popup-tip`)
- Fix close button hover in `ModalPane` showing white background in dark mode
- Add dark mode text color to "Route Schedules" heading on the schedule page

### PopupContent (stop popup)
| Before | After |
|--------|-------|
|PopupContent pre-edit|PopupContent|

### Leaflet popup wrapper (impacts both stop popup and vehicle popup)
| Before | After |
|--------|-------|
|VehiclePopupContent pre-edit|VehiclePopupContent|

### ModalPane close button hover
| Before | After |
|--------|-------|
|Modal Pane close button hover pre-edit|Modal Pane close button hover|

### Route Schedules heading
| Before | After |
|--------|-------|
|Schedule Page pre-edit|Schedule Page|

## Additional Questions
The stop popup currently shows `Arrival time:` inside a purple badge. Since the badge already visually separates it from the time value, should the trailing colon be removed (i.e. just `Arrival time`)? I left it unchanged for now.

## Test plan
  - [x] Toggle dark mode and click a stop marker on a route map. Popup should have dark background with readable text
  - [x] Verify Leaflet popup wrapper matches the dark background (no white border)
  - [x] Hover over the close (X) button on any sidebar pane in dark mode and it should show subtle dark highlight, not white
  - [x] Navigate to a stop's schedule page in dark mode and the "Route Schedules" heading should be readable
  - [x] Verify all above still look correct in light mode
  - [x] `npm run lint` passes

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -61,6 +61,11 @@
 		@apply dark:bg-brand dark:text-brand-foreground dark:hover:bg-brand;
 	}
 
+	.close-button {
+		@apply rounded px-4 py-2;
+		@apply transition duration-300 ease-in-out hover:bg-neutral-200 dark:hover:bg-neutral-700;
+	}
+
 	.button {
 		@apply rounded-md bg-surface px-3 py-2 text-sm font-semibold text-surface-foreground shadow-sm ring-1 ring-inset ring-gray-300 hover:bg-gray-50;
 	}
--- a/src/assets/styles/leaflet-map.css
+++ b/src/assets/styles/leaflet-map.css
@@ -25,3 +25,9 @@
 .leaflet-control {
 	background-color: #fff;
 }
+
+.dark .leaflet-popup-content-wrapper,
+.dark .leaflet-popup-tip {
+	background: #1f2937;
+	color: #f3f4f6;
+}
--- a/src/components/map/PopupContent.svelte
+++ b/src/components/map/PopupContent.svelte
@@ -10,7 +10,7 @@
 	{#if arrivalTime != null}
 		<p class="my-4 flex items-center text-gray-700 dark:text-gray-400">
 			<span class="mr-2 rounded-md bg-purple-600 px-2 py-1 text-white">Arrival time:</span>
-			<span>{formatSecondsFromMidnight(arrivalTime)}</span>
+			<span class="dark:text-gray-100">{formatSecondsFromMidnight(arrivalTime)}</span>
 		</p>
 	{/if}
 	<div class="mt-auto flex justify-center">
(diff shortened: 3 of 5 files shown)


aaronbrethorst (approved):
Hi Sannidhya — nice piece of fit-and-finish work. The screenshots make it easy to see exactly what changed, the diff is small and scoped, and you addressed both Copilot bot suggestions before this round. The fix for the close-button hover (swapping `dark:hover:bg-neutral-200/50` for a solid `dark:hover:bg-neutral-700`) is exactly the right call — the half-transparent light-on-dark was the bug.

I also like that you mirrored the existing `VehiclePopupContent.svelte` palette (`dark:bg-gray-800` / `dark:text-gray-100` / `dark:text-gray-400`) in the stop popup, and that the raw hex values in `leaflet-map.css` (`#1f2937`, `#f3f4f6`) line up with those Tailwind grays — the popup wrapper and its content now match instead of fighting each other.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (1 found)

- **`src/components/map/PopupContent.svelte:11`** — The `<p>` has `text-gray-700 dark:text-gray-400`, but both child `<span>`s set their own colors (the badge is `text-white`; the time value inherits `text-gray-700` in light mode and `dark:text-gray-100` in dark). The classes on the `<p>` itself never actually paint anything. Not a bug — purely a tidiness call. You could drop them or leave them as a safety net for any future unwrapped text in that paragraph. Either is fine.

## Strengths

- Both Copilot bot suggestions were addressed (text-gray-100 on the time value, empty `<style>` block removed).
- The `.close-button` consolidation into `src/app.css` under `@layer components` is the right home for it — sits alongside `.itinerary-tab` and friends.
- Dark-mode color choices match the existing `VehiclePopupContent.svelte` pattern, so the two popups now look like siblings instead of cousins.
- The raw hex codes in `leaflet-map.css` are consistent with the existing `#fff` on `.leaflet-control`, so you're not mixing styles within the file.
- Dropping the trailing colon on "Arrival time" tightens up the badge nicely.
- `npm run lint` passes cleanly.

## Recommended Action

Ship it. The one suggestion above is optional cleanup and doesn't need to block the merge — tracked separately in #475.

## Verdict

**Merge.**

PR #447, opened 2026-03-26
test(surveys): add unit tests for `skipSurvey()` function

### LINKED ISSUE #444: Add unit tests for skipSurvey() function
## Context

## What needs to happen

## Files

sannidhyaroy:
## Summary
Adds unit tests for `skipSurvey()` covering the three cases outlined in #444:

- Recurring survey skip (`allows_multiple_responses: true, always_visible: true`): verifies only `_skipped_timestamp` is set
- One-time survey skip: verifies only `_skipped` is set
- Stale flag cleanup: verifies `_skipped` is removed when a recurring survey is skipped again

Closes #444

## Test plan
- [x] `npm run test` passes

### DIFF
--- a/src/tests/lib/surveysUtils.test.js
+++ b/src/tests/lib/surveysUtils.test.js
@@ -15,7 +15,7 @@ beforeEach(() => {
 	vi.stubGlobal('localStorage', {
 		getItem: vi.fn((key) => store[key] || null),
 		setItem: vi.fn((key, value) => {
-			store[key] = value;
+			store[key] = String(value);
 		}),
 		removeItem: vi.fn((key) => {
 			delete store[key];
@@ -335,7 +335,7 @@ describe('skipSurvey', () => {
 
 		skipSurvey(survey);
 
-		expect(localStorage.getItem('survey_2_skipped')).toBeTruthy();
+		expect(localStorage.getItem('survey_2_skipped')).toBe('true');
 		expect(localStorage.getItem('survey_2_skipped_timestamp')).toBeNull();
 	});
 
--- a/src/tests/lib/surveysUtils.test.js
+++ b/src/tests/lib/surveysUtils.test.js
@@ -6,7 +6,8 @@ import {
 	getMapSurvey,
 	submitHeroQuestion,
 	updateSurveyResponse,
-	getPrioritySurvey
+	getPrioritySurvey,
+	skipSurvey
 } from '../../lib/Surveys/surveyUtils';
 
 beforeEach(() => {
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Hi Sannidhya,

Thanks for picking up #444 and for the heads-up about the `localStorage` mock — that was the right call, and I appreciate you flagging it in the PR description rather than slipping it through unannounced. The mock change is appropriately scoped (every other `setItem` in this file already passes a string, so it's a no-op for them) and brings the mock closer to real browser behavior, which let you write a stronger assertion in the one-time-skip test.

The three tests map cleanly to the cases I outlined in the issue, the stale-flag test is exactly the regression case I cared about, and everything passes locally. A few things to tighten before merge:

## Critical Issues (0 found)

None.

## Important Issues (3 found)

1. **The recurring-survey timestamp assertion is too loose.** [src/tests/lib/surveysUtils.test.js:329]

   `expect(localStorage.getItem('survey_1_skipped_timestamp')).not.toBeNull()` will pass even if a regression stores the string `"undefined"` or `"NaN"` — exactly the kind of bad value that would silently break the 1-week skip window in `getValidSurveys` (which calls `parseInt` on it at line 46). Please assert the stored value parses to a finite number close to `Date.now()`:

   ```js
   const stored = Number(localStorage.getItem('survey_1_skipped_timestamp'));
   expect(stored).toBeGreaterThan(Date.now() - 1000);
   expect(stored).toBeLessThanOrEqual(Date.now());
   ```

2. **Branch coverage gap on the recurring condition.** [src/lib/Surveys/surveyUtils.js:176]

   The condition is `allows_multiple_responses && always_visible` — a logical AND. The current tests only exercise `(true, true)` and `(false, false)`. A survey with `(true, false)` or `(false, true)` falls into the one-time branch, which is correct behavior but not currently asserted. If someone refactored `&&` to `||` in the future, all three of your tests would still pass. One small parameterized test covering at least one mixed-flag case would close that hole.

3. **`showSurveyModal.set(false)` is part of the function's contract but isn't verified.** [src/lib/Surveys/surveyUtils.js:185]

   This is the only user-visible side effect not currently asserted. A `vi.mock('$stores/surveyStore.js', …)` with a spy on `showSurveyModal.set` would catch a future refactor that drops the call. Low-effort, modest value — worth doing while you're in here.

## Suggestions (1 found)

- A one-line comment near the mock at [src/tests/lib/surveysUtils.test.js:17] explaining that the `String(value)` coercion mirrors browser `localStorage` (which always stores strings) would help the next contributor understand why it's there. Otherwise someone may "fix" it back and quietly weaken the assertions that depend on it.

## Strengths

- The stale-flag cleanup test (the actual regression case from PR #420) is structured well, with the pre-assertion at line 346 making the "before state" and intent obvious.
- Each test uses a distinct survey id and `beforeEach` clears `localStorage`, so isolation is clean.
- Descriptive `it()` names that read like a spec.
- The mock fix is small, defensible, and you raised it in the PR comments rather than silently bundling it. That's the right way to handle scope creep — thank you.

## Recommended Action

1. Tighten the timestamp assertion in test #1.
2. Add one mixed-flag case to verify the `&&` semantics.
3. Add an assertion on `showSurveyModal.set(false)`.
4. Optionally drop in the one-line comment near the mock.

Verdict: **Request changes** — all small, quick fixes, then this should be good to merge.

aaronbrethorst:

PR #451, opened 2026-03-26
test: Add unit tests for skipSurvey() function (#444)

### LINKED ISSUE #444: Add unit tests for skipSurvey() function

Hitanshi7556:
Overview

Adds unit tests for the `skipSurvey()` function to improve coverage for the fix introduced in #420.

Covered

* Recurring surveys: ensures `_skipped_timestamp` is set when `allows_multiple_responses: true`
* One-time surveys: ensures `_skipped` flag is set correctly
* Legacy cleanup: verifies removal of stale `_skipped` flag when `_skipped_timestamp` is used

 Testing

* All tests passing (24 total: 3 new, 21 existing)
* Run using: `npm run test -- --run`
* No lint or build issues

Closes #444

### DIFF
--- a/src/tests/lib/surveysUtils.test.js
+++ b/src/tests/lib/surveysUtils.test.js
@@ -1,4 +1,15 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+vi.mock('$stores/surveyStore', () => ({
+	showSurveyModal: {
+		set: vi.fn()
+	},
+	markSurveyAnswered: vi.fn(),
+	surveyStore: {
+		update: vi.fn()
+	}
+}));
+
 import {
 	getValidSurveys,
 	getValidStopSurvey,
@@ -6,7 +17,8 @@ import {
 	getMapSurvey,
 	submitHeroQuestion,
 	updateSurveyResponse,
-	getPrioritySurvey
+	getPrioritySurvey,
+	skipSurvey
 } from '../../lib/Surveys/surveyUtils';
 
 beforeEach(() => {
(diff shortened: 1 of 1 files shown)


Hitanshi7556:
Resolved all comments PTAL @aaronbrethorst 
Thank you

aaronbrethorst (changes requested):
Hi Hitanshi — thanks for picking up #444 and turning around the `not.toBeNull()` fix on Copilot's first round. The three tests directly mirror the issue and they would catch the exact regression PR #420 fixed. That's the right shape.

I want to push back on a couple of assertions before merging, though — the tests cover the *behaviors* the issue listed, but a future regression could slip past them. Small set of asks below.

---

## Important Issues (3)

### Assert the timestamp is actually a timestamp
[`src/tests/lib/surveysUtils.test.js:344`, `:374`]

`not.toBeNull()` would pass even if a future refactor wrote, say, the string `"true"` into `_skipped_timestamp` — and that would silently break the one-week window at `surveyUtils.js:46` where `parseInt(skipTimestamp, 10)` parses the value. Lock the timer and assert the value:

```js
beforeEach(() => {
    vi.useFakeTimers();
    vi.setSystemTime(new Date('2026-01-15T00:00:00Z'));
});

afterEach(() => {
    vi.useRealTimers();
});

// in the recurring-survey test:
expect(localStorage.getItem('survey_123_skipped_timestamp')).toBe(String(Date.now()));
```

That makes the test prove the stored value is the timestamp you expect, not just "something."

### Side effect `showSurveyModal.set(false)` is untested
[`src/lib/Surveys/surveyUtils.js:185`]

You already mock `showSurveyModal.set` at the top of the file — but nothing asserts it gets called. That's the user-visible side effect: closing the modal. If someone deletes that line, all three of your tests still pass. Add one assertion (in either branch is enough):

```js
expect(showSurveyModal.set).toHaveBeenCalledWith(false);
```

And then `vi.clearAllMocks()` in `beforeEach` so the call count doesn't leak across tests.

### Trim the mock to what `skipSurvey` actually uses
[`src/tests/lib/surveysUtils.test.js:8-10`]

`surveyStore.update` and `markSurveyAnswered` aren't touched by `skipSurvey`. They're harmless but misleading to a reader trying to understand which collaborators matter. Drop them — keep just `showSurveyModal.set`.

---

## Fit and Finish (1)

### Missing trailing newline
[`src/tests/lib/surveysUtils.test.js:376`]

The file ends without `\n`. Add it.

---

## Strengths

- The stale-flag test (`:361-375`) is exactly the regression-prevention shape this needs — if anyone reverts the `removeItem` line at `surveyUtils.js:179`, this test fails. Bug-prevention intent satisfied.
- Test names are descriptive and behavior-focused (`should set only _skipped_timestamp for recurring surveys`), which makes failures self-explanatory.
- Per-test `localStorage` isolation via the closure-over-`store` pattern in `beforeEach` is clean.
- Both branches of the `if/else` in `skipSurvey` are exercised.
- You used different `survey.id` values per test (`'123'`, `'456'`, `'789'`) — small thing, but it means a leaky localStorage stub couldn't accidentally mask a missing reset. Nice.

---

## Verdict

**Request Changes.** None of the asks above are large. Once the timestamp assertion proves it's a timestamp and the modal-close side effect has an assertion, this PR's job — preventing the #420 regression from coming back — is fully done.

tarunsinghofficial:
Hi @Hitanshi7556 

Are you still working on this issue?

aaronbrethorst:

PR #456, opened 2026-03-30
Improve preload handling and error resilience

kartik098ki:
This PR improves the server-side preload logic by making it more reliable and efficient.

Previously, preload operations were executed using Promise.all, which could cause the entire request to fail if any one task failed. This has been updated to Promise.allSettled so that failures in individual tasks do not break the request flow.

A guard has also been added to prevent the preload logic from running on every request, reducing unnecessary work and improving performance.

Basic error logging has been included to make debugging easier in case of failures.

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -1,3 +1,4 @@
+```css
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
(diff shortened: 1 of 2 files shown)


kartik098ki:
Fix public otp system

src/app.css
```diff
@@ -1,3 +1,4 @@
+```css
```
aaronbrethorst, line 1: Using an agentic AI tool instead of a ChatGPT browser window will help you avoid accidentally including these triple backtick delimiters.

PR #460, opened 2026-03-31
Missing ONEBUSAWAY_API_KEY causes SDK initialization failure in local environment

raj-aryan-official:
The application throws an error during startup when the ONEBUSAWAY_API_KEY environment variable is not set.

Error:
ONEBUSAWAY_API_KEY missing — OnebusawaySDK instantiation fails in src/lib/obaSdk.js

This issue commonly occurs in local development environments where environment variables are not configured. As a result, the SDK initialization fails and can break parts of the application that depend on it.

Current behavior:
- App crashes or throws errors when the API key is missing
- Local development becomes difficult without manual setup

Expected behavior:
- The app should handle missing environment variables gracefully
- Developers should still be able to run the app locally without crashes

Proposed solution:
- Add a guard condition before initializing OnebusawaySDK
- Provide a fallback or warning message instead of throwing an error
- Optionally document required environment variables in a .env.example file

Example approach:
- Check if ONEBUSAWAY_API_KEY exists before instantiation
- Log a warning instead of breaking execution

This will improve developer experience and make the project easier to set up for new contributors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -319,7 +319,6 @@
 					"url": "https://opencollective.com/csstools"
 				}
 			],
-			"peer": true,
 			"engines": {
 				"node": ">=18"
 			},
@@ -342,7 +341,6 @@
 					"url": "https://opencollective.com/csstools"
 				}
 			],
-			"peer": true,
 			"engines": {
 				"node": ">=18"
 			}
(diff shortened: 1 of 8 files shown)


aaronbrethorst:
the app doesn't work without an API key.

PR #461, opened 2026-03-31
feat: add smooth fade-in transition for route polylines

sejalkutriyar:
## Summary
- Add opt-in opacity-based fade-in animation for route polylines in the OpenStreetMap provider using `requestAnimationFrame`.
- Enable the animation only for route-shape rendering flows (keeping other polyline callers unchanged).

## Implementation details
- Added `animatePolylineOpacity` helper to smoothly animate polyline opacity on the client.
  - Includes safety guards (SSR, map/layer presence, and layer still attached).
- Updated `createPolyline(points, options)` to accept `{ animate }` (default: `false`).
  - When enabled, the polyline starts at opacity `0` and fades to the target opacity.
- Route flows that display primary route shapes now pass `{ animate: true }`:
  - route modal
  - search route selection

## Testing
- `npm run lint` (passes)
- `npm test -- src/tests/lib/serverCache.test.js --run` (passes, all 37 tests)
- Manual verification done on local dev map view.

### DIFF
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -55,13 +55,12 @@
 			shapeData = await shapeResponse.json();
 			const shapePoints = shapeData?.data?.entry?.points;
 			if (shapePoints && isMounted) {
-				await mapProvider.createPolyline(shapePoints);
+				await mapProvider.createPolyline(shapePoints, { animate: true });
 			}
 		}
 
 		const stopTimes = tripData?.data?.entry?.schedule?.stopTimes ?? [];
 		const stops = tripData?.data?.references?.stops ?? [];
-		// TODO: implement better way to transition to route shape
 		const location = calculateMidpoint(stops);
 
 		if (location) {
--- a/src/components/search/SearchPane.svelte
+++ b/src/components/search/SearchPane.svelte
@@ -133,7 +133,7 @@
 			for (const polylineData of polylinesData) {
 				const shape = polylineData.points;
 				let polyline;
-				polyline = mapProvider.createPolyline(shape);
+				polyline = mapProvider.createPolyline(shape, { animate: true });
 				polylines.push(polyline);
 			}
 
--- a/src/lib/Provider/OpenStreetMapProvider.svelte.js
+++ b/src/lib/Provider/OpenStreetMapProvider.svelte.js
@@ -454,6 +454,26 @@ export default class OpenStreetMapProvider {
 		}).addTo(this.map);
 	}
 
+	animatePolylineOpacity(polyline, targetOpacity = 1, durationMs = 350) {
+		if (!browser || !this.map || !polyline) return;
+		if (typeof requestAnimationFrame !== 'function') return;
+
+		const startTime = performance.now();
+		const tick = (now) => {
+			// Stop animation if polyline was removed/detached.
+			if (!this.map || !this.map.hasLayer(polyline)) return;
+
+			const progress = Math.min((now - startTime) / durationMs, 1);
+			polyline.setStyle({ opacity: targetOpacity * progress });
+
+			if (progress < 1) {
+				requestAnimationFrame(tick);
+			}
+		};
+
+		requestAnimationFrame(tick);
+	}
+
 	createPolyline(points, options = {}) {
 		if (!browser || !this.map) return null;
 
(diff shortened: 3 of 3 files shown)


aaronbrethorst (changes requested):
Hi Sejal — nice touch picking this up off the `// TODO: implement better way to transition to route shape` breadcrumb in `RouteMap.svelte`. The shape of the helper is good: rAF-driven, self-terminating via `hasLayer`, preserves the caller's target opacity rather than stomping it to 1. That's all the right instinct. Before this lands, though, there's one rendering issue that breaks the visual effect, plus a couple of smaller things.

---

## Critical Issues (1)

### Arrow decorators render at full opacity while the polyline fades in
[`src/lib/Provider/OpenStreetMapProvider.svelte.js:507-528`]

`arrowDecorator` is a separate `L.polylineDecorator` instance with its own `pathOptions` (`fillOpacity: 0.85`, etc.) and is added to the map independently of the underlying `L.Polyline`. When `animatePolylineOpacity()` calls `polyline.setStyle({ opacity })`, that only mutates the polyline — the decorator's arrowhead symbols are not affected.

The net effect across the 350ms fade: arrows pop in at full opacity *first*, then the line gradually appears beneath them. That's a more jarring transition than the no-animation status quo.

Options to pick from:

1. Hold off on `arrowDecorator.addTo(this.map)` until the animation completes.
2. Group the polyline and decorator into a single `L.LayerGroup` and animate the pane's CSS opacity instead.
3. On each tick, also re-set the decorator's pattern with the interpolated `fillOpacity`/stroke alpha.

I'd lean toward (1) — simplest, no per-frame work on the decorator, and lining up "arrows appear at the end" reads naturally with the route shape settling in.

---

## Important Issues (2)

### Respect `prefers-reduced-motion`
[`src/lib/Provider/OpenStreetMapProvider.svelte.js:457`]

This is exactly the kind of motion (synthetic opacity transition with no functional purpose) that the `prefers-reduced-motion: reduce` media query is designed to suppress. Add an early-out at the top of the helper:

```js
if (window.matchMedia?.('(prefers-reduced-motion: reduce)').matches) {
    polyline.setStyle({ opacity: targetOpacity });
    return;
}
```

That way users who have asked their OS to reduce motion just get the polyline at its final opacity immediately.

### Google Maps provider silently ignores `animate`
[`src/lib/Provider/GoogleMapProvider.svelte.js:462`]

`CLAUDE.md` calls out that both providers implement the same interface. The Google path here accepts the options object but does nothing with `animate`, so users on the Google provider get no fade — silently. That's not in itself a blocker (the PR description does note the OSM-only scope), but the silent ignore is the worst of the available options.

Two reasonable paths:

1. Implement a Google equivalent — Google's `Polyline.setOptions({ strokeOpacity })` works fine with a rAF loop.
2. Drop a `// TODO(animate): not yet supported on Google Maps provider` near the destructure of `options.animate` in `GoogleMapProvider.createPolyline`, and open a follow-up issue.

Either is fine. I don't want the asymmetry to be invisible.

---

## Fit and Finish (2)

### No tests for the new helper
[`src/lib/Provider/OpenStreetMapProvider.svelte.js:457-475`]

The early-out guards (`!browser`, `!this.map`, `!polyline`, missing `requestAnimationFrame`) and the `hasLayer` termination path are all small, isolated, and testable with a mocked map and rAF. A couple of cheap tests would make this safer to refactor later.

### Linear easing on a 350ms fade
[`src/lib/Provider/OpenStreetMapProvider.svelte.js:466-467`]

Linear at this duration looks fine, just flagging that an `ease-out` (`1 - Math.pow(1 - progress, 2)`) would feel a bit more polished. Not a blocker — feel free to skip.

---

## Strengths

- `hasLayer` self-termination is the right pattern for rAF cleanup — no explicit cancellation bookkeeping needed when polylines are removed mid-animation.
- Preserving `options.opacity` as the *target* and only writing 0 into the initial Leaflet style is clean. Callers that pass `opacity: 0.5` fade to 0.5, not 1. Easy thing to get wrong.
- SSR / missing-rAF guards at the top of the helper are correct, even though `createPolyline` already returns early on `!browser` — defense in depth here is cheap.
- Opt-in via `{ animate: true }` rather than retrofitting all callers is the right call.
- Removing the stale `// TODO: implement better way to transition to route shape` from `RouteMap.svelte` once the work was done — good housekeeping.

---

## Verdict

**Request Changes.** The arrow-decorator timing is a real visual regression — it actively makes the fade-in look worse than no animation. Once that's fixed and `prefers-reduced-motion` is honored, this is a nice quality-of-life improvement.

aaronbrethorst:

PR #462, opened 2026-04-02
fix(i18n): internationalize aria-labels and add component unit tests (closes #406, closes #402)

### LINKED ISSUE #402: Add unit tests for small, self-contained components

### LINKED ISSUE #406: Internationalize hardcoded aria-label strings

traveltamers:
## Summary

This PR addresses two issues in tandem:

### Issue #406 — Internationalize hardcoded `aria-label` strings

Two components had hardcoded English strings in their `aria-label` attributes, which would not be translated for non-English users:

- **`src/components/trip-planner/TripPlanSearchField.svelte`**: Changed `aria-label="Clear"` to `aria-label={$t('search.clear')}`
- - **`src/components/surveys/SurveyQuestion.svelte`**: Changed `aria-label="Required question"` to `aria-label={$t('survey.required_question')}` and added the missing `import { t } from 'svelte-i18n'`
- - **`src/locales/en.json`**: Added the two new translation keys:
-   - `search.clear` → `"Clear"`
-   - `survey.required_question` → `"Required question"`
### Issue #402 — Add unit tests for small, self-contained components

Added a new `src/tests/components/` directory with unit tests for four components:

- **`FullPageLoadingSpinner.test.js`**: Tests `role="status"`, `aria-label`, and `aria-live="polite"` attributes, plus loading text rendering
- - **`CompassArrow.test.js`**: Tests all 8 compass directions render without the `hidden` class; tests that unknown/missing direction renders with `hidden`
- - **`OptionsPill.test.js`**: Tests icon and label rendering, label-only rendering, and default empty state
- - **`TabLink.test.js`**: Tests base CSS class, active class toggling via `current` prop, and `href` attribute
## Testing

All new tests use `@testing-library/svelte` + Vitest, consistent with the existing test setup in `vitest-setup.js`. The `svelte-i18n` mock returns translation keys as-is, so assertions use the key string (e.g. `'search.clear'`) as the expected value.

-SAM

### DIFF
--- a/src/components/surveys/SurveyQuestion.svelte
+++ b/src/components/surveys/SurveyQuestion.svelte
@@ -1,5 +1,6 @@
 <script>
 	import { Radio, Checkbox, Input } from 'flowbite-svelte';
+	import { t } from 'svelte-i18n';
 
 	let {
 		question,
@@ -38,7 +39,7 @@
 		{question.content.label_text}
 	{/if}
 	{#if required && question.content.type !== 'label'}
-		<span class="ml-1 text-red-500" aria-label="Required question">*</span>
+		<span class="ml-1 text-red-500" aria-label={$t('survey.required_question')}>*</span>
 	{/if}
 </label>
 
--- a/src/components/trip-planner/TripPlanSearchField.svelte
+++ b/src/components/trip-planner/TripPlanSearchField.svelte
@@ -12,7 +12,6 @@
 	 * @property {any} onClear
 	 * @property {any} onSelect
 	 */
-
 	/** @type {Props} */
 	let {
 		inputId = 'location-input',
@@ -51,7 +50,7 @@
 			type="button"
 			class="absolute inset-y-0 right-0 flex items-center pr-3"
 			onclick={handleClear}
-			aria-label="Clear"
+			aria-label={$t('search.clear')}
 		>
 			<FontAwesomeIcon icon={faX} class="size-5 text-gray-400" />
 		</button>
@@ -71,7 +70,7 @@
 					class="flex w-full cursor-pointer items-center px-4 py-2 text-left hover:bg-gray-100 dark:text-black"
 					onclick={() => handleSelect(result)}
 				>
-					<FontAwesomeIcon icon={faMapMarkerAlt} class="mr-2 text-gray-400  " />
+					<FontAwesomeIcon icon={faMapMarkerAlt} class="mr-2 text-gray-400 " />
 					{result.displayText}
 				</button>
 			{/each}
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1,194 +1,198 @@
 {
-	"units": {
-		"meters": "m",
-		"kilometers": "km",
-		"feet": "ft",
-		"mile": "mile",
-		"miles": "miles"
-	},
-	"trip-planner": {
-		"arrive": "Arrive",
-		"depart": "Depart",
-		"today": "Today",
-		"tomorrow": "Tomorrow",
-		"plan_your_trip": "Plan Your Trip",
-		"itinerary": "Itinerary",
-		"duration": "Duration",
-		"start_time": "Start Time",
-		"end_time": "End Time",
-		"minutes": "min",
-		"distance": "Distance:",
-		"end": "End",
-		"start": "Start",
-		"meters": "meters",
-		"no_itineraries_found": "No itineraries found",
+		"units": {
+					"meters": "m",
+					"kilometers": "km",
+					"feet": "ft",
+					"mile": "mile",
+					"miles": "miles"
+		},
+		"trip-planner": {
+					"arrive": "Arrive",
+					"depart": "Depart",
+					"today": "Today",
+					"tomorrow": "Tomorrow",
+					"plan_your_trip": "Plan Your Trip",
+					"itinerary": "Itinerary",
+					"duration": "Duration",
+					"start_time": "Start Time",
+					"end_time": "End Time",
+					"minutes": "min",
+					"distance": "Distance:",
+					"end": "End",
+					"start": "Start",
+					"meters": "meters",
+					"no_itineraries_found": "No itineraries found",
+					"loading": "Loading",
+					"planning": "Planning",
+					"from": "From",
+					"to": "To",
+					"search_for_a_place": "Search for a place",
+					"trip_itineraries": "Trip Itineraries",
+					"show_steps": "Show Steps",
+					"hide_steps": "Hide Steps",
+					"options": "Options",
+					"trip_options": "Trip Options",
+					"cancel": "Cancel",
+					"done": "Done",
+					"departure_time": "Departure Time",
+					"leave_now": "Leave Now",
+					"depart_at": "Depart At",
+					"arrive_by": "Arrive By",
+					"wheelchair_accessible": "Wheelchair accessible",
+					"wheelchair_desc": "Find routes suitable for mobility devices and wheelchairs",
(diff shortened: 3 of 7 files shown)


aaronbrethorst (changes requested):
Hi traveltamers — bundling #406 and #402 is fine in principle, and the aria-label substitutions themselves are exactly what #406 asked for. Where we need to back up is verification — the PR description says `npm run lint` passes and all tests pass, but on this branch neither is true. A few of these are mechanical fixes; one is a real test bug.

---

## Critical Issues (3)

### `npm run lint` fails

Running `npm run lint` on `84277f1e` reports:

```
[warn] src/locales/en.json
[warn] src/tests/components/CompassArrow.test.js
[warn] src/tests/components/FullPageLoadingSpinner.test.js
[warn] src/tests/components/OptionsPill.test.js
[warn] src/tests/components/TabLink.test.js
[warn] Code style issues found in 5 files. Run Prettier with --write to fix.
```

The PR body says lint passes. It doesn't. Please run `npm run format` (or `npx prettier --write`) on these files and re-check before re-requesting review. This is also `npm run prepush` per CLAUDE.md.

### `FullPageLoadingSpinner.test.js` — failing test

[`src/tests/components/FullPageLoadingSpinner.test.js:27`]

The test asserts:

```js
const loadingText = screen.getByText('loading');
```

But the template at `src/components/FullPageLoadingSpinner.svelte:26` renders `{$t('loading')}...` — note the trailing literal `...`. With the i18n mock returning the key as-is, the actual DOM text is `loading...`, not `loading`. `getByText` does an exact match by default, so this throws.

Fix:

```js
expect(screen.getByText(/loading/)).toBeTruthy();
// or
expect(screen.getByText('loading...')).toBeTruthy();
```

The PR description says all tests pass — but `npx vitest run src/tests/components/` shows 1 failed / 21 passed.

### `en.json` reformatted, no longer matches sibling locale files

[`src/locales/en.json`]

The intent of this PR is to add two keys (`search.clear`, `survey.required_question`). That should be a ~6-line diff. Instead, this PR has 195 deletions / 195 additions on `en.json` because the whole file's indentation has been changed — top-level keys are now indented one tab in (so the entire file is wrapped a level deeper than the root `{`), and nested values are 5 tabs deep. Every other locale file in `src/locales/` (e.g. `es.json`, `de.json`) still uses the original convention: one tab for top-level keys, two for nested.

This produces three problems:

1. Reviewers can't see the actual content change in the diff noise.
2. Future merge conflicts on `en.json` are now near-guaranteed.
3. `en.json` is no longer structurally consistent with the 24 sibling locale files.

Please revert `en.json` to its original formatting and only diff the two new keys. After that, running `npm run format` once should keep the file consistent.

---

## Important Issues (2)

### Test files have mixed tab/space indentation

Looking at any of the new files in `src/tests/components/`, the indentation alternates between 2 spaces, 6 spaces, 9 spaces, and a tab, sometimes on adjacent lines within the same `it` block. Example from `OptionsPill.test.js`:

```js
describe('OptionsPill', () => {
  	it('renders the label text', () => {           // 2 spaces + tab
      		render(OptionsPill, { props: { ... } }); // 6 spaces + 2 tabs
      		expect(screen.getByText('🚶')).toBeTruthy();
    });

         	it('renders with only a label when no icon is provided', () => {  // 9 spaces + tab
```

`npm run format` will fix all of this in one pass — please run it before pushing.

### Other locale files are not updated

The two new keys (`search.clear`, `survey.required_question`) are only added to `en.json`. The 24 other locales will fall back to English at runtime via `svelte-i18n`, so this isn't a runtime bug, but standard practice in this repo (see the existing structure) is to either (a) add the same keys with English fallbacks across all locales, or (b) call out in the PR description that translations are pending. Right now there's no signal. Even a one-liner in the PR body about it would do.

---

## Fit and Finish (1)

### Test indent style aside, the test substance is reasonable

Looking past the formatting noise:

- `CompassArrow.test.js` correctly enumerates all 8 directions and checks both the known and unknown-direction branches.
- `OptionsPill.test.js` covers icon + label, label-only, and no-props branches.
- `TabLink.test.js` covers active/inactive class and `href` rendering.
- `FullPageLoadingSpinner.test.js` covers ARIA contract (role, aria-label, aria-live) and (once the text bug is fixed) the loading text.

The test ideas are right — they just need to compile and pass before they can land.

---

## Strengths

- The two aria-label changes themselves are surgical and correct; the keys are sensibly nested (`search.clear`, `survey.required_question`).
- Adding `import { t } from 'svelte-i18n'` to `SurveyQuestion.svelte` rather than relying on an existing import — explicit, easy to follow.
- The new `src/tests/components/` directory creates a clean home for component tests (separate from `src/lib/__tests__/` and `src/tests/lib/`), which matches the structure CLAUDE.md hints at.
- Test coverage targets (`CompassArrow`, `FullPageLoadingSpinner`, `OptionsPill`, `TabLink`) are exactly the kind of small self-contained components #402 was after.
- The 8-direction iteration in `CompassArrow.test.js` via `directions.forEach` is a clean way to avoid copy-paste.

---

## Verdict

**Request Changes.** Three things need to happen before this can merge:

1. Run `npm run format` so lint is green.
2. Fix the `getByText('loading')` → `getByText('loading...')` (or regex) so the test suite is green.
3. Revert the en.json whitespace churn so the diff shows only the two added keys.

None of these are large. Once they're done this is a nice contribution — the actual fix for #406 is small and correct, and the tests for #402 hit useful surface area.

aaronbrethorst:

Issue #485, opened 2026-05-26
a11y: Route stop dot markers have no accessible name


tarunsinghofficial:
**Issue:**
1. The small white-dot markers added via addStopRouteMarker are plain SVGs with no role, no label, no keyboard access. A screen reader user has no idea they're there. Also, the main stop markers use keyboard: false in Leaflet, which kills tab navigation to them.
2. addStopRouteMarker sets iconSize: [20, 20]. WCAG 2.2 requires at least 24×24.

tarunsinghofficial:
This Issue will be closed once the new PR #551 from PR #487 merges.

tarunsinghofficial:
Closing this issue as PR is merged!

PR #487, opened 2026-05-27
fix(a11y): make Leaflet and vehicle markers keyboard accessible

tarunsinghofficial:
## What

Fixes accessibility issues on map markers as per WCAG 2.2 audit

## Why

Three real problems caught by axe DevTools and manual keyboard testing:

1. **Route-on-map dot markers** (small white circles shown after picking a 
   route) had no accessible name, no keyboard handler, and were 20×20 px — 
   below the WCAG 2.2 §2.5.8 minimum touch target of 24×24 px.

2. **Vehicle markers** had no accessible name at all. Leaflet wraps every 
   interactive marker in a `<div role="button" tabindex="0">` automatically, 
   so a screen reader would just announce "button" with no context.

### `OpenStreetMapProvider.svelte.js`

- **`addMarker`**: removed the stray `keyboard: false` option. The marker 
  has `interactive: false`, so `keyboard: false` was dead config,, the inner 
  Svelte `<button>` was the focus target either way. Cleaner without it.

- **`addStopRouteMarker`**: 
  - Bumped `iconSize` from `[20, 20]` to `[24, 24]` (WCAG 2.5.8).
  - Marked the inner SVG `aria-hidden="true"` so the Leaflet wrapper is 
    the only thing the AT sees.
  - Set `title`, `alt`, and `aria-label` on the wrapper using `stop.name`.
  - Added explicit `keydown` listener for Enter/Space (Leaflet's built-in 
    keyboard handler wasn't firing consistently in our setup, this makes 
    it deterministic).
  - Fixed a latent bug: was using bare `L.divIcon` / `L.marker` — switched 
    to `this.L` for consistency with the rest of the class.

### `GoogleMapProvider.svelte.js`

- **`addStopRouteMarker`**: added `title: stop.name` to the marker options. 
  Google's `Marker.title` shows a tooltip and is exposed to assistive tech.

## Things I considered but didn't do

- **Wrapping the SVG in a `<button>` inside the divIcon.** First attempt did 
  this - turned out to be wrong. Leaflet's outer wrapper *already* has 
  `role="button"` + `tabindex="0"`, so an inner button created (a) nested 
  interactive controls (ARIA violation) and (b) two tab stops per marker 
  (terrible keyboard UX). Letting Leaflet's wrapper be the single 
  interactive element is the correct pattern.

- **Removing `interactive: false` from `addMarker`.** Those main stop markers 
  use a Svelte component with its own `<button>` for the click target, so the 
  Leaflet wrapper being non-interactive is intentional — keeps the focus on 
  the Svelte button, which is where `sr-only` text + `onClick` already live.

## Tests
All tests passed just like earlier

## Summary by CodeRabbit

* **New Features**
  * Stop markers now expose stop names via tooltip/title text for better accessibility
  * Vehicle markers now present localized destination details, including dynamic updates when trip headsign changes
  * Added keyboard navigation (Enter/Space) to open stop and vehicle popups
* **Localization**
  * Extended vehicle text bundle with new labels and “next stop”/head-sign phrasing
* **Tests**
  * Expanded accessibility and interaction test coverage for marker keyboard behavior, aria-label/title updates, and marker option settings

### DIFF
--- a/src/lib/Provider/GoogleMapProvider.svelte.js
+++ b/src/lib/Provider/GoogleMapProvider.svelte.js
@@ -192,6 +192,7 @@ export default class GoogleMapProvider {
 		const marker = new google.maps.Marker({
 			position: { lat: stop.lat, lng: stop.lon },
 			map: this.map,
+			title: stop.name,
 			icon: {
 				path: google.maps.SymbolPath.CIRCLE,
 				scale: 5,
--- a/src/lib/Provider/OpenStreetMapProvider.svelte.js
+++ b/src/lib/Provider/OpenStreetMapProvider.svelte.js
@@ -14,6 +14,18 @@ import TripPlanPinMarker from '$components/trip-planner/tripPlanPinMarker.svelte
 import { mount, unmount } from 'svelte';
 import { env } from '$env/dynamic/public';
 import { buildVehiclePopupData } from '$lib/vehicleUtils';
+import { get } from 'svelte/store';
+import { t } from 'svelte-i18n';
+
+// activeTrip is always truthy here: the sole caller (vehicleUtils.js) guards on
+// it, and buildVehiclePopupData reads activeTrip.tripHeadsign without optional
+// chaining. Keep this contract consistent rather than implying null is expected.
+function getVehicleLabel(activeTrip) {
+	const translate = get(t);
+	return activeTrip.tripHeadsign
+		? translate('vehicle.to_headsign', { values: { headsign: activeTrip.tripHeadsign } })
+		: translate('vehicle.label');
+}
 
 export default class OpenStreetMapProvider {
 	constructor(handleStopMarkerSelect) {
@@ -118,6 +130,12 @@ export default class OpenStreetMapProvider {
 			iconSize: [40, 40]
 		});
 
+		// keyboard: false is load-bearing. Leaflet's Marker._initIcon stamps
+		// tabindex="0" + role="button" on the wrapper <div> whenever keyboard is
+		// truthy (its default), regardless of interactive. That wrapper holds the
+		// mounted StopMarker, which already has its own real <button> with an
+		// sr-only accessible name — so leaving keyboard on would create nested
+		// interactive controls plus a second, unlabeled, dead tab stop per stop.
 		const marker = this.L.marker([options.position.lat, options.position.lng], {
 			icon: customIcon,
 			interactive: false,
@@ -194,20 +212,47 @@ export default class OpenStreetMapProvider {
 		marker.props.isHighlighted = false;
 	}
 
+	// Leaflet only activates markers on Enter, so we add Space for ARIA button parity.
+	attachKeyboardActivation(el, onActivate) {
+		el.addEventListener('keydown', (e) => {
+			if (e.key === 'Enter' || e.key === ' ') {
+				e.preventDefault();
+				e.stopPropagation();
+				onActivate();
+			}
+		});
+	}
+
 	addStopRouteMarker(stop, stopTime = null) {
-		const customIcon = L.divIcon({
-			html: `<svg width="15" height="15" viewBox="0 0 24 24" fill="#FFFFFF" stroke="#000000" stroke-width="1" stroke-linecap="round" stroke-linejoin="round" class="feather feather-circle"><circle cx="12" cy="12" r="10"/></svg>`,
+		if (!this.map || !this.L) return;
+
+		const customIcon = this.L.divIcon({
+			html: `<svg width="15" height="15" viewBox="0 0 24 24" fill="#FFFFFF" stroke="#000000"
+				stroke-width="1" stroke-linecap="round" stroke-linejoin="round"
+				style="display:block;margin:auto;" aria-hidden="true">
+				<circle cx="12" cy="12" r="10"/>
+			</svg>`,
 			className: 'route-stop-marker',
-			iconSize: [20, 20],
-			iconAnchor: [10, 10]
+			iconSize: [24, 24],
+			iconAnchor: [12, 12]
 		});
 
-		const marker = L.marker([stop.lat, stop.lon], { icon: customIcon }).addTo(this.map);
+		const marker = this.L.marker([stop.lat, stop.lon], {
+			icon: customIcon,
+			title: stop.name
+		}).addTo(this.map);
+
+		const open = () => this.openStopMarker(stop, stopTime);
+
+		const el = marker.getElement();
+		if (el) {
+			el.setAttribute('aria-label', stop.name);
+			this.attachKeyboardActivation(el, open);
+		}
+
+		marker.on('click', open);
 
 		this.stopsMap.set(stop.id, stop);
-
-		marker.on('click', () => this.openStopMarker(stop, stopTime));
-
 		this.stopMarkers.push(marker);
 	}
 
@@ -283,17 +328,27 @@ export default class OpenStreetMapProvider {
 
 		const vehicleIconSvg = createVehicleIconSvg(vehicle?.orientation, color, routeType);
 		const customIcon = this.L.divIcon({
-			html: `<img src="data:image/svg+xml;charset=UTF-8,${encodeURIComponent(vehicleIconSvg)}" alt="" />`,
+			html: `<img alt="" src="data:image/svg+xml;charset=UTF-8,${encodeURIComponent(vehicleIconSvg)}" />`,
 			iconSize: [iconWidth, iconHeight],
 			iconAnchor: [iconWidth / 2, iconHeight / 2],
 			className: '',
 			zIndexOffset: 1000
 		});
 
+		const label = getVehicleLabel(activeTrip);
+
 		const marker = this.L.marker([vehicle.position.lat, vehicle.position.lon], {
 			icon: customIcon,
-			zIndexOffset: 1000
+			zIndexOffset: 1000,
+			title: label
 		}).addTo(this.map);
+
+		const el = marker.getElement();
+		if (el) {
+			el.setAttribute('aria-label', label);
+			// preventDefault inside attachKeyboardActivation is load-bearing here: bindPopup makes Leaflet attach its own _onKeyPress (Enter -> toggle popup) on the keypress event. Suppressing the default keydown stops that synthesized keypress, so our openPopup() doesn't double-fire and flash the popup open-then-closed. stopPropagation does NOT cover this (keydown and keypress are separate events) — do not remove preventDefault.
+			this.attachKeyboardActivation(el, () => marker.openPopup());
+		}
 
 		this.vehicleMarkers.push(marker);
 
@@ -334,7 +389,7 @@ export default class OpenStreetMapProvider {
 
 		const updatedIconSvg = createVehicleIconSvg(vehicleStatus.orientation, color, routeType);
 		const updatedIcon = this.L.divIcon({
-			html: `<img src="data:image/svg+xml;charset=UTF-8,${encodeURIComponent(updatedIconSvg)}" alt="" />`,
+			html: `<img alt="" src="data:image/svg+xml;charset=UTF-8,${encodeURIComponent(updatedIconSvg)}" />`,
 			iconSize: [iconWidth, iconHeight],
 			iconAnchor: [iconWidth / 2, iconHeight / 2],
 			className: '',
@@ -350,6 +405,15 @@ export default class OpenStreetMapProvider {
 			{ routePaths: this._getRoutePaths() }
 		);
 		marker.setIcon(updatedIcon);
+
+		// Leaflet reuses the existing <div> on setIcon and skips re-applying options.title, so refresh both the tooltip and the accessible name when the headsign changes mid-trip; otherwise the hover tooltip goes stale.
+		const updatedLabel = getVehicleLabel(activeTrip);
+		marker.options.title = updatedLabel;
+		const el = marker.getElement();
+		if (el) {
+			el.setAttribute('aria-label', updatedLabel);
+			el.setAttribute('title', updatedLabel);
+		}
 
 		Object.assign(
 			marker.vehicleData,
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -139,7 +139,9 @@
 		"number": "Vehicle #",
 		"data_updated": "Data updated",
 		"no_real_time_data": "We don't have real-time data for this vehicle now.",
-		"next_stop": "Next stop:"
+		"next_stop": "Next stop:",
+		"label": "Vehicle",
+		"to_headsign": "Vehicle to {headsign}"
 	},
 	"arrivals_and_departures_for_stop": {
 		"title": "Arrivals and Departures"
--- a/src/tests/lib/OpenStreetMapProvider.test.js
+++ b/src/tests/lib/OpenStreetMapProvider.test.js
@@ -1,0 +1,263 @@
+import { describe, test, expect, vi, beforeEach } from 'vitest';
+import OpenStreetMapProvider from '$lib/Provider/OpenStreetMapProvider.svelte.js';
+
+// Minimal Svelte component stubs — only imported by the module, never called
+// during these unit tests because we mock openStopMarker directly and never
+// trigger popupopen events.
+vi.mock('$components/map/StopMarker.svelte', () => ({ default: {} }));
+vi.mock('$components/map/PopupContent.svelte', () => ({ default: {} }));
+vi.mock('$components/map/VehiclePopupContent.svelte', () => ({ default: {} }));
+vi.mock('$components/map/ContextMenuPopup.svelte', () => ({ default: {} }));
+vi.mock('$components/trip-planner/tripPlanPinMarker.svelte', () => ({ default: {} }));
+vi.mock('$lib/MapHelpers/generateVehicleIcon', () => ({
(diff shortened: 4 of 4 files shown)


tarunsinghofficial:
Fixes #485

aaronbrethorst (changes requested):
Hi Tarun — The write-up is genuinely excellent: the "Things I considered but didn't do" section, where you explain *why* you avoided nesting an inner `<button>` inside Leaflet's already-`role="button"` wrapper, is exactly the reasoning that's hard to reconstruct later. You clearly tested with real AT and axe, and it shows. The WCAG 2.5.8 sizing fix is done correctly — bumping `iconSize`/`iconAnchor` to 24×24 while keeping the inner SVG at 15px with `margin:auto` grows the focus/touch target without changing the visual dot.

I verified the keyboard behavior against Leaflet 1.9.4's source, and I'm satisfied the core mechanism is sound — including that your manual `keydown` handler is *not* redundant (more on that below). There are a couple of things I'd like tightened before this goes in.

## Critical Issues (0 found)

None. The change is functionally correct: markers get accessible names, the WCAG sizing is right, and keyboard activation works.

## Important Issues (2 found)

- **The `alt` marker option is dead code on `divIcon` markers** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:206`, `:316`]. Leaflet only writes `alt` to the icon element when `icon.tagName === 'IMG'` (leaflet-src.js:7910). Both your stop-route and vehicle markers use `L.divIcon`, which produces a `<div>`, so `alt: stop.name` / `alt: vehicleLabel` never reach the DOM. The accessible name is actually coming from `title` (applied at leaflet-src.js:7903) plus the `aria-label` you set manually — both of which work. Please drop the `alt:` options so the code doesn't imply coverage it doesn't have. (The `title` and `aria-label` are doing the real work here, so removing `alt` changes nothing behaviorally.)

- **No tests for the keyboard-activation path** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:214-220`]. This is a pure-accessibility PR, and the single most valuable behavior in it — Enter/Space on a focused stop-route marker invoking `openStopMarker` — is exactly the kind of thing that regresses silently: no console error, no visual change, it just stops working for keyboard and screen-reader users. The good news is it's cheap to test. The provider takes `this.L` / `this.map` as injectable fields, so a unit test can construct the provider with a fake `L` whose `marker()` returns an object exposing `getElement: () => document.createElement('div')`, dispatch a `KeyboardEvent('keydown', { key: 'Enter' })` (and one for `' '`), and assert the open path fires plus that `preventDefault` is called for Space. There are currently zero provider tests, so this does mean standing up a small new file (e.g. `src/tests/lib/OpenStreetMapProvider.test.js`) — but for a change whose entire purpose is keyboard access, asserting the keyboard actually activates is the bar. The icon-size and Google `title` assertions are optional; the Enter/Space activation test is the one I'd ask for.

## Suggestions (3 found)

- **The vehicle-label expression is duplicated verbatim** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:308-310` and `:377-379`]. `activeTrip?.tripHeadsign ? \`Vehicle to ${activeTrip.tripHeadsign}\` : 'Vehicle'` appears in both `addVehicleMarker` and `updateVehicleMarker`. A single small helper (module-scope `function vehicleLabel(activeTrip)`, matching the repo's preference for the `function` keyword) gives it one source of truth so the two sites can't drift.

- **The `keydown` listener is tied to the icon's DOM node, which Leaflet may replace** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:214`]. You attach via raw `el.addEventListener` on `marker.getElement()`. Leaflet's `setIcon()` destroys and recreates `_icon` (leaflet-src.js:7864 → `_initIcon` → `_removeIcon`), which would silently drop the handler. It's latent today — stop-route markers never get `setIcon()` called on them (only vehicle markers do, in `updateVehicleMarker`) — so this is not an active bug. But it's a non-obvious foot-gun for the next person. Either bind through Leaflet's own event API, or add a one-line comment noting "do not call setIcon on these markers or the keydown handler is lost." (Note: vehicle markers don't need a `keydown` handler because they `bindPopup`, and Leaflet's `_onKeyPress` opens the popup on Enter for popup-bound markers — leaflet-src.js:10489/10595. Nicely consistent.)

- **Provider parity: Google vehicle markers get no accessible name** [`src/lib/Provider/GoogleMapProvider.svelte.js:193`]. You added `title: stop.name` to the Google stop-route marker, but Google vehicle markers still have no label, while the OSM side now does. Issue #485 is Leaflet-specific so this is reasonably out of scope, but `google.maps.Marker` supports an `ariaLabel` option if you want closer parity. Fine to defer to a follow-up.

## Strengths

- Correct architectural call to let Leaflet's wrapper be the single interactive element rather than nesting a button — and you documented why.
- WCAG 2.5.8 target-size fix done right (focus target grows, visual dot unchanged).
- `aria-hidden="true"` on the decorative inner SVG is the right move.
- Switching bare `L.divIcon`/`L.marker` to `this.L.*` matches the rest of the class.
- `e.preventDefault()` on Space correctly prevents page-scroll — the classic keyboard a11y bug, handled.

## Recommended Action

1. Remove the dead `alt:` options on the two `divIcon` markers (2 lines).
2. Add a unit test covering Enter/Space activation of the stop-route marker (and ideally the `Vehicle`/`Vehicle to …` label branches).
3. Optionally extract the `vehicleLabel` helper and add the `setIcon` caveat comment.
4. Re-request review after the fixes.

---

**Verdict: Request Changes.** The mechanism is correct and the write-up is great — this is close. I just want the dead `alt` removed (it misrepresents behavior in a PR all about a11y correctness) and at least one test proving the keyboard actually activates, since that's the whole point of the change.

aaronbrethorst (changes requested):
Hi Tarun — thanks for this. This is exactly the kind of accessibility work that's easy to hand-wave and hard to actually get right, and the PR shows you did the hard part: you went to the WCAG criteria (2.5.8 touch target, 4.1.2 name/role/value), tested with axe and a real keyboard, and — importantly — your PR description documents the dead ends (the nested-button attempt, why `interactive: false` stays). That writeup made reviewing this a pleasure. The keyboard tests dispatching real `KeyboardEvent`s and asserting `preventDefault` on Space are textbook.

I verified the Leaflet behavior claims in this PR against `node_modules/leaflet/dist/leaflet-src.js` directly, ran the new tests (7/7 pass), and confirmed lint + format are clean. A few things need attention before this merges.

## Critical Issues (0 found)

None. The core feature is sound — I confirmed the stop markers are genuinely keyboard-focusable (Leaflet sets `tabIndex=0` + `role="button"` in `_initIcon` whenever `keyboard` is truthy, which it is by default, leaflet-src.js:7914-7916), so the `keydown` handler is reachable in a real browser, not just in the unit test.

## Important Issues (3 found)

- **The `setIcon()` comment is factually wrong about Leaflet's behavior.** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:218-220`] The comment says Leaflet "destroys and recreates the `_icon` DOM node on setIcon, which would silently drop this handler." I checked the source: for a `DivIcon` (which these markers use), `createIcon(oldIcon)` *reuses* the existing `<div>` node (`(oldIcon && oldIcon.tagName === 'DIV') ? oldIcon : ...`, leaflet-src.js:11079). Because the same node comes back, `_removeIcon()` is skipped — `setIcon` only replaces the inner HTML (your SVG), not the outer element. A `keydown` listener on `getElement()` would actually *survive* `setIcon`. The guidance ("keep these markers static") is harmless since the method never calls `setIcon`, but leaving an incorrect technical justification in the tree will mislead the next person who reads it. Please correct it to describe what `setIcon` really does (replaces inner HTML, wiping the SVG + its `aria-hidden`), or restate the rule without the wrong cause.

- **Screen-reader strings are hardcoded in English.** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:17-18`] `"Vehicle to ${tripHeadsign}"` and `"Vehicle"` go straight into `aria-label`/`title`, so every non-English user hears English — which partially undercuts an a11y PR's own goal. There's already a `vehicle` namespace in `src/locales/en.json:137`. I'll be fair: there's existing precedent for un-localized strings in these providers (`GoogleMapProvider.svelte.js:441` hardcodes `'Your Location'`), so you're not introducing a new sin so much as repeating one — but this is the right moment to do it properly. `svelte-i18n`'s `t` store is usable outside components via `get(t)` from `svelte/store`.

- **Vehicle markers respond to Enter but not Space.** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:317-324`] Stop markers got an explicit handler for *both* keys, but vehicle markers rely on Leaflet's native popup activation — and Leaflet's `Popup._onKeyPress` only fires on `keyCode === 13` (Enter). Since these markers advertise `role="button"` (Leaflet sets it), the ARIA button contract expects Space to activate them too. It's an inconsistency with the stop markers you fixed in this same PR. Attaching the same Enter/Space handler to the vehicle element (opening its popup) closes the gap — and would deserve a matching test.

## Suggestions (5 found)

- **`addStopRouteMarker` is missing the `if (!this.map || !this.L) return` guard** that both `addVehicleMarker` (line 299) and `updateVehicleMarker` (line 356) have. [`src/lib/Provider/OpenStreetMapProvider.svelte.js:196`] Both callers are async-driven; without the guard an early call throws a raw `TypeError` instead of failing gracefully. Nice fix switching the bare `L.divIcon`/`L.marker` to `this.L` here, by the way — that was a latent bug.
- **The vehicle aria-label *update* path has no test.** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:375-377`] Vehicles re-poll every ~30s and re-apply the label after `setIcon`. That's the continuously-running path; a test calling `updateVehicleMarker` with a changed headsign and asserting the new `aria-label` would be the single highest-value test to add.
- **`GoogleMapProvider` only gets `title: stop.name`** — no `aria-label` and no keyboard handler on stop markers, and vehicle markers get nothing. [`src/lib/Provider/GoogleMapProvider.svelte.js:194`, `:347`] I understand this PR is OSM-focused, so I don't expect parity here — but please note in the PR description that the keyboard fix is OSM-only so it's not mistaken for cross-provider. Google parity can be a separate follow-up.
- **Test name `"...when activeTrip is undefined-ish"` is misleading.** [`src/tests/lib/OpenStreetMapProvider.test.js:126`] It passes `{}`, not `undefined`/`null`. Worth noting: `getVehicleLabel` is null-safe (optional chaining) but `buildVehiclePopupData` does a bare `activeTrip.tripHeadsign` and would throw on a real `null` — so a true-null test would surface that asymmetry. Either rename to reflect the empty-object case, or add the null case knowingly.
- **`title` and `aria-label` are set to the same string** on each marker. [`:208-217`, `:320-324`] With `role="button"` present, some screen readers announce both (name + description). Not a blocker — just worth a quick SR pass to confirm there's no double announcement.

## Strengths

- The `$state(...)` + `Object.assign(marker.vehicleData, ...)` change is the correct Svelte 5 runes pattern for propagating updates to mounted props, and it fixes a real pre-existing reactivity issue.
- `alt=""` on the decorative vehicle `` with the real label on the marker element is exactly right.
- Removing the dead `keyboard: false` from `addMarker` (the marker is `interactive: false` anyway) is a clean, correct simplification.
- WCAG 2.5.8 sizing bump (20→24px) with the anchor updated to match.
- The negative keyboard test (Tab/ArrowDown do nothing) and the honest comment explaining why the component mocks are never invoked show real test discipline.

## Recommended Action

1. Correct the inaccurate `setIcon()` comment (or remove the false mechanism).
2. Localize the `"Vehicle to X"` / `"Vehicle"` strings via the existing `vehicle` i18n namespace.
3. Add Enter/Space keyboard activation to vehicle markers for parity with stop markers (+ a test).
4. Consider the suggestions (map guard, update-path test, test rename).

**Verdict: Request Changes.** This is good, careful work that's close — the blockers are small and quick. The comment correction is non-negotiable (it's verifiably wrong and will mislead), and an a11y PR shouldn't ship English-only screen-reader text when the i18n namespace is already sitting there. Fix those three Important items and this is a clear merge.

aaronbrethorst (changes requested):
Hi Tarun — thanks for this. The framing is exactly right: Leaflet auto-wraps interactive markers in a `div[role="button"][tabindex="0"]`, so the correct pattern is to let that single wrapper be the accessible name + focus target rather than nesting your own button inside it. Your "Things I considered but didn't do" section shows you already understand the nested-interactive-controls trap, and the new test suite is genuinely good — descriptive names, negative cases, and even a test for re-applying the label after `setIcon`.

That same trap, though, is what trips up one of the changes in this PR. The verdict is **Request Changes** for a single, narrowly-scoped fix.

## Critical Issues (1 found)

### Removing `keyboard: false` from `addMarker` regresses keyboard a11y on every primary stop marker — `src/lib/Provider/OpenStreetMapProvider.svelte.js:126-129`

The PR description justifies dropping `keyboard: false` as removing "dead config" because the marker is `interactive: false`. That reasoning doesn't hold — I verified it against the Leaflet 1.9.4 source.

In `Marker._initIcon` (`node_modules/leaflet/src/layer/marker/Marker.js:235-238`), the keyboard block is **not** gated by `interactive`:

```js
if (options.keyboard) {        // defaults to true; was being suppressed by keyboard:false
    icon.tabIndex = '0';
    icon.setAttribute('role', 'button');
}
```

`keyboard` defaults to `true`, so removing `keyboard: false` makes Leaflet stamp `tabindex="0"` + `role="button"` onto the wrapper `<div>`. That wrapper holds the mounted `StopMarker` component, which already has its own real `<button>` with an `sr-only` accessible name (`src/components/map/StopMarker.svelte:72`). The net effect on every primary stop marker:

- A `div[role="button"][tabindex="0"]` wrapping a real `<button>` — **nested interactive controls**, the exact ARIA violation you deliberately avoided for the route/vehicle markers.
- **Two tab stops per marker**, where the outer one has no accessible name and no handler (`addMarker` attaches no click/keydown to the wrapper, and the marker is `interactive: false`, so Leaflet won't fire one either). A keyboard user lands on a dead, unlabeled control.

This scales with the number of stops on screen, and it's a keyboard regression inside a keyboard-accessibility PR. `keyboard: false` was load-bearing — please restore it on `addMarker`. The route-dot and vehicle markers are unaffected; they correctly rely on Leaflet's wrapper as the single interactive element.

## Suggestions (4 found)

- **Document that `preventDefault` is load-bearing on the vehicle keydown handler** — `src/lib/Provider/OpenStreetMapProvider.svelte.js:202-209`. Vehicle markers call `bindPopup`, so Leaflet binds its own `_onKeyPress` (Enter → `_openPopup`, which *toggles*). Your handler also calls `openPopup()`. These don't double-fire only because `e.preventDefault()` on the **keydown** suppresses the **keypress** Leaflet listens for. Note that `e.stopPropagation()` is *not* what guards this (keydown and keypress are separate events) — so if someone later "cleans up" the `preventDefault` assuming `stopPropagation` covers it, Enter will flash the popup open-then-closed. A one-line comment naming `preventDefault` as the safeguard would prevent that.

- **`updateVehicleMarker` refreshes `aria-label` but not `title`** — `src/lib/Provider/OpenStreetMapProvider.svelte.js:387-389`. On `setIcon`, Leaflet reuses the existing `<div>` and skips re-applying `options.title`, so when a vehicle's headsign changes mid-trip the screen-reader label updates but the hover tooltip goes stale. Set both for consistency.

- **`getVehicleLabel`'s `activeTrip?.tripHeadsign` disagrees with its neighbor** — `src/lib/Provider/OpenStreetMapProvider.svelte.js:19-24`. Three lines away, `buildVehiclePopupData` reads `activeTrip.tripHeadsign` with no optional chaining. The sole caller (`vehicleUtils.js`) already guarantees `activeTrip` is truthy, so the `?.` is dead-defensive and just signals — misleadingly — that null is expected. Match the contract: drop the `?.` here, or add it to `buildVehiclePopupData`. Pick one.

- **Test gaps worth a follow-up** (the suite is good, these are the edges it can't reach):
  - No test exercises the `addMarker` wrapper, so the regression above slipped through green tests.
  - `getElement()` is mocked to always return an element, so the production-critical `if (el)` null path (feature silently no-ops if the marker isn't rendered yet) is unverified. A real-Leaflet jsdom or Playwright check that a marker added to a map exposes a focusable, labeled element would be the highest-value addition.
  - No parity test for the `GoogleMapProvider` `title: stop.name` change.
  - The i18n mock hardcodes its own copies of `vehicle.label` / `vehicle.to_headsign`; importing the real `en.json` would catch a future key rename instead of staying green.

## Strengths

- Correct core architecture: one interactive wrapper per route/vehicle marker, decorative SVG marked `aria-hidden`, WCAG 2.5.8 touch target bumped to 24×24.
- Switched the bare `L.divIcon`/`L.marker` to `this.L`, fixing a latent inconsistency.
- The test suite dispatches real `KeyboardEvent`s against a real DOM node, so the Enter/Space/other-key branching and `preventDefault` are genuinely protected — not mock theater for the handler logic.
- A clear, honest PR description that documents the road not taken.

## Recommended Action

1. Restore `keyboard: false` on `addMarker` (the critical fix).
2. Add the clarifying comment about `preventDefault` on the vehicle keydown handler.
3. Consider the `title`-sync and `getVehicleLabel` consistency tweaks.
4. Re-run the suite and re-request review.

**Verdict: Request Changes.**

aaronbrethorst (approved):
terrific work! this is ready to merge

PR #488, opened 2026-05-27
fix(a11y): avoid rendering empty h2 in trip details pane

tarunsinghofficial:
## What
Fixes accessibility issue- an empty `<h2>` 
was being rendered in `TripDetailsPane.svelte` whenever route info wasn't 
available yet.

## Why
The `<h2>` tag was unconditional but its content sat inside `{#if routeInfo}`, 
so before the API returned (or when it didn't include route references) the 
DOM contained `<h2 class="h2"><!----></h2>`. Axe flagged this as a 
"Headings should not be empty" violation, and screen reader users lose 
their place when navigating by headings.

## Changes
- Moved the `{#if routeInfo}` block to wrap the entire `<h2>`, so either 
  the heading renders with content or it doesn't render at all.
- Removed a trailing ` -` from the heading text. It was a dangling dash with 
  nothing after it, leftover from older code.

## Verify
1. Open a stop, click an arrival to expand the trip details pane.
2. While loading — no empty heading in the DOM.
3. After loading, the heading reads "Route X" (no trailing dash).
4. Re-run axe DevTools,, the violation on `.trip-details-pane > h2` is gone.

### DIFF
--- a/src/components/oba/TripDetailsPane.svelte
+++ b/src/components/oba/TripDetailsPane.svelte
@@ -106,7 +106,8 @@
 	{:else if tripDetails}
 		{#if routeInfo}
 			<h2 class="h2">
-				{$_('trip_details.route')} {routeInfo.shortName}
+				{$_('trip_details.route')}
+				{routeInfo.shortName}
 			</h2>
 		{/if}
 		{#if tripDetails.schedule?.stopTimes.length > 0}
--- a/src/components/oba/TripDetailsPane.svelte
+++ b/src/components/oba/TripDetailsPane.svelte
@@ -104,11 +104,11 @@
 	{#if error}
 		<p>{error}</p>
 	{:else if tripDetails}
-		<h2 class="h2">
-			{#if routeInfo}
-				{$_('trip_details.route')} {routeInfo.shortName} -
-			{/if}
-		</h2>
+		{#if routeInfo}
+			<h2 class="h2">
+				{$_('trip_details.route')} {routeInfo.shortName}
+			</h2>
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hi Tarun — this is a clean, well-scoped accessibility fix. The PR description is excellent: it names the exact axe violation, explains *why* the empty `<h2><!----></h2>` was appearing (unconditional tag wrapping conditional content), and gives clear verification steps. That context made reviewing this a pleasure.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (0 found)

None blocking. The change is minimal and idiomatic Svelte 5. Worth noting for the record:

- `routeInfo` is either `null` (initial state) or the result of `data.references.routes.find(...)`, which can be `undefined` when no matching route is present. Wrapping the entire `<h2>` in `{#if routeInfo}` correctly guards both cases — no empty heading in either state. ✅
- The two text expressions (`{$_('trip_details.route')}` and `{routeInfo.shortName}`) on separate lines render as `"Route 44"` with Svelte's standard whitespace collapsing, so dropping the trailing ` -` produces clean output. ✅

## Strengths

- Solves the actual a11y problem at the right level — moving the conditional to wrap the element rather than patching around the symptom.
- Removes leftover dead syntax (the dangling dash) while you're in there, without scope creep.
- No behavioral regression: when route info is unavailable, omitting the heading entirely is the correct, screen-reader-friendly outcome.
- Passes Prettier formatting.

## Recommended Action

Merge as-is.

PR #489, opened 2026-05-27
test: add Go-back guard and LanguageSwitcher safeTranslate fallback tests

### LINKED ISSUE #486: Add tests for error page Go-back guard and LanguageSwitcher safeTranslate fallbacks
Two small test gaps left over from PR #478 that don't justify holding up the merge but are worth backfilling.

## 1. `history.length > 1` guard on the error page is untested

`src/routes/+error.svelte:115-128` — the Go back button falls back to `/` when there's no history (e.g., deep link, refresh, new tab). Neither branch has test coverage. A couple of small tests with `vi.spyOn(window.history, 'back')` and a `window.location.href` spy would lock the behavior in.

## 2. `LanguageSwitcher`'s `safeTranslate` has no fallback tests

`src/components/navigation/LanguageSwitcher/LanguageSwitcher.svelte:12-20` — the same three-mode pattern used in `src/routes/__tests__/error.test.js` (non-function `$_`, throwing `$_`, returns-key) would drop in symmetrically. Real risk is low (LanguageSwitcher renders after i18n is initialized in practice), but the guard exists for a reason and pinning it would prevent silent regressions.

Both are belt-and-braces, neither blocks anything.

tarunsinghofficial:
Fixes #486 

## Summary by CodeRabbit

* **Tests**
  * Expanded language switcher test coverage for error scenarios and fallback behavior
  * Added error page navigation tests for proper back button functionality

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/wayfinder/pull/489?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/navigation/LanguageSwitcher/__tests__/LanguageSwitcher.test.js
+++ b/src/components/navigation/LanguageSwitcher/__tests__/LanguageSwitcher.test.js
@@ -2,6 +2,7 @@ import { render, screen, waitFor } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
 import { expect, test, describe, vi, beforeEach, afterEach } from 'vitest';
 import LanguageSwitcher from '../LanguageSwitcher.svelte';
+import { __setTranslatorMode } from 'svelte-i18n';
 
 // Mock languages array - must be defined inline in mock factory
 vi.mock('$lib/i18n', () => ({
@@ -20,7 +21,30 @@ let currentLocale = 'en';
 const localeSubscribers = [];
 
 vi.mock('svelte-i18n', () => {
-	// Create the mock locale object inside the factory
+	let translatorMode = 'normal';
+	const tSubscribers = [];
+
+	const normalT = (key, options) => {
+		if (key === 'language_switcher.select_language') {
+			return `Select language: ${options?.values?.language || ''}`;
+		}
+		if (key === 'language_switcher.available_languages') {
+			return 'Available languages';
+		}
+		return key;
+	};
+
+	const getTranslator = () => {
+		if (translatorMode === 'non-function') return null;
+		if (translatorMode === 'throws') {
+			return () => {
+				throw new Error('Translator error');
+			};
+		}
+		if (translatorMode === 'returns-key') return (key) => key;
+		return normalT;
+	};
+
 	const mockLocale = {
 		subscribe: vi.fn((fn) => {
 			fn(currentLocale);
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Nice, focused PR, Tarun. This is exactly the kind of coverage backfill I like to see: you took two real branches that had shipped untested (`safeTranslate`'s fallback paths and the error page's history-aware Go-back button) and pinned them down with tests that actually distinguish the right behavior from the wrong behavior. I ran both files locally — **58 tests pass** — and `npm run lint` is clean.

**Verdict: Merge.**

The two cleanup nits below are genuinely minor (they only bite on an *already-failing* assertion) and are tracked in #498 so they don't get lost — no need to hold this PR for them.

---

## Critical Issues (0 found)

None. Both pieces of code under test have all branches covered, and the assertions match the components' actual behavior.

## Important Issues (0 found)

None blocking.

## Suggestions (2 found) — tracked in #498

- **Mock cleanup can leak on assertion failure** [`src/routes/__tests__/error.test.js`]. The Go-back tests restore their spies and the `Object.defineProperty(window, 'location', ...)` stub *inline* at the end of each test body, and the file has no top-level `afterEach`. If an assertion throws first (e.g. `expect(hrefValue).toBe('/')`), the restore never runs and `window.location` / `history.back` / `history.length` stay mocked for the rest of the run. The sibling `LanguageSwitcher.test.js` already guards this with `afterEach(() => vi.restoreAllMocks())` — since this file is modeled on that one, matching it (plus a `try/finally` around the location redefine) would close the gap.
- **Parity on the error-page "throws" test** [`src/routes/__tests__/error.test.js:205`]. Your LanguageSwitcher "throws" test asserts the `[i18n fallback]` `console.warn` fires with the right message — really good. The error-page equivalent only asserts the rendered fallback, so deleting the `console.warn` line in `+error.svelte` wouldn't fail any error-page test. One mirrored assertion would cover that side-effect.

## Strengths

- **The negative assertions are what make these tests worth having.** The Go-back tests assert both directions of the conditional: `history.back()` called / `location.href` untouched when `length > 1`, and `back()` *not* called / `href === '/'` when `length <= 1`. That's precisely what catches a flipped condition — easy to skip, and you didn't.
- **The `returns-key` distinction is handled correctly for the error page.** `+error.svelte` treats a key-returning translator as a miss (`result !== key ? result : fallback`), and your test asserts the hardcoded fallback *appears* while the raw key *does not* — the exact assertion that exercises that guard.
- **The `__setTranslatorMode` subscriber-broadcast refactor is the right model, not a shortcut.** Re-pushing the new translator to live `tSubscribers` on mode change is necessary because the component subscribes to `$t` reactively; a naive value-only stub wouldn't update the rendered output. It's also consistent with the pattern already established in `error.test.js`.
- The LanguageSwitcher "throws" test is the strongest of the set — it exercises both the button and listbox call sites and asserts the warning with its error argument.
- New `safeTranslate` tests assert on `aria-label`/`role` rather than brittle class selectors. Good instinct.

## Recommended Action

1. Merge as-is.
2. Pick up #498 when convenient (one `afterEach` + a `try/finally` + one mirrored warn assertion).

PR #495, opened 2026-05-30
refactor: extract buildVehiclePopupData helper for map providers

### LINKED ISSUE #471: Refactor: extract shared buildVehiclePopupData helper for both map providers
## Context

PR #419 fixes the vehicle popup staleness bug by making `marker.vehicleData` a `$state` proxy and mutating it via `Object.assign` in `updateVehicleMarker`. The fix works correctly, but the vehicle data shape is now duplicated four times across the two providers:

- `GoogleMapProvider.svelte.js` — `addVehicleMarker` and `updateVehicleMarker`
- `OpenStreetMapProvider.svelte.js` — `addVehicleMarker` and `updateVehicleMarker`

If a new prop is added to `VehiclePopupContent` and only one path is updated, that field will show stale data silently — the exact class of bug PR #419 just fixed.

## Proposal

Extract a small shared helper:

```js
export function buildVehiclePopupData(vehicle, activeTrip, stopsMap) {
    return {
        nextDestination: activeTrip.tripHeadsign,
        vehicleId: vehicle.vehicleId,
        lastUpdateTime: vehicle.lastUpdateTime,
        nextStopName: stopsMap.get(vehicle.nextStop)?.name,
        predicted: vehicle.predicted
    };
}
```

Use it in all four call sites. Creation path:

```js
const vehicleData = $state(buildVehiclePopupData(vehicle, activeTrip, this.stopsMap));
marker.vehicleData = vehicleData;
```

Update path:

```js
Object.assign(marker.vehicleData, buildVehiclePopupData(vehicleStatus, activeTrip, this.stopsMap));
```

## Minor cleanup to fold in

While you're there: `let vehicleData = $state({...})` in both providers should be `const` since the binding is never reassigned (mutations happen via `Object.assign` on the proxy, not by replacing the variable). Flipping to `const` better signals intent.

## Why a separate issue

PR #419 is correct and ready to merge — the author already flagged this as a follow-up in the PR description. Tracking it here so it doesn't get lost.

VishalRaut2106:
### Description
Closes #471

This PR refactors both `GoogleMapProvider` and `OpenStreetMapProvider` to extract and use a shared helper `buildVehiclePopupData` inside the vehicle utility layer. This deduplicates the vehicle popup details schema across both providers, ensuring that any future property additions automatically synchronize across both map engines.

Additionally, local state proxy variables inside the providers were updated from `let` to `const` to signal Svelte 5 state intent.

### Technical Changes
- Extracted and exported `buildVehiclePopupData` inside `src/lib/vehicleUtils.js`.
- Imported and integrated the helper inside `src/lib/Provider/GoogleMapProvider.svelte.js` (`addVehicleMarker` and `updateVehicleMarker`).
- Imported and integrated the helper inside `src/lib/Provider/OpenStreetMapProvider.svelte.js` (`addVehicleMarker` and `updateVehicleMarker`).
- Optimized `$state` proxy variable declarations from `let` to `const`.

### Verification
- **Formatting**: Formatted all three files with Prettier.
- **Linter**: ESLint checked the three files and passed with 0 errors and 0 warnings.
- **Tests**: Ran the Svelte component and utility tests.

## Summary by CodeRabbit

* **Refactor**
  * Vehicle popup data handling refactored to improve code organization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/wayfinder/pull/495?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/Provider/GoogleMapProvider.svelte.js
+++ b/src/lib/Provider/GoogleMapProvider.svelte.js
@@ -9,6 +9,7 @@ import VehiclePopupContent from '$components/map/VehiclePopupContent.svelte';
 import { createVehicleIconSvg, iconHeight, iconWidth } from '$lib/MapHelpers/generateVehicleIcon';
 import TripPlanPinMarker from '$components/trip-planner/tripPlanPinMarker.svelte';
 import { mount, unmount } from 'svelte';
+import { buildVehiclePopupData } from '$lib/vehicleUtils';
 
 export default class GoogleMapProvider {
 	constructor(apiKey, handleStopMarkerSelect) {
@@ -351,13 +352,7 @@ export default class GoogleMapProvider {
 
 		this.vehicleMarkers.push(marker);
 
-		let vehicleData = $state({
-			nextDestination: activeTrip.tripHeadsign,
-			vehicleId: vehicle.vehicleId,
-			lastUpdateTime: vehicle.lastUpdateTime,
-			nextStopName: this.stopsMap.get(vehicle.nextStop)?.name,
-			predicted: vehicle.predicted
-		});
+		const vehicleData = $state(buildVehiclePopupData(vehicle, activeTrip, this.stopsMap));
 
 		marker.vehicleData = vehicleData;
 
@@ -395,13 +390,10 @@ export default class GoogleMapProvider {
 			anchor: new google.maps.Point(iconWidth / 2, iconHeight / 2)
 		});
 
-		Object.assign(marker.vehicleData, {
-			nextDestination: activeTrip.tripHeadsign,
-			vehicleId: vehicleStatus.vehicleId,
-			lastUpdateTime: vehicleStatus.lastUpdateTime,
-			nextStopName: this.stopsMap.get(vehicleStatus.nextStop)?.name,
-			predicted: vehicleStatus.predicted
-		});
+		Object.assign(
+			marker.vehicleData,
+			buildVehiclePopupData(vehicleStatus, activeTrip, this.stopsMap)
+		);
 	}
 
 	removeVehicleMarker(marker) {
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Nice cleanup, Vishal. Closing #471 by pulling the duplicated vehicle-popup schema into a single helper is exactly the right instinct — the whole value here is that a future field addition now lands in both map engines automatically instead of drifting out of sync. The diff is small, focused, and does precisely what the description says.

I checked out commit `e2b9b3b`, ran the full suite (1275 tests passing), and confirmed `npm run lint` is clean (formatting + ESLint, 0 errors).

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (1 found)

- **`buildVehiclePopupData` has no unit test** [`src/lib/vehicleUtils.js:96`]. The function is pure and trivial, and the inline code it replaces was equally untested — so this is not blocking. But extracting and *exporting* it changes its status: it's now a shared contract two providers depend on, and the deduplication this PR is built around is only guaranteed if that contract holds. Two details make it worth pinning: the `stopsMap.get(vehicle.nextStop)?.name` lookup (returns `undefined` when the stop isn't loaded yet — a routine case the popup falls back on via `$t('NA')`), and the field-source asymmetry where `nextDestination` comes from `activeTrip.tripHeadsign` while everything else comes from `vehicle`. A ~15-line, mock-free test at `src/lib/__tests__/vehicleUtils.test.js` would cover it. I've filed this as a follow-up so it doesn't block the merge: **OneBusAway/wayfinder#501**.

## Strengths

- **Faithful extraction.** The helper returns exactly the same five fields, computed by the same expressions, that both providers built inline. All four call sites (`GoogleMapProvider.svelte.js:355` & `:395`, `OpenStreetMapProvider.svelte.js:296` & `:345`) pass the matching arguments — output is identical.
- **Correct Svelte 5 usage.** `const vehicleData = $state(buildVehiclePopupData(...))` is equivalent to the prior inline `$state({...})` literal — `$state` deep-proxies the returned object regardless of how it was constructed, and reactivity into `VehiclePopupContent` is preserved through both the initial assignment and the later `Object.assign`. The `let`→`const` change is right since the binding is never reassigned, only its properties mutated.
- **Edge cases stay safe.** The sole caller gates on `if (activeTrip && ...)`, so `activeTrip.tripHeadsign` is never a null deref, and `this.stopsMap` is always a `Map`. No guards were added, but none were removed — behavior matches the prior code exactly.
- Used the `$lib` path alias and the existing utility module, consistent with project conventions.

## Recommended Action

This is ready to go as-is. The single suggestion is tracked in #501 for a future PR.

PR #503, opened 2026-06-02
test: add unit tests for buildVehiclePopupData helper

### LINKED ISSUE #501: Add unit tests for buildVehiclePopupData helper
Follow-up from #495, which extracted `buildVehiclePopupData(vehicle, activeTrip, stopsMap)` into `src/lib/vehicleUtils.js` as a shared, exported contract used by both `GoogleMapProvider` and `OpenStreetMapProvider`.

The function is currently untested. Since it's now a public contract two providers depend on, a small unit test would lock its behavior cheaply. Suggested location: `src/lib/__tests__/vehicleUtils.test.js` (matches existing convention, e.g. `distanceUtils.test.js`).

Minimal cases (plain objects + a `Map`, no mocking):
1. `nextStopName` resolves when the stop is present in `stopsMap` (`stopsMap.get(vehicle.nextStop)?.name`).
2. `nextStopName` is `undefined` when `vehicle.nextStop` is missing from `stopsMap` (the `?.` short-circuit; component falls back to `$t('NA')`).
3. Passthrough fields map correctly — note the field-source asymmetry: `nextDestination` comes from `activeTrip.tripHeadsign`, while `vehicleId`, `lastUpdateTime`, `predicted` come from `vehicle`.

VishalRaut2106:
Closes #501

## Description
This PR adds unit tests for the `buildVehiclePopupData` helper introduced in PR #471

## Changes
- Added `src/lib/__tests__/vehicleUtils.test.js`
- Added a test to verify `nextStopName` correctly resolves to the stop's name when it exists in `stopsMap`.
- Added a test to verify `nextStopName` safely falls back to `undefined` when the stop is missing from `stopsMap`.

### DIFF
--- a/src/lib/__tests__/vehicleUtils.test.js
+++ b/src/lib/__tests__/vehicleUtils.test.js
@@ -1,0 +1,51 @@
+import { describe, it, expect } from 'vitest';
+import { buildVehiclePopupData } from '$lib/vehicleUtils.js';
+
+describe('buildVehiclePopupData', () => {
+	it('returns popup data with stop name when stopsMap has the nextStop', () => {
+		const vehicle = {
+			vehicleId: 'v-123',
+			lastUpdateTime: 1600000000,
+			nextStop: 'stop-abc',
+			predicted: true
+		};
+		const activeTrip = {
+			tripHeadsign: 'Downtown'
+		};
+		const stopsMap = new Map();
+		stopsMap.set('stop-abc', { name: 'Main St & 1st Ave' });
+
+		const result = buildVehiclePopupData(vehicle, activeTrip, stopsMap);
+
+		expect(result).toEqual({
+			nextDestination: 'Downtown',
+			vehicleId: 'v-123',
+			lastUpdateTime: 1600000000,
+			nextStopName: 'Main St & 1st Ave',
+			predicted: true
+		});
+	});
+
+	it('returns popup data with nextStopName as undefined when stopsMap lacks the nextStop', () => {
+		const vehicle = {
+			vehicleId: 'v-456',
+			lastUpdateTime: 1600000100,
+			nextStop: 'stop-unknown',
+			predicted: false
+		};
+		const activeTrip = {
+			tripHeadsign: 'Uptown'
+		};
+		const stopsMap = new Map();
+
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Nice, surgical PR, Vishal. This is exactly the kind of change I like to see follow a feature: PR #471 introduced `buildVehiclePopupData`, and this locks its contract down cheaply with two focused tests. Both tests pass (~1ms), formatting is clean, and using `toEqual` on the whole returned object is the right call — it catches missing *and* extra keys, not just the one field under test. You covered both branches of the `stopsMap.get(vehicle.nextStop)?.name` short-circuit, which was the whole point of the issue.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (2 found — follow-up, not blocking)

- **Source-asymmetry test.** `nextDestination` is sourced from `activeTrip.tripHeadsign`, while `vehicleId`/`lastUpdateTime`/`predicted` come from `vehicle`. Your tests exercise this mapping, but because every value is a distinct string, a future source-swap bug (e.g. `nextDestination: vehicle.tripHeadsign`) would only be caught incidentally. A test where `vehicle` and `activeTrip` share field names with *different* values would make the contract explicit. [`src/lib/__tests__/vehicleUtils.test.js`]
- **Latent null `activeTrip`.** This is a note on the *function*, not your tests: `buildVehiclePopupData` assumes `activeTrip` is non-null, but at the call sites it comes from `activeTripMap.get(...)`, which can return `undefined` — in which case `activeTrip.tripHeadsign` throws. Worth a guard or a documented precondition in a future pass. [`src/lib/vehicleUtils.js:96`]

## Strengths

- Correct location and naming (`src/lib/__tests__/`, matching `distanceUtils.test.js`).
- Full-object `toEqual` assertions — robust against both missing and extra keys.
- Both branches of the optional-chaining lookup covered.
- `predicted: false` in the second case confirms falsy values pass through faithfully.
- No mocking, deterministic, fast — exactly the right weight for a pure helper.

## Recommended Action

Merge as-is. The two suggestions are genuine follow-ups, not changes this PR needs to make — tracked in [#508](https://github.com/OneBusAway/wayfinder/issues/508).

PR #504, opened 2026-06-04
fix: clear trip planner from/to pins from the map

### LINKED ISSUE #490: Bug: From/To pins linger on map after closing trip itineraries modal
When I plan a trip (enter From + To, hit "Plan your trip"), the itineraries show up, and the route is drawn on the map. But when I close the itineraries modal, the route polyline clears while the From and To pin markers stay floating on the map. The form inputs also stay filled in.

So I'm left with two pins on the map with no route between them, which looks a bit broken.

**What I think the fix should be:** Match how Google/Apple Maps behave on closing the itineraries modal:
 - Remove the From and To pin markers from the map (along with the route, which already clears)
 - Keep the From/To inputs populated in the search form, so the user can tweak options and re-plan with one click

Basically: closing the results = clean map + search still remembered. Pins come back when the user hits Plan again.

I considered also clearing the form inputs on close, or clearing them when "Plan trip" is pressed, but both feel worse since the user often wants to re-search with the same locations.

**SS:**

Image

tarunsinghofficial:
Fixes #490 

## What this fixes

Two related bugs with the trip planner pins (From and To markers) sticking around on the map when they shouldn't.

**Bug 1: pins stay after closing the itineraries modal**

When you planned a trip and then closed the itineraries modal, the route line cleared but the From and To pins stayed floating on the map. You were left with two pins and no route between them, which looked broken.

**Bug 2: pins stay after switching tabs** (this comes new apart from what I discussed in the issue #490)

When you switched from the Plan Trip tab back to Stops and Stations, the pins also stayed on the map. This happened because the plan tab can get unmounted before its cleanup runs, so the pins never got removed.

## How it works now

Closing the itineraries modal:
- Removes the From and To pins and the route from the map, so the map is clean
- Keeps the From and To inputs filled in, so you can tweak options and plan again in one click (same as Google and Apple Maps)
- The pins come back the next time you hit Plan your trip

Switching away from the Plan Trip tab:
- Clears the pins, the route, and the itinerary state right away, before the tab unmounts
- Added a safety net so the pins are removed even if the component is already gone

## Tests

- Unit tests for `clearTripPlanPins` covering both pins, one pin, and no pins.
- A SearchPane test that checks the map state is cleared when leaving the plan tab.
- All trip planner and search tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trip planner now properly clears map pins and resets planning state when closing the modal or switching away from the Plan Trip tab.
  * Improved lifecycle handling ensures map markers and trip itineraries are consistently cleaned up during state transitions.

### DIFF
--- a/src/components/search/SearchPane.svelte
+++ b/src/components/search/SearchPane.svelte
@@ -193,6 +193,11 @@
 
 	function handleTabSwitch() {
 		if (isContextMenuTrigger) return;
+
+		if (activeTab === 'plan') {
+			window.dispatchEvent(new CustomEvent('tripPlanModalClosed'));
+			clearTripItineraries();
+		}
 		const event = new CustomEvent('tabSwitched');
 		window.dispatchEvent(event);
 	}
--- a/src/components/search/__tests__/SearchPane.test.js
+++ b/src/components/search/__tests__/SearchPane.test.js
@@ -84,6 +84,7 @@ describe('SearchPane', () => {
 		// Mock props
 		mockProps = {
 			clearPolylines: vi.fn(),
+			clearTripItineraries: vi.fn(),
 			handleRouteSelected: vi.fn(),
 			handleViewAllRoutes: vi.fn(),
 			handleTripPlan: vi.fn(),
@@ -172,6 +173,26 @@ describe('SearchPane', () => {
 			await user.click(tripTab);
 			expect(global.dispatchEvent).toHaveBeenCalledWith(
 				expect.objectContaining({ type: 'planTripTabClicked' })
+			);
+		});
+
+		test('clears trip plan map state when leaving the plan tab', async () => {
+			const user = userEvent.setup();
+
+			render(SearchPane, { props: mockProps });
+
+			await user.click(screen.getByRole('tab', { name: 'Plan Trip' }));
+			vi.mocked(global.dispatchEvent).mockClear();
+			mockProps.clearTripItineraries.mockClear();
+
+			await user.click(screen.getByRole('tab', { name: 'Stops & Stations' }));
+
+			expect(global.dispatchEvent).toHaveBeenCalledWith(
+				expect.objectContaining({ type: 'tripPlanModalClosed' })
+			);
+			expect(mockProps.clearTripItineraries).toHaveBeenCalled();
+			expect(global.dispatchEvent).toHaveBeenCalledWith(
+				expect.objectContaining({ type: 'tabSwitched' })
 			);
 		});
 
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -221,12 +221,7 @@
 			const data = await fetchTripPlan(selectedFrom, selectedTo);
 
 			if (data) {
-				const tripPlanData = {
-					data,
-					fromMarker,
-					toMarker
-				};
-				handleTripPlan(tripPlanData);
+				handleTripPlan({ data });
 
 				// Save to recent trips
 				try {
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -134,12 +134,18 @@
 			fromPlace = '';
 			fromResults = [];
 			selectedFrom = null;
-			mapProvider.removePinMarker(fromMarker);
+			if (fromMarker) {
+				mapProvider.removePinMarker(fromMarker);
+				fromMarker = null;
+			}
 		} else {
 			toPlace = '';
 			toResults = [];
 			selectedTo = null;
-			mapProvider.removePinMarker(toMarker);
+			if (toMarker) {
+				mapProvider.removePinMarker(toMarker);
+				toMarker = null;
+			}
 		}
 		clearTripItineraries();
 	}
@@ -291,6 +297,12 @@
 	});
 
 	onDestroy(() => {
+		// Safety when the plan tab unmounts before tabSwitched runs.
+		if (mapProvider && (fromMarker || toMarker)) {
+			const result = clearTripPlanPins({ fromMarker, toMarker, mapProvider });
+			fromMarker = result.fromMarker;
+			toMarker = result.toMarker;
+		}
 		if (browser) {
 			if (tabSwitchedHandler) {
 				window.removeEventListener('tabSwitched', tabSwitchedHandler);
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -17,7 +17,7 @@
 	import { formatDepartureDisplay } from '$lib/dateTimeFormat';
 	import { env } from '$env/dynamic/public';
 	import { createRequestFromTripOptions, buildOTPParams, validateCoordinates } from '$lib/otp';
-	import { swapTripLocations } from '$lib/tripPlanUtils';
+	import { swapTripLocations, clearTripPlanPins } from '$lib/tripPlanUtils';
 	import { recentTrips } from '$stores/recentTripsStore';
 	import RecentTripsList from './RecentTripsList.svelte';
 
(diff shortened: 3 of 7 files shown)


aaronbrethorst (changes requested):
Hi Tarun — thanks for this. Going beyond the original issue to also catch the tab-switch case (Bug 2) was a good instinct, and the writeup makes the two scenarios easy to follow. I especially like that you extracted the clearing logic into a small pure `clearTripPlanPins` helper and unit-tested its three branches — that's exactly the right testability move, and it keeps the null-guarded `removePinMarker` calls in one place. Moving the From/To marker ownership entirely into `TripPlan.svelte` and dropping the now-unused `$state` from `+page.svelte` is a real simplification too.

Local tests and lint both pass (`26 passed` across the two affected test files; Prettier + ESLint clean).

One thing held me back from a straight merge: the two mechanisms that actually fix the bugs — the `tripPlanModalClosed` listener and the `onDestroy` safety net — have no test that exercises them through the component. Details below.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

- **The new pin-clearing wiring has no regression coverage.** The pure helper `clearTripPlanPins` is well tested, and the `SearchPane` test verifies that leaving the Plan tab *dispatches* `tripPlanModalClosed` (against a mocked `dispatchEvent`). But nothing verifies that `TripPlan.svelte` actually removes the live pins in response — neither the `handleTripPlanModalClosed` listener [src/components/trip-planner/TripPlan.svelte:252-256] nor the `onDestroy` safety net [src/components/trip-planner/TripPlan.svelte:300-305]. The `onDestroy` block is the entire fix for Bug 2 (pins lingering after a tab switch when the component unmounts first), and if someone deletes it every existing test still passes while the bug silently returns. There's no `TripPlan.svelte` component test today (only `TripPlanSwap`/`TripPlanSearchField`/`TripPlanPinMarker`/`RecentTripsList`), so this is a new file. A focused test that renders `TripPlan`, selects From/To, then (a) dispatches `window.dispatchEvent(new CustomEvent('tripPlanModalClosed'))` and asserts `mapProvider.removePinMarker` fired for both markers *and* `selectedFrom`/`fromPlace` survive, and (b) unmounts the component with active markers and asserts the pins are removed, would close the gap. That one file covers the heart of both fixes.

## Suggestions (3 found)

- **Dead payload in `planTrip`.** `planTrip` still packs `fromMarker`/`toMarker` into `tripPlanData` [src/components/trip-planner/TripPlan.svelte:224-228], but `handleTripPlan` no longer reads them (you correctly removed the consumption and the matching `@param` lines in `+page.svelte`). Since this PR is what made those fields dead, it'd be clean to drop them from the object here so the producer matches the consumer.

- **`tripPlanModalClosed` is dispatched twice on a tab switch.** `SearchPane.handleTabSwitch` dispatches it [src/components/search/SearchPane.svelte:197-200] and then dispatches `tabSwitched`, which triggers `+page.handleTabSwitched`, which dispatches `tripPlanModalClosed` again [src/routes/+page.svelte:235-238]. It's harmless today because `clearTripPlanPins` is idempotent and null-guarded, but it's worth picking a single owner of the teardown so the flow has one source of truth. Not blocking.

- **Defensive consistency on `mapProvider`.** `onDestroy` guards with `if (mapProvider && (fromMarker || toMarker))` [src/components/trip-planner/TripPlan.svelte:301], but `handleTripPlanModalClosed` calls `clearTripPlanPins` with no `mapProvider` check, and `clearTripPlanPins` dereferences it unguarded [src/lib/tripPlanUtils.js:59-60]. In practice a marker can only exist if `mapProvider` was non-null when it was created, so this is effectively unreachable — but matching the `onDestroy` guard (or guarding inside the helper) would make the contract obvious. Optional; the JSDoc could alternatively state that `mapProvider` must be non-null.

## Strengths

- Clean extraction of `clearTripPlanPins` as a pure, exhaustively unit-tested helper (both/one/none marker branches plus return shape).
- The `SearchPane` integration test drives the real user flow (click Plan Trip, then click away) rather than calling the handler directly.
- Good defensive null-checks added to `clearInput` before `removePinMarker` [src/components/trip-planner/TripPlan.svelte:137-148].
- Comments earn their place — the `onDestroy` comment explains the non-obvious "why" (unmount-before-`tabSwitched`), and the `clearTripPlanPins` JSDoc documents what it deliberately does *not* touch (form inputs/coordinates) and why (matching Google/Apple Maps).
- Removing the unused `fromMarker`/`toMarker` `$state` from `+page.svelte` and the dead `closePane` wiring is a nice tightening.

## Recommended Action

1. Add a `TripPlan.svelte` component test covering the `tripPlanModalClosed` listener and the `onDestroy` pin cleanup (the Important issue).
2. While you're in there, drop the dead `fromMarker`/`toMarker` from `planTrip`'s `tripPlanData`.
3. Consider consolidating the double `tripPlanModalClosed` dispatch and the `mapProvider` guard (both optional).
4. Re-request review after the test lands.

**Verdict: Request Changes.**

tarunsinghofficial:
Thanks @aaronbrethorst for the feedback. I have addressed all of them:

- Added a TripPlan.svelte component test covering both fixes: the tripPlanModalClosed listener (removes both pins, keeps the form inputs) and the onDestroy safety net (removes pins on unmount). Delete either mechanism and a test fails now.
- Dropped the dead fromMarker/toMarker from planTrip's tripPlanData so the producer matches the consumer.
- Consolidated the double tripPlanModalClosed dispatch: SearchPane is now the single owner of teardown when leaving the Plan tab, and +page.handleTabSwitched just closes the modal.
- Guarded mapProvider inside clearTripPlanPins so it matches the onDestroy guard.

All tests are passed.

aaronbrethorst (approved):
Tarun — this is exactly the follow-up I was hoping for. Every item from the last pass is addressed, and the headline fix — adding regression coverage for the two mechanisms that actually do the work — is done properly. I checked your "delete either mechanism and a test fails now" claim the hard way: I removed the `onDestroy` safety net and the unmount test went red; I removed the `tripPlanModalClosed` listener registration and the modal-closed test went red. Both tests are real guards, not coincidental passes. That's the kind of test I wanted here.

Verification on `67b6531`:
- `npx vitest run` on the three affected files: **28 passed**.
- Prettier + ESLint: clean.
- Mutation check: removing either fix mechanism fails a `TripPlan.test.js` test (confirmed manually).
- No dangling consumers: `TripPlanModal.svelte` does not read `fromMarker`/`toMarker`, so dropping those props is safe; the only remaining references live in `TripPlan.svelte` (owner) and `tripPlanUtils.js` (helper).

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None. The previously-missing regression coverage now exists in `src/components/trip-planner/__tests__/TripPlan.test.js`, exercising both the `tripPlanModalClosed` listener (removes both pins, keeps the form inputs) and the `onDestroy` safety net (removes pins on unmount).

## Suggestions (0 found)

All three suggestions from the prior review are resolved:
- **Dead payload removed.** `planTrip` now calls `handleTripPlan({ data })` [src/components/trip-planner/TripPlan.svelte:224] and `+page.svelte` dropped the `fromMarker`/`toMarker` `$state` plus the stale `@param` lines.
- **Double dispatch consolidated.** `SearchPane.handleTabSwitch` is now the single owner of trip-plan teardown when leaving the Plan tab [src/components/search/SearchPane.svelte:197-200]; `+page.handleTabSwitched` just closes the modal, with a comment explaining the ownership split [src/routes/+page.svelte:236].
- **`mapProvider` guard added.** `clearTripPlanPins` now guards `if (mapProvider)` [src/lib/tripPlanUtils.js:59], matching the `onDestroy` guard.

## Strengths

- The new `TripPlan.svelte` test drives the real marker-creation path via the `setTripPlanLocation` event instead of mocking the geocode fetch — it tests behavior, not implementation, and the helper comment explains why.
- Asserting that the form inputs survive (`#from-location-input` / `#to-location-input` retain their values) locks in the deliberate Google/Apple-Maps-style behavior, not just that pins were removed.
- Teardown now has one clear owner, and the `+page.svelte` comment documents the split so the next person doesn't re-introduce a duplicate dispatch.
- Clean producer/consumer alignment: the dead marker payload is gone from both ends.

## Verdict

Merge.

PR #506, opened 2026-06-05
fix(a11y): announce loading state in LoadingSpinner for screen readers

### LINKED ISSUE #505: Add aria-live to inline LoadingSpinner

tarunsinghofficial:
Fixes #505 #405 

## Summary by CodeRabbit

* **Accessibility**
  * Loading spinner now features ARIA accessibility attributes and live region announcements, enabling screen readers to accurately communicate loading state to users. Includes full localization support with translated announcements across multiple languages.

* **Tests**
  * Test suite updated to comprehensively verify accessibility features function correctly across language localization and assistive technology scenarios.

### DIFF
--- a/src/components/LoadingSpinner.svelte
+++ b/src/components/LoadingSpinner.svelte
@@ -3,6 +3,9 @@
 </script>
 
 <div
+	aria-label={$t('loading')}
+	role="status"
+	aria-live="polite"
 	class="absolute inset-0 z-50 flex items-center justify-center bg-neutral-800 bg-opacity-80 md:rounded-lg"
 >
 	<div class="flex items-center text-white">
--- a/src/components/__tests__/LoadingSpinner.test.js
+++ b/src/components/__tests__/LoadingSpinner.test.js
@@ -1,18 +1,72 @@
-import { render, screen } from '@testing-library/svelte';
-import { expect, test, describe } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/svelte';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 import LoadingSpinner from '../LoadingSpinner.svelte';
+import { locale } from 'svelte-i18n';
+
+vi.mock('svelte-i18n', () => {
+	let currentLocale = 'en';
+	const tSubscribers = [];
+	const translations = {
+		en: {
+			loading: 'Loading'
+		},
+		es: {
+			loading: 'Cargando'
+		}
+	};
+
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Nice, focused fix, Tarun. This does exactly what issue #505 asks for and nothing more: the inline `LoadingSpinner` now carries `role="status"`, `aria-live="polite"`, and a localized `aria-label`, so screen reader users finally get told something is happening. It mirrors the pattern already established in `FullPageLoadingSpinner.svelte` to the letter, which is exactly what I want to see — consistency over cleverness.

The test rewrite is the real star here. Swapping in a local `svelte-i18n` mock that actually flips `en`→`es` turns what could have been a tautological "the mock returns the key" check into a genuine reactivity test: it proves both the `aria-label` *and* the visible text re-render on a locale change. That's a meaningfully better test than what it replaced.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None. Tests pass (`npx vitest run` — 4/4 green), `npm run lint` is clean, and the accessible-name assertions hold up under scrutiny: removing the `aria-label` from the component would actually fail the Spanish test, since `aria-label` wins over text content for the computed accessible name. The coverage is real, not decorative.

## Suggestions (1 found)

- The inner `<svg>` is purely decorative but isn't marked `aria-hidden="true"` [`src/components/LoadingSpinner.svelte:12`]. In practice the parent `role="status"` already owns the accessible name via `aria-label`, so the SVG contributes nothing to assistive tech and this is belt-and-suspenders. It also applies equally to the existing `FullPageLoadingSpinner.svelte`, so it's not a regression you introduced. Out of scope for this PR — tracked separately (see below) rather than gating the merge.

## A note on the CodeRabbit comment

The automated suggestion to import `t` from a `src/lib/i18n.js` wrapper is a false alarm — `i18n.js` exports `languages`, `getInitialLocale`, and `isRTL`, but **not** `t`. Importing `t` directly from `svelte-i18n` is the established convention across the codebase (`ArrivalDeparture`, `AlertsModal`, `SearchPane`, and ~15 others). Your import is correct; no change needed.

## Strengths

- Implementation is minimal and matches the sibling component exactly.
- Satisfies every acceptance-criteria checkbox in issue #505.
- Test mock is a legitimate, reusable fixture that makes locale reactivity genuinely testable.
- `beforeEach(() => locale.set('en'))` properly isolates the mutable mock state between tests.
- `waitFor` correctly handles the async reactivity flush after `locale.set('es')`.

## Recommended Action

Merge as-is. The single suggestion above is fit-and-finish and belongs in a follow-up, not this PR.

PR #507, opened 2026-06-05
feat: add reset to defaults in trip options modal

### LINKED ISSUE #491: Feat: add "Reset to defaults" button in Trip Options modal
Right now, once a user changes trip options (departure time, wheelchair, fewest transfers, walk distance, distance unit), the only way to get back to defaults is to manually flip each one. If they planned a trip with a bunch of custom options and now want a fresh search, this is annoying.

**What I'd like to add:**
A "Reset to defaults" button in the Trip Options modal, probably in the header (next to Cancel) or at the bottom of the content area, styled subtly so it doesn't compete with Done.

**Clicking it would:**

- Reset all options back to defaults (Leave now, no wheelchair, fastest, 1 mile, auto unit)
- Clear the option pills shown above the Plan trip button
- Clear the persisted values in localStorage too

tarunsinghofficial:
Fixes #491 

## What this does

Add a "Reset to defaults" button to the Trip Options modal that restores all options to their default values.

## Why it was needed

After changing several options, the only way to start fresh was to undo each setting manually. That is annoying when you want a clean search with default routing again.

## How it works

- Reset is draft-only, like every other control in the modal. handleReset() only updates local $state copies — nothing is persisted until the user clicks Done.
- Cancel discards the reset and keeps the previously saved preferences untouched.
- Done commits the reset: persisted options are written back to defaults via setPersisted, which clears localStorage keys when the value equals the default (forward-compatible — users track future default changes rather than being pinned to today's values).
- resetAll() was removed from the store; it was dead code never wired to any UI.

## How to test

1. Plan Trip → Options → change wheelchair, optimization, walk distance, etc. → Done (pills show).
2. Reopen Options → click **Reset to defaults**.
3. Confirm UI shows defaults and pills are gone.
4. Reload the page and confirm options stayed at defaults.

## SS:
Screenshot 2026-06-05 at 11 12 08 PM

### DIFF
--- a/src/components/trip-planner/TripOptionsModal.svelte
+++ b/src/components/trip-planner/TripOptionsModal.svelte
@@ -75,9 +75,9 @@
 		}
 	}
 
-	// Reset everything back to defaults immediately: clears the store and localStorage
+	// Reset the local editing copies back to defaults. Like every other control
+	// in this modal, this is draft-only: nothing is persisted until handleDone() runs, so Cancel still discards the reset and keeps the saved preferences
 	function handleReset() {
-		tripOptions.resetAll();
 		departureType = DEFAULT_TRIP_OPTIONS.departureType;
 		departureTime = DEFAULT_TRIP_OPTIONS.departureTime || '';
 		departureDate = DEFAULT_TRIP_OPTIONS.departureDate || '';
@@ -407,7 +407,7 @@
 				<button
 					type="button"
 					onclick={handleReset}
-					class="text-sm font-medium text-gray-500 underline-offset-2 hover:text-gray-700 hover:underline focus:outline-none dark:text-gray-400 dark:hover:text-gray-200"
+					class="rounded text-sm font-medium text-gray-500 underline-offset-2 hover:text-gray-700 hover:underline focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 dark:text-gray-400 dark:hover:text-gray-200"
 				>
 					{$t('trip-planner.reset_to_defaults')}
 				</button>
--- a/src/components/trip-planner/TripOptionsModal.svelte
+++ b/src/components/trip-planner/TripOptionsModal.svelte
@@ -6,6 +6,7 @@
 		effectiveDistanceUnit,
 		getWalkDistanceOptions,
 		snapToClosestOption,
+		DEFAULT_TRIP_OPTIONS,
 		UNIT_METRIC,
 		UNIT_IMPERIAL
 	} from '$stores/tripOptionsStore';
@@ -72,6 +73,18 @@
 			departureTime = getCurrentTimeForInput(regionTz);
 			departureDate = getTodayDateForInput(regionTz);
 		}
+	}
+
+	// Reset everything back to defaults immediately: clears the store and localStorage
+	function handleReset() {
+		tripOptions.resetAll();
+		departureType = DEFAULT_TRIP_OPTIONS.departureType;
+		departureTime = DEFAULT_TRIP_OPTIONS.departureTime || '';
+		departureDate = DEFAULT_TRIP_OPTIONS.departureDate || '';
+		wheelchair = DEFAULT_TRIP_OPTIONS.wheelchair;
+		optimize = DEFAULT_TRIP_OPTIONS.optimize;
+		maxWalkDistance = DEFAULT_TRIP_OPTIONS.maxWalkDistance;
+		distanceUnit = DEFAULT_TRIP_OPTIONS.distanceUnit;
 	}
 </script>
 
@@ -387,6 +400,17 @@
 						{/if}
 					</button>
 				</div>
+			</div>
+
+			<!-- Reset to defaults -->
+			<div class="mt-4 flex justify-center">
+				<button
+					type="button"
+					onclick={handleReset}
+					class="text-sm font-medium text-gray-500 underline-offset-2 hover:text-gray-700 hover:underline focus:outline-none dark:text-gray-400 dark:hover:text-gray-200"
+				>
+					{$t('trip-planner.reset_to_defaults')}
+				</button>
 			</div>
 		</div>
 	</div>
--- a/src/components/trip-planner/TripOptionsModal.svelte
+++ b/src/components/trip-planner/TripOptionsModal.svelte
@@ -76,7 +76,8 @@
 	}
 
 	// Reset the local editing copies back to defaults. Like every other control
-	// in this modal, this is draft-only: nothing is persisted until handleDone() runs, so Cancel still discards the reset and keeps the saved preferences
+	// in this modal, this is draft-only: nothing is persisted until handleDone()
+	// runs, so Cancel still discards the reset and keeps the saved preferences.
 	function handleReset() {
 		departureType = DEFAULT_TRIP_OPTIONS.departureType;
 		departureTime = DEFAULT_TRIP_OPTIONS.departureTime || '';
--- a/src/components/trip-planner/__tests__/TripOptionsModal.test.js
+++ b/src/components/trip-planner/__tests__/TripOptionsModal.test.js
@@ -2,6 +2,11 @@ import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { render, screen } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
 import TripOptionsModal from '../TripOptionsModal.svelte';
+import { tripOptions, DEFAULT_TRIP_OPTIONS } from '$stores/tripOptionsStore';
+
+// browser = true so the real store's setPersisted touches the (mocked)
+// localStorage, letting us assert the commit/draft-only persistence behavior.
+vi.mock('$app/environment', () => ({ browser: true }));
 
 // Only the translation layer is mocked; the real tripOptionsStore is used so the test exercises the actual reset wiring (handleReset + DEFAULT_TRIP_OPTIONS)
 vi.mock('svelte-i18n', () => {
@@ -40,11 +45,22 @@ function getOptionButton(label) {
 	return screen.getByText(label).closest('button');
 }
 
+function getStoreValue() {
+	let value;
+	tripOptions.subscribe((v) => (value = v))();
+	return value;
+}
+
 describe('TripOptionsModal reset', () => {
 	let user;
 
 	beforeEach(() => {
 		user = userEvent.setup();
+		// Start each test from a clean, default store and clear localStorage spies.
+		tripOptions.set({ ...DEFAULT_TRIP_OPTIONS });
+		localStorage.setItem.mockClear();
+		localStorage.removeItem.mockClear();
+		localStorage.getItem.mockClear();
 	});
 
 	it('resets all locally edited options back to defaults', async () => {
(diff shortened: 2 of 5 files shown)


aaronbrethorst (changes requested):
Hi Tarun — this is a genuinely useful quality-of-life addition, and the implementation is small and focused. I especially like that you leaned on the store's existing `resetAll()` rather than reinventing the persistence logic, and that you added a store test that correctly flips `browser = true` to exercise the localStorage paths (that mock override trips people up constantly). The PR description and screenshot make the intent crystal clear too.

There are a couple of things I'd like resolved before this merges — one is an objective accessibility regression, and one is a UX consistency question worth a deliberate decision.

## Critical Issues (0 found)

None — the change is functionally correct, tests pass, and lint is clean.

## Important Issues (2 found)

- **The Reset button removes the keyboard focus indicator with no replacement** [`src/components/trip-planner/TripOptionsModal.svelte:410`]. The class uses `focus:outline-none` but, unlike every other control in this modal, adds no replacement ring. The text/date inputs and the walk-distance select all pair `focus:outline-none` with `focus:ring-1 focus:ring-blue-500` (lines 211, 217, 324), and the header/section buttons keep the native outline entirely. As written, a keyboard user who tabs to "Reset to defaults" gets no visible focus — `hover:underline` only fires on mouse hover. That's a WCAG 2.4.7 failure, and we've been actively tightening a11y in this area recently. Fix: drop `focus:outline-none`, or replace it, e.g. `focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-blue-500`.

- **Reset commits immediately, which means Cancel can't undo it** [`src/components/trip-planner/TripOptionsModal.svelte:79-88`]. Every other control in this modal is draft-only: edits live in local `$state` and are persisted only by `handleDone()` (lines 54-67), while Cancel discards them. `handleReset()` breaks that contract — `tripOptions.resetAll()` runs `set(defaults)` and wipes localStorage immediately (`tripOptionsStore.js:95-103`), so a user who clicks Reset and then Cancel has already lost their saved preferences with no undo. I know issue #491 asked to clear localStorage, and your PR description calls this out as intentional, so I'm not insisting on a behavior change — but the surprise is real. My preference would be to make Reset draft-only (reset just the local `$state` copies and let the existing `handleDone()` path persist them on Done), which keeps Cancel a true escape hatch and matches the rest of the modal. If you want to keep the immediate-commit behavior, please confirm that's a deliberate call and we'll leave it.

## Suggestions (3 found)

- **The new `handleReset()` logic has no test coverage** [`src/components/trip-planner/TripOptionsModal.svelte:79-88`]. The added store test is good, but it covers `resetAll()` — which already existed before this PR — not the new component wiring. The dual-write in `handleReset()` (reset the store *and* manually re-assign seven local `$state` copies) is exactly the kind of code that drifts: add a new option later, wire it into `handleDone()` but forget `handleReset()`, and reset silently leaves it stale with no test to catch it. There's an established `@testing-library/svelte` pattern in `src/components/trip-planner/__tests__/` — a test that toggles a few options, clicks Reset, and asserts the UI returns to defaults would lock this down.

- **The defaults assertion skips `maxWalkDistance`** [`src/stores/__tests__/tripOptionsStore.test.js:33-41`]. The `toMatchObject` pins six of the seven default fields but omits `maxWalkDistance`, which `handleReset()` reads at line 86 and whose default (`DEFAULT_WALK_DISTANCE_METERS = 1609`) is a deliberately non-obvious value. One added line — `maxWalkDistance: 1609` — guards against a silent default change.

- **`DEFAULT_TRIP_OPTIONS = defaults` exports a live mutable reference** [`src/stores/tripOptionsStore.js:35`]. It's the same object `resetAll()` passes to `set(defaults)`, so the store's value and the exported constant become the same instance. It's safe today (every writer spreads into a new object and the modal only reads scalars off it), but it's a latent footgun — any future in-place mutation would corrupt the canonical defaults. Cheap hardening: export `{ ...defaults }` (and/or `Object.freeze`). Not blocking.

## Strengths

- Reused the existing `resetAll()` store method instead of duplicating persistence logic.
- The store test correctly overrides the global `$app/environment` mock to `browser: true` and isolates modules with `vi.resetModules()` + dynamic import — both easy to get wrong.
- Svelte 5 runes usage is correct: `handleReset()` reassigns top-level `$state` variables, which is the right way to update rune state.
- Clear PR description with explicit test steps and a screenshot.

## Recommended Action

1. Add a focus indicator to the Reset button (Important).
2. Decide on the Reset/Cancel interaction — make it draft-only, or confirm the immediate-commit is intentional (Important).
3. Consider a component test for `handleReset()` and the one-line `maxWalkDistance` assertion (Suggestions).
4. Re-run the review after changes.

tarunsinghofficial:
Thanks, @aaronbrethorst, for the review. 

 I have addressed the feedback, a visible keyboard focus ring on the Reset button (was missing, a11y).
- Switched Reset from immediate commit to draft-only so Cancel works as a real undo.
- Added a component test that changes a few options, clicks Reset, and checks that the UI goes back to defaults.
- Added the maxWalkDistance default to the store test.
- Exported DEFAULT_TRIP_OPTIONS as a frozen copy so it can't be mutated by accident.

aaronbrethorst (changes requested):
Hi Tarun — thanks for this. It's a genuinely nice little feature and the implementation gets the hard part right: the reset is *draft-only*, so it slots cleanly into the modal's existing "nothing happens until you hit Done, Cancel throws it all away" contract instead of fighting it. Exporting a frozen `DEFAULT_TRIP_OPTIONS` as the single source of truth is the right instinct, the Svelte 5 runes usage is idiomatic, and you wrote tests instead of leaving that to someone else. Formatting and lint are clean, and the four new tests pass locally.

There's one behavioral gap worth resolving before this merges, plus a test that promises more than it checks. Details below.

## Critical Issues (0 found)

None. The feature is functionally correct and there are no crashes or regressions.

## Important Issues (2 found)

- **Reset → Done rewrites the defaults into `localStorage` instead of clearing it — which both misses issue #491's literal ask and sets a forward-compatibility trap.** `handleReset()` resets the draft (`TripOptionsModal.svelte:80-88`), then `handleDone()` (`TripOptionsModal.svelte:54-67`) calls `setPersisted()` for each key. For the three non-null defaults that takes the `setItem` branch (`tripOptionsStore.js:113-117`), writing `tripOptions_wheelchair='false'`, `tripOptions_optimize='fastest'`, `tripOptions_maxWalkDistance='1609'`; only `distanceUnit=null` gets removed. Issue #491 explicitly asked to "clear the persisted values in localStorage too." It's behaviorally equivalent *today* — on next load the store reads those values and computes the same effective state — but a user who clicks "Reset to defaults" is now **pinned** to today's defaults: if a future release changes, say, `DEFAULT_WALK_DISTANCE_METERS`, that user keeps the old value because it was written explicitly, while a user with no key picks up the new default. That quietly defeats the point of "reset to defaults." Make Reset actually clear the persisted keys (so reset means "track the defaults," not "freeze the current ones").

- **The "draft-only" test doesn't actually verify the draft-only guarantee.** `TripOptionsModal.test.js:77-86` ("does not persist the reset until Done") only asserts `expect(onClose).toHaveBeenCalled()`. It never seeds the store with non-default values before opening the modal, and never reads store/localStorage state after Cancel — so it would pass even if Reset wrote straight to localStorage, or even if Cancel persisted everything. The PR's central promise is the one thing left unproven. Seed the store with non-default persisted values, render, click Reset, click Cancel, then assert the store still holds the non-default values **and** that `localStorage.setItem`/`removeItem` were not called during reset/cancel.

## Suggestions (4 found)

- **Add a test for the real commit path: Reset → Done → store + localStorage.** This is the feature's *only* persistence path and it currently has zero coverage. Note the store test that looks like it covers reset — `tripOptionsStore.test.js:44-66` — exercises `resetAll()`, which the feature never calls (`handleReset` only touches draft `$state`). A real test would click "Reset to defaults", click "Done", then assert the store equals defaults and the persisted keys were handled as you decide to resolve the Important issue above. [`TripOptionsModal.test.js`]

- **`resetAll()` is dead production code, and now there are two diverging definitions of "reset."** `resetAll()` (`tripOptionsStore.js:95-103`) has no production callers (it's referenced only by its own test) — that predates this PR, but you've now added a second reset path with *different* localStorage semantics: `resetAll()` does `removeItem` for all four keys, while Reset → Done does `setItem`. `handleReset` correctly can't just call `resetAll()` (it persists immediately and would break Cancel-discards), but the divergence is a maintenance smell. Consider either deleting `resetAll()` or refactoring so both paths share one definition of how defaults map to storage.

- **The "resets all" test only covers three of the reset fields.** `TripOptionsModal.test.js:50-75` toggles and re-checks wheelchair, departureType, and optimize, but `handleReset` also resets `maxWalkDistance` and `distanceUnit`. Change the unit (Auto→Metric/Imperial) and walk distance to a non-default, then reset, and assert they revert — these two also interact via `handleUnitChange`'s snapping, so they're worth exercising together. (Lower priority: `departureTime`/`departureDate` clearing back to `''` also isn't asserted.)

- **Comment polish.** The `DEFAULT_TRIP_OPTIONS` comment (`tripOptionsStore.js:34`) attributes the wrong cause: protection of the internal `defaults` object comes from the spread `{ ...defaults }` (already a separate object), not from `Object.freeze` — what the freeze actually buys you is stopping consumers from mutating *this exported snapshot*. It's also a ~230-char single line, inconsistent with the wrapped comments elsewhere in the same file. Same single-line/missing-period nit on the `handleReset` comment (`TripOptionsModal.svelte:78-79`), whose content is otherwise accurate and worth keeping.

## Strengths

- Draft-only design is the correct architectural choice and is consistent with the rest of the modal.
- `DEFAULT_TRIP_OPTIONS` as a frozen single source of truth, guarded by a shape test.
- Tests assert on user-visible behavior (✓ markers, `aria-checked`, button text) rather than internal `$state`, so they survive refactors.
- Using the real store instead of mocking it is the right call — it makes the persistence assertions above cheap to add.
- The store-test `beforeEach` (module re-import + `browser = true` override) is a clean isolation pattern, and the comment explaining *why* `$app/environment` is overridden is a good "why" comment.
- Subtle, accessible button styling (focus-visible ring, hover states, dark mode) that doesn't compete with Done.

## Recommended Action

1. Resolve the localStorage behavior so Reset actually clears the persisted keys (or, if you and the team decide equivalence is acceptable, say so explicitly in the PR and accept the pinned-default tradeoff).
2. Strengthen the Cancel/draft-only test to seed non-default state and assert the store is left unmutated.
3. Add a Reset → Done persistence test covering the real commit path.
4. Consider the suggestions (dead `resetAll()`, remaining field coverage, comment polish).

**Verdict: Request Changes.**

aaronbrethorst (changes requested):
Hi Tarun — thanks for this. The "Reset to defaults" button is a genuinely nice quality-of-life addition, and the implementation is thoughtful in a way that isn't obvious at first glance: making the reset *draft-only* (committed by Done, discarded by Cancel) so it behaves exactly like every other control in the modal is the right call, and the tests that lock that invariant in are excellent. The `setPersisted` "clear the key when it equals the default" change is also a smart, forward-compatible touch.

I traced the persistence logic carefully and it's correct — for every persisted key, the load-time fallback equals `defaults[key]`, so clearing on reset never produces a surprising value. Removing `resetAll()` is safe too; it was dead code, never wired to any UI. Tests pass (7/7) and formatting is clean.

That said, I'd like a couple of things tightened before this merges.

## Critical Issues (0 found)

None. No correctness bugs in the shipped code.

## Important Issues (2 found)

1. **The PR description contradicts the code it's shipping.** The "How it works" section says *"The store already had `tripOptions.resetAll()` … It was just not wired to the UI"* and *"Reset is immediate and persistent: if you reset then hit Cancel, defaults still apply."* Both are false for the final diff: `resetAll()` is **removed** (`src/stores/tripOptionsStore.js`), and reset is **draft-only** — `handleReset()` only mutates local `$state`, and `handleCancel()` persists nothing, so Cancel *discards* the reset (`TripOptionsModal.svelte:50-52`, `:81-89`). Your own code comment at `:78-80` says this correctly; the description just describes an earlier, discarded design. Please update it (and the CodeRabbit "clears previously saved overrides" line, which is only true on Done) so the merge record isn't misleading.

2. **Test 3's key-clear assertions pass for the wrong reason** [`src/components/trip-planner/__tests__/TripOptionsModal.test.js:127-155`]. The seed only overrides `wheelchair` and `optimize`, leaving `maxWalkDistance` (1609) and `distanceUnit` (null) at their defaults. Because `setPersisted` calls `removeItem` for a default value regardless of whether reset touched it, the `removeItem` assertions for `maxWalkDistance` and `distanceUnit` (`:154-155`) would still pass even if `handleReset` never reset those fields. Seed them to non-defaults (e.g. `maxWalkDistance: 4828`, `distanceUnit: 'metric'`) so those two assertions actually depend on the reset.

## Suggestions (3 found)

- **Cover the departure reset→commit path** [`TripOptionsModal.svelte:56-58`, `:82-84`]. `handleReset` now resets `departureType/departureTime/departureDate`, but no test asserts the *committed* session fields, and the non-`'now'` branch of `handleDone`'s ternary is unexercised. A test that seeds `{ departureType: 'departAt', departureTime: '14:30', departureDate: '2026-08-01' }`, resets, hits Done, then asserts the store has `departureType === 'now'` and both time/date `=== null` would close that gap. (Folding this into the fix for Important #2 is natural.)
- **Fix a stale test comment** [`TripOptionsModal.test.js:11`]. It says *"Only the translation layer is mocked,"* but the file also mocks `$app/environment` (`:9`). Reword to acknowledge the browser-env override.
- **(Out of scope — noting for awareness)** After reset, `maxWalkDistance` is 1609 (a mile) while `distanceUnit` stays auto. In a metric-default region 1609 isn't one of the walk-distance `<option>` values, so the `<select>` briefly has no matching option. This is pre-existing — a fresh user in a metric region already lands on 1609 — so it's not this PR's problem, but it's worth a future look at default-unit handling.

## Strengths

- Draft-only semantics are the correct model, and the Cancel test (`:103-125`) nails it by asserting both store state *and* that neither `setItem` nor `removeItem` fired.
- Tests drive the real store through the DOM (only i18n mocked), so the reset wiring and persistence contract are exercised end-to-end rather than against a mock.
- `DEFAULT_TRIP_OPTIONS` as a frozen, spread copy with a clear "why" comment is a clean way to share the source of truth with the modal.
- The `setPersisted` forward-compatibility change is a nice bit of foresight, and its comment explains the reasoning well.

## Recommended Action

1. Update the PR description to match the shipped design (resetAll removed; reset is draft-only, Cancel discards).
2. Strengthen Test 3 so the `maxWalkDistance`/`distanceUnit` clear assertions aren't vacuous, ideally adding the departure reset→commit assertions in the same test.
3. Consider the two remaining suggestions.
4. Re-run tests after the changes.

**Verdict: Request Changes** — all low-effort, and the code underneath is solid.

aaronbrethorst (approved):
Tarun — this is fully resolved. Every item from the last round has been addressed cleanly, and the resolutions are exactly the kind I like to see: not just "made the assertion pass" but "made the assertion _mean something_." Nice work sticking with it across the rounds.

Reviewing `0676f2b` against base `develop`. Tests pass (7/7), Prettier is clean, and the true PR diff is tight — one locale line, the store change, the modal button/handler, and two test files.

## Resolution of prior feedback

- **Important #1 — PR description matched to shipped design.** ✅ The body now correctly states `resetAll()` was removed as dead code and that reset is draft-only (Cancel discards, Done commits). The merge record is no longer misleading.
- **Important #2 — Test 3's clear-assertions are no longer vacuous.** ✅ `TripOptionsModal.test.js:138-139` now seeds `maxWalkDistance: 4828` and `distanceUnit: 'metric'`, so the `removeItem` assertions for those keys genuinely depend on `handleReset` touching them.
- **Suggestion — departure reset→commit path covered.** ✅ Test 3 seeds `{ departureType: 'departAt', departureTime: '14:30', departureDate: '2026-08-01' }`, resets, hits Done, and asserts the committed store has `departureType === 'now'` with time/date `null` (`:152-154`). That's exactly the gap I flagged.
- **Suggestion — stale test comment fixed.** ✅ Line 11 now reads "Translation layer and browser env are mocked."

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (0 blocking)

- The only open item is the pre-existing, out-of-scope one I raised last time: after reset in a metric-default region, `maxWalkDistance` (1609m) isn't one of the metric `<select>` options. Not this PR's problem — tracked separately in #545.

## Strengths

- The `setPersisted` "clear the key when it equals the default" change plus the new `tripOptionsStore.test.js` locks in the forward-compatibility contract directly at the store level — good instinct to test it there, not just through the modal.
- Draft-only semantics remain correct and are pinned by the Cancel test asserting neither `setItem` nor `removeItem` fired.
- Tests drive the real store through the DOM, so the reset wiring and persistence contract are exercised end-to-end.
- The store comment explaining _why_ writing a default clears the key is genuinely useful for the next reader.

## Note on tracking

The pre-existing metric-region default-unit mismatch is tracked in #545 for a future pass on default-unit handling.

## Recommended Action

Merge. Follow-up for the default-unit mismatch is tracked in #545.

PR #511, opened 2026-06-08
fix: add null safeguard and source-asymmetry test for buildVehiclePopupData

### LINKED ISSUE #508: Follow-up: harden buildVehiclePopupData tests and guard against null activeTrip
Follow-up from PR #503, which added focused tests for `buildVehiclePopupData` (closing #501).

Two optional improvements were identified during review but are out of scope for that narrowly-scoped test PR:

1. **Source-asymmetry test.** `buildVehiclePopupData` sources `nextDestination` from `activeTrip.tripHeadsign` while `vehicleId`/`lastUpdateTime`/`predicted` come from `vehicle`. The current tests exercise this mapping but use all-distinct values, so a future source-swap bug (e.g. `nextDestination: vehicle.tripHeadsign`) would only be caught incidentally. Add a test where `vehicle` and `activeTrip` share field names with different values, asserting each field resolves from the correct argument.

2. **Null `activeTrip` guard (code concern).** `buildVehiclePopupData` assumes `activeTrip`, `vehicle`, and `stopsMap` are non-null. At the call sites (`OpenStreetMapProvider.svelte.js` and the Google equivalent), `activeTrip` comes from `activeTripMap.get(...)` and could in principle be `undefined`, in which case `activeTrip.tripHeadsign` throws. Consider adding a guard or documenting the precondition.

Neither is blocking; tracking here so they aren't lost.

VishalRaut2106:
This PR addresses the two follow-up suggestions from the maintainer regarding `buildVehiclePopupData` edge cases.

## Changes
1. **Null Safeguard:** Added optional chaining (`activeTrip?.tripHeadsign`) inside `buildVehiclePopupData` so that if `activeTripMap.get(...)` ever returns `undefined`, the code gracefully returns `undefined` instead of throwing a fatal TypeError.
2. **Asymmetry Test:** Added a new test in `vehicleUtils.test.js` where `vehicle` and `activeTrip` have identically-named properties with different values. This explicitly proves that the helper is pulling data from the correct source object.
3. **Null-State Test:** Added a test that passes `undefined` for `activeTrip` to guarantee the optional chaining correctly prevents crashes.

Closes #508

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented vehicle labels and popup details from failing when trip information is unavailable.
  * Vehicle information now remains visible even when no active trip or destination is present.

* **Tests**
  * Added coverage for missing trip data and overlapping vehicle/trip fields.

### DIFF
--- a/src/lib/Provider/OpenStreetMapProvider.svelte.js
+++ b/src/lib/Provider/OpenStreetMapProvider.svelte.js
@@ -26,11 +26,10 @@ import { t } from 'svelte-i18n';
 import { ROUTE_PANE_Z_INDEX } from '$lib/mapPanes.js';
 
 // activeTrip is always truthy here: the sole caller (vehicleUtils.js) guards on
-// it, and buildVehiclePopupData reads activeTrip.tripHeadsign without optional
-// chaining. Keep this contract consistent rather than implying null is expected.
+// it. The optional chaining here and in buildVehiclePopupData is belt-and-braces.
 function getVehicleLabel(activeTrip) {
 	const translate = get(t);
-	return activeTrip.tripHeadsign
+	return activeTrip?.tripHeadsign
 		? translate('vehicle.to_headsign', { values: { headsign: activeTrip.tripHeadsign } })
 		: translate('vehicle.label');
 }
--- a/src/lib/__tests__/vehicleUtils.test.js
+++ b/src/lib/__tests__/vehicleUtils.test.js
@@ -72,6 +72,50 @@ describe('buildVehiclePopupData', () => {
 			predicted: false
 		});
 	});
+
+	it('extracts fields from the correct sources when vehicle and activeTrip share properties', () => {
+		const vehicle = {
+			vehicleId: 'v-789',
+			lastUpdateTime: 1600000200,
+			predicted: true,
+			tripHeadsign: 'Wrong Destination' // Testing asymmetry
+		};
+		const activeTrip = {
+			tripHeadsign: 'Right Destination',
+			vehicleId: 'wrong-id' // Testing asymmetry
+		};
+		const stopsMap = new Map();
+
+		const result = buildVehiclePopupData(vehicle, activeTrip, stopsMap);
+
+		expect(result).toEqual({
+			nextDestination: 'Right Destination',
+			vehicleId: 'v-789',
+			lastUpdateTime: 1600000200,
+			nextStopName: undefined,
+			predicted: true
+		});
+	});
+
+	it('returns undefined nextDestination safely when activeTrip is undefined', () => {
+		const vehicle = {
+			vehicleId: 'v-999',
+			lastUpdateTime: 1600000300,
+			predicted: false
+		};
+		const activeTrip = undefined;
+		const stopsMap = new Map();
+
+		const result = buildVehiclePopupData(vehicle, activeTrip, stopsMap);
+
+		expect(result).toEqual({
+			nextDestination: undefined,
+			vehicleId: 'v-999',
+			lastUpdateTime: 1600000300,
+			nextStopName: undefined,
+			predicted: false
+		});
+	});
 });
 
 // This suite used to drive the now-deleted single-route `updateVehicleMarkers`
--- a/src/lib/__tests__/vehicleUtils.test.js
+++ b/src/lib/__tests__/vehicleUtils.test.js
@@ -82,7 +82,9 @@ describe('buildVehiclePopupData', () => {
 		};
 		const activeTrip = {
 			tripHeadsign: 'Right Destination',
-			vehicleId: 'wrong-id' // Testing asymmetry
+			vehicleId: 'wrong-id', // Testing asymmetry
+			lastUpdateTime: 999999999,
+			predicted: false
 		};
 		const stopsMap = new Map();
 
--- a/src/lib/__tests__/vehicleUtils.test.js
+++ b/src/lib/__tests__/vehicleUtils.test.js
@@ -48,4 +48,48 @@ describe('buildVehiclePopupData', () => {
 			predicted: false
 		});
 	});
+
+	it('extracts fields from the correct sources when vehicle and activeTrip share properties', () => {
+		const vehicle = {
+			vehicleId: 'v-789',
+			lastUpdateTime: 1600000200,
+			predicted: true,
+			tripHeadsign: 'Wrong Destination' // Testing asymmetry
+		};
+		const activeTrip = {
+			tripHeadsign: 'Right Destination',
+			vehicleId: 'wrong-id' // Testing asymmetry
+		};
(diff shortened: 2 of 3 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Adding optional chaining to `activeTrip?.tripHeadsign` contradicts the explicit contract comment in `OpenStreetMapProvider.svelte.js`, which states that `buildVehiclePopupData` reads `activeTrip.tripHeadsign` "without optional chaining" and asks to "keep this contract consistent rather than implying null is expected". The sibling reader `getVehicleLabel` still dereferences `activeTrip` unguarded and runs first (line 461, before `buildVehiclePopupData` at 478), so the new guard cannot prevent the crash it targets (bug due to src/lib/Provider/OpenStreetMapProvider.svelte.js:28-31)

https://github.com/OneBusAway/wayfinder/blob/0697fd13cd67cbe13185019c3884510d07ab9cf9/src/lib/vehicleUtils.js#L96-L100

aaronbrethorst (changes requested):
The two tests here are genuinely good and are the part of #508 I most wanted. The asymmetry test is constructed correctly — giving `vehicle` and `activeTrip` the same field names with different values is the only way that source-swap regression gets caught deliberately rather than by luck, and putting the wrong value on the *other* object is the right way to write it. The `activeTrip: undefined` test would really throw a `TypeError` without the optional chaining, so it's a true regression guard rather than a passing-by-construction test.

What needs another pass is the guard itself — not because it's wrong, but because it leaves the file contradicting itself.

`OpenStreetMapProvider.svelte.js` currently carries this, three lines above `getVehicleLabel`:

```js
// activeTrip is always truthy here: the sole caller (vehicleUtils.js) guards on
// it, and buildVehiclePopupData reads activeTrip.tripHeadsign without optional
// chaining. Keep this contract consistent rather than implying null is expected.
```

That comment is accurate on `develop` — I confirmed the guard it refers to at `vehicleUtils.js`, where `applyRouteVehicles` does `if (activeTrip && activeTrip.routeId === routeId && ...)` before ever calling `addVehicleMarker` or `updateVehicleMarker`. After this PR, the middle clause is simply false, and the last sentence is arguing against the change sitting one file over.

There's a functional wrinkle underneath the comment, too. In `addVehicleMarker`, `getVehicleLabel(activeTrip)` runs at line 461 and reads `activeTrip.tripHeadsign` unguarded; `buildVehiclePopupData` isn't reached until line 478. Same ordering in the update path (547 before 557). So if the contract ever did break, the crash just moves seventeen lines earlier and the optional chaining never gets a chance to help.

So one of these two, your call:

1. **Keep the guard** — then update that comment block to say the precondition is enforced by the caller and the chaining is belt-and-braces, and give `getVehicleLabel` the same treatment so the file is internally consistent.
2. **Drop the guard** and take the other branch #508 offered — document the precondition on `buildVehiclePopupData` instead. The caller already guarantees it, so this is defensible.

I'd lean toward (1). Either way, keep both tests exactly as they are — under option 2 the `undefined` test becomes a documentation-of-precondition test and should assert the throw instead.

Small fix, and the testing instinct here is right. Send it back and I'll take another look.

VishalRaut2106:
hey @aaronbrethorst   , I have implemented the requested changes  kindly review and let me know if any changes required

aaronbrethorst:
### Code review

Found 1 issue:

1. Character-encoding corruption in `src/lib/__tests__/vehicleUtils.test.js`: seven em dashes (`—`, UTF-8 `E2 80 94`) were rewritten as the mojibake sequence `ΓÇö` (`CE 93 C3 87 C3 B6`) — the CP437 misreading of a UTF-8 em dash, almost certainly from an editor/terminal writing the file back in the wrong codepage. `develop` has all seven as clean em dashes; this branch has all seven corrupted (lines 127, 128, 612, 616, 652, 679, 779). Line 616 is not a comment — it corrupts a live suite name, so the test runner now prints `fetchAndUpdateVehiclesForRoutes ΓÇö live highlight`. Prettier and ESLint will not flag this, so CI passes and it lands silently. The three intentional changes in this PR (the two new tests and the `activeTrip?.` guards) are unrelated to these lines and are correct. (bug due to src/lib/\_\_tests\_\_/vehicleUtils.test.js:616)

https://github.com/OneBusAway/wayfinder/blob/c04edfebd0b731940c1319edb2ba5e74602f7056/src/lib/__tests__/vehicleUtils.test.js#L615-L617

aaronbrethorst (changes requested):
You took option 1 and you took it correctly. The contract comment above `getVehicleLabel` now says the right thing — precondition enforced by the caller, chaining is belt-and-braces — and `getVehicleLabel` got the same guard, which closes the real wrinkle I raised: it runs at line 461 (and 547 on the update path) *before* `buildVehiclePopupData`, so without that change the crash just moved seventeen lines earlier and the new optional chaining never got a chance to help. I re-checked both call paths at your head commit and `activeTrip` isn't dereferenced anywhere else in either one, so the window is genuinely closed. `vehicleUtils.js:64` still guards with `if (activeTrip && ...)`, so the comment's claim is accurate.

Both tests survive intact, and extending the asymmetry test to also conflict on `lastUpdateTime` and `predicted` made it strictly stronger. These are real assertions — revert the `?.` and the `undefined` test throws; swap any source in the implementation and the asymmetry test fails.

One mechanical thing to fix before this lands.

**The branch mangles seven em dashes into mojibake in `vehicleUtils.test.js`.**

Every `—` (UTF-8 `E2 80 94`) on the lines this PR touched came back as `ΓÇö` (`CE 93 C3 87 C3 B6`) — the classic CP437 misreading of a UTF-8 em dash. That's an editor or terminal encoding setting somewhere in your toolchain, not something you typed. I hexdumped the raw file at `c04edfe` to be sure, and confirmed `develop` has seven clean em dashes and zero mojibake while your head has the inverse.

Lines 127, 128, 612, 616, 652, 679, 779. Line 616 is the one that actually bites:

```js
describe('fetchAndUpdateVehiclesForRoutes ΓÇö live highlight', () => {
```

That's a string literal, not a comment, so the corruption shows up in test runner output rather than sitting quietly in the source. Prettier and ESLint don't normalize text inside comments or string literals, which is why CI is green on this.

Restoring those seven lines to their `develop` bytes is the whole fix — nothing else in the diff needs to change. Worth checking your editor's file encoding is set to UTF-8 so it doesn't recur.

The substance here is done and done well. Push the encoding fix and I'll merge it.

tarunsinghofficial:
Hey @VishalRaut2106 , Let me know if you're still working on this PR :)

VishalRaut2106:
> Hey @VishalRaut2106 , Let me know if you're still working on this PR :)

Hey Tarun, I was working on this but I’d like you to take a look and handle fixing it. Once you’re done, please let me know how you solved it — I’d like to understand the approach.

PR #477, opened 2026-05-19
fix(types): align StopPane and StopMarker props with actual SDK data shapes

leocavalhere6:
## Summary

This PR improves type safety in `StopPane.svelte` and `StopMarker.svelte` by replacing incorrect `any` types with accurate OneBusAway SDK types that reflect the actual runtime data structures used by the components.

### StopPane.svelte
- Replaced `any` type for `stop` with:
  - `StopRetrieveResponse.Data.Entry`
- Typed `arrivalsAndDeparturesResponse` using:
  - `ArrivalAndDepartureListResponse | null`

This aligns the component with the actual API response structure and avoids mismatches between runtime data and type definitions.

### StopMarker.svelte
- Replaced `any` type for `stop` with:
  - `StopsForLocationListResponse.Data.List` extended with:
    - `routes: References.Route[] & { code?: string }[]`
- Replaced `any` for `onClick` with a proper function type
- Replaced `any` for `icon` with `IconDefinition` from FontAwesome

## Rationale

Previous types were either too generic (`any`) or incorrectly modeled higher-level SDK response wrappers rather than the actual data shape used in the components.

This change ensures:
- Better type safety
- Improved IDE autocomplete and validation
- Reduced risk of runtime/type mismatch between API and UI logic

## Notes

This is a focused type-only change with no functional or behavioral modifications to the UI.

## Summary by CodeRabbit

* **Refactor**
  * Tightened internal type definitions in stop/map components; no user-facing behavior changes.

* **Chores**
  * Minor locale file formatting adjustment; translations unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/wayfinder/pull/477?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #512, opened 2026-06-08
fix(types): add Stop/MapProvider typedefs and type Tier-1 component props

dsumeet14:
Refs #404 — partial contribution; leaves the issue open for the remaining work.
Adds src/lib/types.js with shared Stop and MapProvider typedefs, and replaces any on the unambiguous props of five components not covered by the existing PRs:

TripDetailsPane
SearchResultItem
MapView
TripPlanModal
TripPlanSearchField

The four example files (StopMarker, StopPane, TabLink, Accordion) are intentionally untouched — already covered by #417 and #477.
AccordionItem.data is left generic on purpose: it's a reusable container holding different payloads at each call site, so narrowing it would be incorrect.
Object-shape props (itineraries, geocode results / onSelect, selectedTrip / selectedRoute) are deferred to a follow-up PR that defines their typedefs.

### DIFF
--- a/src/components/map/MapView.svelte
+++ b/src/components/map/MapView.svelte
@@ -19,8 +19,8 @@
 	 * @property {any} [selectedRoute]
 	 * @property {boolean} [showRoute]
 	 * @property {boolean} [showRouteMap]
-	 * @property {any} [mapProvider]
-	 * @property {any} [stop] - Currently selected stop to preserve visual context
+	 * @property {import('$lib/types').MapProvider | null} [mapProvider]
+	 * @property {import('$lib/types').Stop | null} [stop] - Currently selected stop to preserve visual context
 	 * @property {{ lat: number, lng: number } | null} [initialCoords] - Optional initial coordinates from URL params
 	 */
 
--- a/src/components/oba/TripDetailsPane.svelte
+++ b/src/components/oba/TripDetailsPane.svelte
@@ -7,9 +7,9 @@
 
 	/**
 	 * @typedef {Object} Props
-	 * @property {any} stop
-	 * @property {any} tripId
-	 * @property {any} [serviceDate]
+	 * @property {import('$lib/types').Stop} stop
+	 * @property {string} tripId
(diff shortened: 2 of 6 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is exactly the right shape for this kind of work — narrow, verifiable, and honest about what it isn't doing. Typing only the unambiguous props, leaving `AccordionItem.data` generic because it genuinely is generic, and deferring the object-shape props to a follow-up are all the correct calls. The PR description explaining *why* the four example files are untouched saved me a round trip.

I checked the type references rather than assuming they resolve, and they all do:

- `StopsForLocationListResponse.Data.List` is a real nested interface in `onebusaway-sdk/resources/stops-for-location.d.ts`.
- `References.Route` exists in `resources/shared.d.ts` and indeed has no `code` field — so your note that `code` is non-SDK is accurate, not defensive hand-waving.
- `IconDefinition` is genuinely re-exported from `@fortawesome/fontawesome-svg-core`.
- The `Stop` typedef's claim that `MapView` joins `routeIds` against route references is true — `MapView.svelte:264` does exactly that `routeLookup.get(routeId)` join.
- `(value: string) => void` for `onInput` is right: the field calls `onInput(event.target.value)`, and both `TripPlan` call sites take a `query` string.
- `serviceDate` as `number | null` matches the `arrival.serviceDate` that `StopPane` passes and the `null` default.

Modeling `MapProvider` as a union of the two provider classes is the honest representation given both implement the same interface — better than inventing a hand-written interface that would drift.

One thing you should know, since it isn't your fault and isn't visible from the PR: **no Lint or Test workflow has ever run against this branch's head commit** — only the CLA check, which is why the status looks deceptively quiet. Since I couldn't lean on CI, I ran Prettier against your files with the repo's own `.prettierrc` and they're clean. The changes are confined to JSDoc comment blocks plus a module with no runtime code beyond `export {}`, so there's no path for them to affect tests.

Merging.

PR #513, opened 2026-06-09
fix: fix hover contrast on brand CTA buttons

### LINKED ISSUE #500: a11y: hover state on brand CTAs drops below WCAG AA contrast
Follow-up from #494.

PR #494 correctly fixed the **resting** contrast of the brand CTA buttons by swapping `bg-brand` → `bg-brand-accent` (white text on `#486621` ≈ 6.6:1, passes AA). However, these buttons now hover to `bg-brand` (`#78aa36`), where white text is only ≈ 2.76:1 — below the 4.5:1 AA threshold for normal text.

Affected:
- `src/components/stops/StopPane.svelte` (View Details / View Schedule CTAs, lines ~199, ~205)
- `src/components/stops/StopPageHeader.svelte` (Back to Map button, line ~18)

The resting state is the persistent/dominant state, so #494 is a net improvement and axe default scans pass. But the hover state still technically fails. We lack a brand token darker than `brand-accent`, so a clean fix likely means introducing a darker hover shade (or adjusting hover to darken rather than lighten) so white text stays ≥ 4.5:1 in both states.

tarunsinghofficial:
Fixes #500 

## Summary by CodeRabbit

* **Style**
  * Updated hover states for navigation links and buttons to use a darker accent color for improved visual feedback and consistency.

* **New Features**
  * Added a darker accent color variant to the design system for enhanced visual hierarchy.

### DIFF
--- a/src/components/stops/StopPageHeader.svelte
+++ b/src/components/stops/StopPageHeader.svelte
@@ -15,7 +15,7 @@
 	<div class="mb-4 flex justify-start">
 		<a
 			href="/"
-			class="inline-flex items-center gap-2 rounded-md bg-brand-accent px-4 py-2 text-sm font-medium text-white shadow-sm transition-colors hover:bg-brand focus:outline-none focus:ring-2 focus:ring-brand-accent focus:ring-offset-2"
+			class="inline-flex items-center gap-2 rounded-md bg-brand-accent px-4 py-2 text-sm font-medium text-white shadow-sm transition-colors hover:bg-brand-accent-dark focus:outline-none focus:ring-2 focus:ring-brand-accent focus:ring-offset-2"
 		>
 			<FontAwesomeIcon icon={faArrowLeft} class="h-4 w-4" />
 			<FontAwesomeIcon icon={faMap} class="h-4 w-4" />
--- a/src/components/stops/StopPane.svelte
+++ b/src/components/stops/StopPane.svelte
@@ -196,13 +196,13 @@
 							<div class="mt-auto flex justify-end gap-2">
 								<a
 									href={`/stops/${stop.id}`}
-									class="inline-block rounded-lg border border-brand-accent bg-brand-accent px-3 py-1 text-sm font-medium text-white shadow-md transition duration-200 ease-in-out hover:bg-brand"
+									class="inline-block rounded-lg border border-brand-accent bg-brand-accent px-3 py-1 text-sm font-medium text-white shadow-md transition duration-200 ease-in-out hover:bg-brand-accent-dark"
 								>
 									{$isLoading ? '' : $t('stop_details.view_details')}
 								</a>
 								<a
 									href={`/stops/${stop.id}/schedule`}
-									class="inline-block rounded-lg border border-brand-accent bg-brand-accent px-3 py-1 text-sm font-medium text-white shadow-md transition duration-200 ease-in-out hover:bg-brand"
+									class="inline-block rounded-lg border border-brand-accent bg-brand-accent px-3 py-1 text-sm font-medium text-white shadow-md transition duration-200 ease-in-out hover:bg-brand-accent-dark"
 								>
 									{$isLoading ? '' : $t('schedule_for_stop.view_schedule')}
 								</a>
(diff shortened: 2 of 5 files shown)


aaronbrethorst (approved):
Nice, tightly-scoped fix, Tarun. This does exactly what issue #500 asked for and nothing more: it introduces a `brand-accent-dark` token and points the stop-page CTA hover states at it, so white text stays above WCAG AA in both the resting *and* hover states. I verified the math — the old `hover:bg-brand` (`#78aa36`) gave white text only **2.76:1** (fails AA), while the new `hover:bg-brand-accent-dark` (`#3d571c`) lands at **8.14:1**, actually *higher* than the resting `brand-accent` state's 6.57:1. Hover now darkens instead of lightens, which is the correct direction. Tests pass (57/57) and lint is clean.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None that block this PR. The error-handling observation below is real but is a pre-existing pattern this PR faithfully mirrors, so I'm treating it as a follow-up rather than a blocker (see Suggestions).

## Suggestions (3 found)

1. **Silent `#000000` fallback in `darkenColor` is a copy of an existing pattern, worth hardening project-wide later** [`src/lib/colorUtils.js:124-134`]. `darkenColor` returns `#000000` for falsy/invalid input with no warning — and for a *darken* function, black is indistinguishable from a legitimate extreme result (`darkenColor('#fff', 1.0)` correctly returns `#000000`). The sibling `generatePalette` (same file, fed the same env var) instead `console.warn`s and falls back to a visible brand color. **However**, your `darkenColor` is a precise mirror of the existing `lightenColor` (`colorUtils.js:143-153`), which is equally silent (`#ffffff`). Making only `darkenColor` loud would create a new asymmetry, so this belongs in a follow-up that fixes both helpers — not this PR. Also note: an invalid `COLOR_BRAND_ACCENT` already breaks the resting `brand-accent` color visibly, so the black hover isn't masking an otherwise-silent failure.

2. **No test pins the actual a11y intent (contrast ratio)** [`src/tests/lib/colorUtils.test.js:250-287`]. The suite is genuinely thorough — boundary `amount` values, null/undefined/empty, invalid hex, and a brightness-invariant property test that `lightenColor` lacks. But every assertion is about the arithmetic ("equals `#3d571c`", "is darker"), not the *goal* (≥4.5:1 against white). If someone later changes the `0.15` in `tailwind.config.js` or sets a lighter `COLOR_BRAND_ACCENT`, the hover could silently drop below AA and all tests would still pass. A single test asserting the computed `brand-accent-dark` clears 4.5:1 against `#ffffff` would lock in the intent. Nice-to-have, not required here.

3. **3-digit-hex symmetry gap** [`src/tests/lib/colorUtils.test.js`]. `lightenColor` has a 3-digit-hex case; `darkenColor` doesn't. Covered transitively via `hexToRgb`, so it's a symmetry nit, not a risk.

## Out of Scope (tracked separately)

While reviewing I found several other brand CTAs with the *inverse* of #500: they rest on `bg-brand` (`#78aa36`) with white / `text-brand-foreground` text at **2.76:1** — failing AA in the persistent resting state:
- `src/components/map/ContextMenuPopup.svelte:11,18`
- `src/components/map/PopupContent.svelte:19`
- `src/components/trip-planner/TripPlan.svelte:412`

These are pre-existing, not a regression from this PR, and out of scope here. Tracked in #528. The code-quality follow-ups from Suggestions 1–2 (silent fallbacks, contrast regression test, validate-env 8-digit hex gap) are tracked in #529.

## Strengths

- Surgical scope: solves #500 precisely, touching only what's needed.
- `darkenColor` is a clean, idiomatic mirror of the existing `lightenColor`, reusing `hexToRgb`/`mixColors`/`rgbToHex`.
- `tailwind.config.js:20` follows the adjacent `brand-accent` line exactly (same env var, same default), computing the token at config-eval time like the rest of the palette.
- Test suite mirrors the `lightenColor` contract one-for-one and adds a behavior-based brightness invariant.

## Recommended Action

Merge as-is. The suggestions are follow-up items, not blockers.

PR #518, opened 2026-06-16
feat: Smooth route fit, draw animation, and vehicle movement on the map

Ahmedhossamdev:
Improves the map UX for route display and real-time vehicles.

### Changes

- **Fit to route bounds:** new `fitToPolylines()` on both map providers centers the full route with padding and a max-zoom cap, replacing the hardcoded zoom in `RouteMap` and `SearchPane` (arrival --> trip, and "view all routes" flows).
- **Smooth, glitch-free reveal (OSM):** the route is hidden during the `flyToBounds` glide (avoids the MapLibre GL vs SVG desync), then "draws" itself start --> end via `stroke-dashoffset`.
- **Synced stop markers:** stop markers now appear with the route reveal (after the camera settles) and fade in, instead of popping in first.
- **No flicker on stop focus:** `flyTo` accepts `{ animate: false }`; the route stop list uses it so the route stays glued to the basemap.
- **Animated vehicles:** new `animateMarker` helper interpolates vehicle markers between position updates (~1.2s ease-out) so they glide along the route instead of teleporting; animations are cancelled on marker removal.

## Summary by CodeRabbit

* **New Features**
  * Route stop markers now fade in smoothly.
  * Vehicle markers animate along route geometry.
  * Maps can fit the viewport to the currently displayed route shapes.
* **Improvements**
  * Route camera transitions prefer route-shape fitting, with reliable midpoint fallbacks.
  * Clicking stops keeps the route visually stable by disabling movement animation.
  * Vehicle and route marker animations cancel cleanly during updates/removals.
  * Search interactions ignore outdated loads and prevent stale camera/polling updates.
  * Locale switching now handles both promise and non-promise outcomes consistently.
* **Bug Fixes**
  * More robust polyline decoding and safer route rendering during load.
* **Tests**
  * Expanded animate-marker test coverage for snapping, cancellation, and frame-based motion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2843,9 +2843,9 @@
 			}
 		},
 		"node_modules/caniuse-lite": {
-			"version": "1.0.30001760",
-			"resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001760.tgz",
-			"integrity": "sha512-7AAMPcueWELt1p3mi13HR/LHH0TJLT11cnwDJEs3xA4+CK/PLKeO9Kl1oru24htkyUKtkGCvAx4ohB0Ttry8Dw==",
+			"version": "1.0.30001799",
+			"resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001799.tgz",
+			"integrity": "sha512-hG1bReV+OUU+MOqK4t/ZWI0tZOyz3rqS9XuhOUz1cIcbwBKjOyJEJuw9ER5JuNyqxNk8u/JUVbGibBOL1yrjFw==",
 			"dev": true,
 			"funding": [
 				{
--- a/src/assets/styles/leaflet-map.css
+++ b/src/assets/styles/leaflet-map.css
@@ -31,3 +31,18 @@
 	background: #1f2937;
 	color: #f3f4f6;
 }
+
+/* Fade route stop markers in so they appear gently with the route reveal.
+   Only opacity is animated — Leaflet drives the marker position via transform. */
+.route-stop-marker {
+	animation: route-stop-marker-fade-in 0.3s ease-in;
+}
+
+@keyframes route-stop-marker-fade-in {
+	from {
+		opacity: 0;
+	}
+	to {
+		opacity: 1;
+	}
+}
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -61,11 +61,18 @@
 
 		const stopTimes = tripData?.data?.entry?.schedule?.stopTimes ?? [];
 		const stops = tripData?.data?.references?.stops ?? [];
-		// TODO: implement better way to transition to route shape
-		const location = calculateMidpoint(stops);
 
-		if (location) {
-			mapProvider.flyTo(location.lat, location.lon, 13);
+		// Fit the view to the full route shape so it's always centered and
+		// visible regardless of route length. Fall back to the stops' midpoint
+		// when no polyline could be drawn (e.g. missing shape data). Awaiting the
+		// fit lets the stop markers appear in sync with the route reveal instead
+		// of popping in before the camera has settled.
+		const fitted = await mapProvider.fitToPolylines?.();
+		if (!fitted) {
+			const location = calculateMidpoint(stops);
+			if (location) {
+				mapProvider.flyTo(location.lat, location.lon, 13);
+			}
 		}
 
 		for (const stopTime of stopTimes) {
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -38,52 +38,62 @@
 	});
 
 	async function loadRouteData() {
-		mapProvider.clearAllPolylines();
-		mapProvider.removeStopMarkers();
+		try {
+			mapProvider.clearAllPolylines();
+			mapProvider.removeStopMarkers();
 
-		const tripResponse = await fetch(`/api/oba/trip-details/${tripId}`);
-		tripData = await tripResponse.json();
+			const tripResponse = await fetch(`/api/oba/trip-details/${tripId}`);
+			tripData = await tripResponse.json();
 
-		const tripReferences = tripData?.data?.references?.trips;
-		const moreTripData = tripReferences?.find((t) => t.id == tripId);
+			const tripReferences = tripData?.data?.references?.trips;
+			const moreTripData = tripReferences?.find((t) => t.id == tripId);
 
-		shapeId = moreTripData?.shapeId;
-		const routeId = moreTripData?.routeId;
+			shapeId = moreTripData?.shapeId;
+			const routeId = moreTripData?.routeId;
 
-		if (shapeId && isMounted) {
-			const shapeResponse = await fetch(`/api/oba/shape/${shapeId}`);
-			shapeData = await shapeResponse.json();
-			const shapePoints = shapeData?.data?.entry?.points;
-			if (shapePoints && isMounted) {
-				await mapProvider.createPolyline(shapePoints);
+			if (shapeId && isMounted) {
+				const shapeResponse = await fetch(`/api/oba/shape/${shapeId}`);
+				shapeData = await shapeResponse.json();
+				const shapePoints = shapeData?.data?.entry?.points;
+				if (shapePoints && isMounted) {
+					await mapProvider.createPolyline(shapePoints);
+				}
 			}
-		}
 
-		const stopTimes = tripData?.data?.entry?.schedule?.stopTimes ?? [];
-		const stops = tripData?.data?.references?.stops ?? [];
+			const stopTimes = tripData?.data?.entry?.schedule?.stopTimes ?? [];
+			const stops = tripData?.data?.references?.stops ?? [];
 
-		// Fit the view to the full route shape so it's always centered and
-		// visible regardless of route length. Fall back to the stops' midpoint
-		// when no polyline could be drawn (e.g. missing shape data). Awaiting the
-		// fit lets the stop markers appear in sync with the route reveal instead
-		// of popping in before the camera has settled.
-		const fitted = await mapProvider.fitToPolylines?.();
-		if (!fitted) {
-			const location = calculateMidpoint(stops);
-			if (location) {
-				mapProvider.flyTo(location.lat, location.lon, 13);
+			// Fit the view to the full route shape so it's always centered and
+			// visible regardless of route length. Fall back to the stops' midpoint
+			// when no polyline could be drawn (e.g. missing shape data). Awaiting the
+			// fit lets the stop markers appear in sync with the route reveal instead
+			// of popping in before the camera has settled. Guard the fit on its own
+			// so a camera failure can't stop the stop markers/vehicles from rendering.
+			let fitted = false;
+			try {
+				fitted = await mapProvider.fitToPolylines?.();
+			} catch (error) {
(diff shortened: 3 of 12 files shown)


aaronbrethorst (changes requested):
Hi Ahmed — this is a really enjoyable PR to review. The map has needed exactly this kind of polish for a long time, and you've gone after it thoughtfully: the route now frames itself to its full extent, draws itself in, and vehicles glide along the shape instead of teleporting. The `buildRoutePath` projection math is genuinely nice work, and it's backed by tests that read well and cover the tricky cases (corner-following, reversal, off-route rejection, multi-shape selection). The comments throughout are unusually careful and accurate — I checked the load-bearing ones against the actual library source and they hold up.

There's one issue that has to be fixed before this can merge: the Google `fitToPolylines` implementation can hang the entire route view with no error surfaced. Details below.

## Critical Issues (1 found)

- **`GoogleMapProvider.fitToPolylines()` can hang forever — no timeout fallback.** [`src/lib/Provider/GoogleMapProvider.svelte.js:603-625`]
  The returned Promise resolves *only* from inside the one-shot `idle` listener. Google Maps does not guarantee an `idle` event when `fitBounds` produces no visible viewport change (bounds already fitted, a degenerate single-point bound that survives the `isEmpty()` check, or the map mid-gesture). If `idle` never fires, the Promise never settles.

  Both callers `await` this and gate everything downstream on it:
  - `SearchPane.handleRouteClick` [`src/components/search/SearchPane.svelte:138`] — `showStopsOnRoute` (line 146) and `fetchAndUpdateVehicles` (line 152) never run.
  - `RouteMap.loadRouteData` [`src/components/map/RouteMap.svelte:70`] — the stop-marker loop (78-83) and vehicle polling (86) never run.

  The result is a textbook silent failure: no throw (so the surrounding `try/catch` never fires), no log, no feedback — just a half-drawn route with no stop markers and no vehicles. Your own OSM sibling already guards this exact scenario with a `setTimeout` fallback [`OpenStreetMapProvider.svelte.js:664-666`], which is what makes the omission on the Google side look like an oversight. Mirror it, with a `settled` guard to avoid a double-resolve:

  ```js
  return new Promise((resolve) => {
      let settled = false;
      const finish = () => {
          if (settled) return;
          settled = true;
          if (this.map.getZoom() > maxZoom) this.map.setZoom(maxZoom);
          resolve(true);
      };
      window.google.maps.event.addListenerOnce(this.map, 'idle', finish);
      setTimeout(finish, 1000); // safety net if `idle` never fires
      this.map.fitBounds(bounds, options.padding ?? 50);
  });
  ```

## Important Issues (3 found)

- **Un-cancelled `setTimeout` in `_revealPolylinesWithDraw` fires against polylines removed mid-draw.** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:617-624`]
  The per-polyline timeout that clears inline styles and adds the arrow decorator is never tracked or cancelled. Both callers start a new route load with `clearAllPolylines()` (which runs `polyline.remove()`). If a user clicks a second route within the ~1.2s draw window, the pending callback runs `addDecorator()` against a polyline that was just cleared, leaving a stray arrow decorator with no backing line. No crash, but an intermittent ghost-arrow on rapid route switching that's hard to diagnose. Store the timer id on the polyline and `clearTimeout` it in `removePolyline`/`clearAllPolylines`, and bail out of the callback if `!this.map.hasLayer(polyline)`.

- **OSM `fitToPolylines` reveal closure can draw the *wrong* route after a fast switch.** [`src/lib/Provider/OpenStreetMapProvider.svelte.js:654-666`]
  `reveal()` calls `_revealPolylinesWithDraw`, which iterates `this.polylines` — the live array, not a snapshot. If route B loads before route A's `moveend`/fallback fires, A's pending `reveal` draws B's polylines (or the now-empty set). The `revealed` guard only dedupes A's own two triggers; it doesn't protect against a superseded load. Capture the polyline set (or a load token) when `fitToPolylines` starts and no-op the reveal if it no longer matches.

- **The animation runtime is essentially untested.** [`src/tests/lib/animateMarker.test.js`]
  `buildRoutePath` is well covered, but the module's actual entry point — `animateMarkerTo` — and `cancelMarkerAnimation` have no tests, and neither does the cumulative-distance interpolation (`measurePath`/`positionAlong`) that decides whether the glide looks correct. These are cheap to add with the patterns already in this repo (`vi.useFakeTimers`, `vi.stubGlobal('requestAnimationFrame', …)`). Worth covering before merge:
  1. `cancelMarkerAnimation` — calls `cancelAnimationFrame` with the stored id, nulls `_animationFrameId`, and no-ops on a null marker (the `removeVehicleMarker` path can hit that).
  2. `animateMarkerTo` instant-move fallback — `from === to` and no-`requestAnimationFrame` both call `setPosition(to…)` once and schedule no frame.
  3. Interpolation end-state — the final frame lands exactly on `to`, a midpoint on an L-shaped route sits on the geometry (not the diagonal), and a zero-length path doesn't divide by zero.

  A test asserting `fitToPolylines` resolves even when no camera move occurs would also have caught the Critical issue above. The provider methods that depend on live Leaflet/Google globals (`_revealPolylinesWithDraw`, the bounds math) are fine to leave untested — that's consistent with the rest of the repo.

## Suggestions (3 found)

- **Make the Google `flyTo` options contract explicit.** [`src/lib/Provider/GoogleMapProvider.svelte.js:580-586`] The comment correctly notes the trailing `options` arg is "harmlessly ignored," but relying on silent arg-dropping rots if the OSM signature changes. Consider `flyTo(lat, lng, zoom = 15, _options = {})` so the shared interface is enforced rather than coincidental.
- **Normalize the `createPolyline` null-vs-throw contract.** OSM `createPolyline` returns `null` on decode failure; Google throws. Both are handled safely in the current call paths, but the asymmetry means a future caller can't assume a uniform contract. Worth documenting or normalizing.
- **The 7-line i18n comment is at the verbose end.** [`src/lib/i18n.js:118-127`] Accurate and the nuance is real, but the parenthetical `en-US`-closest-matching example is the load-bearing part; the rest could be trimmed.

## Strengths

- `buildRoutePath` and the projection helpers are sound — equirectangular scaling is appropriate over short vehicle-update spans, segment projection clamps correctly, and forward/backward vertex walking handles the same-segment case.
- Animation cleanup is good defensive design: `cancelMarkerAnimation` runs before each re-animation and on every marker removal/clear path on both providers — no rAF leaks.
- The off-route / null-shape fallbacks degrade gracefully to straight-line interpolation; the marker always moves.
- The i18n `Promise.resolve(locale.set(...))` fix is correct (verified against `svelte-i18n@4.0.1` source) and properly tested.
- The OSM "hide route during the glide, draw it once the basemap settles" approach is a clever, correctly-reasoned fix for the MapLibre GL / SVG desync.

## Recommended Action

1. Fix the Critical Google `fitToPolylines` hang (add the `setTimeout` + `settled` guard).
2. Address the two OSM rapid-switch races (un-cancelled timer, stale reveal closure).
3. Add the three animation-runtime tests.
4. Consider the suggestions.

**Verdict: Request Changes** — the route view can permanently hang on the Google provider with no error surfaced, and that path is reachable in normal use.

Ahmedhossamdev:
> GoogleMapProvider.fitToPolylines() can hang forever — no timeout fallback. [src/lib/Provider/GoogleMapProvider.svelte.js:603-625]
> The returned Promise resolves only from inside the one-shot idle listener. Google Maps does not guarantee an idle event when fitBounds produces no visible viewport change (bounds already fitted, a degenerate single-point bound that survives the isEmpty() check, or the map mid-gesture). If idle never fires, the Promise never settles.


Fair point, I'm working on the fix.

aaronbrethorst (approved):
Ahmed — this is some of the most impressive work I've reviewed in a long time. Smooth, glitch-free map motion is deceptively hard: the route-following vehicle interpolation, the draw-on-reveal animation, and the camera-fit choreography are all the kind of thing that's easy to fake badly and very hard to do *well* — and you did it well. I traced the geometry in `animateMarker.js` by hand (`projectOnSegment`, the forward/backward shape-walking in `buildRoutePath`, `measurePath`, `positionAlong`) and it's correct, including the subtle direction-walking and multi-shape selection cases. The comments explaining the MapLibre-GL/SVG desync are exactly the hard-won rationale future maintainers will thank you for, the teardown paths cancel both animation frames and draw timeouts everywhere they should, and the test coverage on the pure helpers is excellent. Thank you for the care you put into this.

I went ahead and pushed a few follow-up commits to get it over the merge line rather than bouncing it back to you — the core of the feature was already right, and what remained was hardening around the edges that the new `await` points introduced. Here's what changed and why, so nothing is a surprise.

## What changed before merge

### 1. A rapid route-switch race (the important one)

Moving `handleRouteClick` from a synchronous body to one with several `await` suspension points (`createPolyline`, `fitToPolylines`) opened a window where clicking route A then route B in quick succession let A's now-stale handler wake up mid-flight and fight B for the map — flying the camera back to A, re-adding A's stop markers over B's, and overwriting the vehicle-polling interval. Your `_fitToken` already guarded the *draw* inside the OSM provider, but the component-level orchestration (fetch → polylines → camera → markers → polling) had no equivalent.

**Fix:** a `routeLoadToken` in `SearchPane` that's bumped on each click and checked after every await, so a superseded load bails quietly. The vehicle interval is now captured in a local and torn down if it started on a stale load, so it can never clobber the live load's interval. `RouteMap` deliberately *doesn't* get this — it loads once per mount and its existing `isMounted` guard is the right altitude there.

### 2. Unified the `createPolyline` failure contract

The two providers diverged on decode failure — OSM returned `null`, Google threw — and the doc comments described a "Google throws" contract the code didn't actually enforce (it had no guard and would silently build an empty polyline). That asymmetry meant one corrupt shape from the API produced *different* failure modes per provider: a silent hole on OSM, a whole-route abort on Google.

**Fix:** Google's `createPolyline` now guards the decode and returns `null` like OSM does, both doc comments now describe the real behavior (including the async-vs-sync difference that actually matters to callers), and the call sites filter nulls so one bad segment degrades the route instead of breaking the whole load.

### 3. Error handling around the new awaits

`RouteMap.loadRouteData` had no try/catch, so the newly-awaited `fitToPolylines` (or any fetch) could surface as an unhandled rejection and leave the route half-loaded.

**Fix:** wrapped it, with an *isolated* guard around the fit specifically so a camera hiccup can never stop stop-markers and vehicles from rendering, and the failure log now carries the `tripId`.

### Smaller polish

`fitToPolylines` is now uniformly `async` on both providers (no more `boolean | Promise<boolean>`); the Google `idle` listener is cleaned up on the timeout path; OSM `_getRoutePaths` flattens with `flat(1)` (cheaper than the unbounded variant on the per-vehicle hot path) and is hardened against multi-segment shapes; `TripPlanModal` now `await`s `createPolyline` instead of pushing unresolved promises; and I added an off-route fallback test for `animateMarkerTo`.

## Verification

- 1311/1311 tests pass; lint and format clean; production build succeeds.
- Exercised live against the Puget Sound server: route 8 and route 40 both draw, fit-to-bounds, render stops, and show live vehicles, with a clean transition when switching between them and no new console errors.

## Follow-up (not blocking)

Route-load and shape-decode failures are still silent to the *user* — they only log to the console, because the app has no toast/notification primitive yet. This is pre-existing (and these changes actually improved it — decode failures degrade gracefully now), so it's out of scope here, but it's worth a tracking issue: a "couldn't load this route — tap to retry" affordance would close the gap.

## Verdict: **Merge.** 🎉

Really nice work, Ahmed — this meaningfully levels up how the map feels.

PR #522, opened 2026-06-17
fix(a11y): add labels to trip options

### LINKED ISSUE #515: fix(a11y): date, time, and select inputs in trip options have no labels
**What's broken**
The departure time input, departure date input, and walking distance select have no label, aria-label, or aria-labelledby. Screen readers announce them as unlabelled controls and users have no idea what to type or select.

**Fix**
Add aria-label to each:
- "Departure time"
- "Departure date"
- "Maximum walking distance"

WCAG: 1.3.1 Info and Relationships

tarunsinghofficial:
Fixes #515 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility of trip planning form inputs with descriptive labels.
  * Added localization support for trip planner interface.

### DIFF
--- a/src/components/trip-planner/TripOptionsModal.svelte
+++ b/src/components/trip-planner/TripOptionsModal.svelte
@@ -195,12 +195,14 @@
 						<input
 							type="time"
 							bind:value={departureTime}
+							aria-label={$t('trip-planner.departure_time')}
 							class="flex-1 rounded-lg border border-gray-300 bg-white px-3 py-2 text-gray-900 focus:border-blue-500 focus:outline-none focus:ring-1 focus:ring-blue-500 dark:border-gray-600 dark:bg-gray-800 dark:text-white"
 						/>
 						<input
 							type="date"
 							bind:value={departureDate}
 							min={getTodayDateForInput(regionTz)}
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hi Tarun — thanks for this. Bare `type="time"`/`type="date"` inputs and an unlabeled `<select>` are exactly the kind of controls screen reader users stumble over, and `aria-label` is the right tool here since none of these three controls have an associated `<label for>`. The fix is small, correct, and scoped tightly to #515. Ship it.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Notes (not blocking)

- The new value `"departure_date": "Departure date"` is sentence case while its sibling `"departure_time": "Departure Time"` is Title Case. Harmless inconsistency — fine to leave as-is. [src/locales/en.json:38]

## Strengths

- Correct accessibility approach: `aria-label` rather than a visually-redundant `<label>`, appropriate for these inline controls.
- Reused the existing `departure_time` and `max_walking_distance` keys rather than duplicating strings.
- Tight, single-purpose diff that does exactly what #515 asked and nothing more.

---

**Verdict: Merge** — clean, correct, and ready to go.

PR #525, opened 2026-06-25
fix(a11y): wrap autocomplete result buttons in li elements

### LINKED ISSUE #517: fix(a11y): autocomplete results list has invalid markup (buttons inside ul)
**What's broken**
- The search autocomplete renders results as `<button>` elements directly inside a `<ul>`. HTML only allows `<li>` elements as direct children of `<ul>`. Screen readers may not announce the list count and items correctly, or may ignore the list structure entirely.

**Fix**
- Wrap each button in an `<li>`. Also consider adding role="listbox" to the ul and
role="option" to each li for a proper combobox pattern.

WCAG: 1.3.1 Info and Relationships

tarunsinghofficial:
Fixes #517 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the search results dropdown so each result is properly grouped in a list item, making the selection list more consistent and reliable.
* **Style**
  * Updated the results list structure for better semantic HTML.

### DIFF
--- a/src/components/trip-planner/TripPlanSearchField.svelte
+++ b/src/components/trip-planner/TripPlanSearchField.svelte
@@ -67,13 +67,15 @@
 			class="absolute z-10 mt-1 max-h-60 w-full overflow-auto rounded-md border border-gray-300 bg-white shadow-lg"
 		>
 			{#each results as result}
-				<button
-					class="flex w-full cursor-pointer items-center px-4 py-2 text-left hover:bg-gray-100 dark:text-black"
-					onclick={() => handleSelect(result)}
-				>
-					<FontAwesomeIcon icon={faMapMarkerAlt} class="mr-2 text-gray-400  " />
-					{result.displayText}
-				</button>
+				<li>
+					<button
+						class="flex w-full cursor-pointer items-center px-4 py-2 text-left hover:bg-gray-100 dark:text-black"
+						onclick={() => handleSelect(result)}
+					>
+						<FontAwesomeIcon icon={faMapMarkerAlt} class="mr-2 text-gray-400  " />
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Nice, surgical fix, Tarun. This is exactly the kind of change I like to see for an a11y bug: it does one thing, it does it correctly, and it leaves the surrounding markup untouched. Wrapping each result `<button>` in an `<li>` makes the results list valid HTML — `<ul>` only permits `<li>` as direct children — and that's precisely what #517 asked for.

I confirmed locally:
- All 24 tests in `TripPlanSearchField.test.js` pass, including `has proper semantic structure`, which asserts the list role and the button count.
- Prettier reports the file clean.

## Critical Issues (0 found)
None.

## Important Issues (0 found)
None.

## Suggestions (1 found)
- The full WAI-ARIA combobox pattern (`role="combobox"` + `aria-expanded`/`aria-controls`/`aria-activedescendant` on the input, `role="listbox"` on the `<ul>`, `role="option"` on each item, and arrow-key navigation) is still the gold standard for an autocomplete, and #517 flagged it as a "consider." That's intentionally **out of scope** for this PR — it's a separate, larger piece of work and shouldn't block this fix. Tracked as a follow-up in #527. [`src/components/trip-planner/TripPlanSearchField.svelte:66`]

## Strengths
- Minimal, focused diff — no incidental changes, no scope creep.
- Existing test coverage already exercises the list semantics, so the change is verified rather than assumed.
- Indentation and formatting match the surrounding code; lint passes cleanly.

## Verdict

This is ready to merge. The combobox enhancement is a separate follow-up and does not gate this PR. See #527.

PR #531, opened 2026-06-27
fix(a11y): fix recent trips nested buttons, heading level, and Clear All contrast

### LINKED ISSUE #516: fix(a11y): recent trips cards have nested buttons and skipped heading level + "Clear All" button fails contrast
**What's broken**
Two separate but co-located issues:

1. Each recent trip card is a div[role="button"] that contains a real <button> (the delete button). ARIA does not allow interactive descendants inside an interactive role. Screen readers either skip the delete button or announce things twice.

2. The section heading is an h3 with no h2 above it. Heading levels should not skip.
3. The "Clear All" button uses text-gray-400 on a white background. That gives roughly 2.56:1 contrast, well below the 4.5:1 required for normal-size text.

**Fix**
1. Replace div[role="button"] with a real <button> element, or restructure so the delete button sits outside the card button as a sibling.
2. Wrap the recent trips section in an h2 (or visually hidden h2 above the h3).
3. Change text-gray-400 to text-gray-600 (about 4.6:1 on white, passes AA).

WCAG: 4.1.2 Name Role Value, 1.3.1 Info and Relationships and WCAG: 1.4.3 Contrast Minimum for clear All button.

tarunsinghofficial:
Fixes #516 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the recent trips list accessibility and keyboard behavior by using native interactive elements.
  * Updated the “Recent Searches” heading structure for better screen reader support.
  * Refined the clear-all and trip item styling, including dark-mode text and hover states.

### DIFF
--- a/src/components/trip-planner/RecentTripsList.svelte
+++ b/src/components/trip-planner/RecentTripsList.svelte
@@ -19,13 +19,13 @@
 {#if $recentTrips.length > 0}
 	<div class="mt-4">
 		<div class="mb-2 flex items-center justify-between">
-			<h3 class="flex items-center gap-1.5 text-sm font-semibold text-gray-500 dark:text-gray-400">
+			<h2 class="flex items-center gap-1.5 text-sm font-semibold text-gray-500 dark:text-gray-400">
 				<FontAwesomeIcon icon={faClockRotateLeft} class="h-3.5 w-3.5" />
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
nice work, this is ready to merge! I logged a followup task, #541 for you.

PR #532, opened 2026-06-27
fix(a11y): make route schedule table reachable by keyboard with all translations

### LINKED ISSUE #514: fix(a11y): route schedule table unreachable by keyboard
**What's broken**
On `/stops/{id}/schedule`, tabbing past the route accordion skips the entire departure time table, sighted keyboard users can't reach it, and screen readers never enter table navigation mode. Three specific gaps:

1. `<table>` has no `tabindex="0"` : no entry point for keyboard users
2. AM/PM section rows are plain `<td>` : screen readers don't announce them as section headers
3. `<table>` has no caption or `aria-label` : no context when focus lands on it

**What is NOT broken**
Individual minute spans do not need `tabindex`; WCAG 2.1.1 covers functionality, not read-only data. Screen readers navigate table cells natively once the table is reachable.

tarunsinghofficial:
## What this does

Fixes #514 
On /stops/{id}/schedule, tabbing past the route accordion used to skip the entire departure time table. Sighted keyboard users could not reach it and screen readers never entered table navigation mode. This PR closes those gaps.

## Changes

- The scrollable table is now wrapped in a focusable region (role="region", tabindex="0", aria-label). This gives keyboard users an entry point and announces what the table is when focus lands on it. I put the focus on the scroll container rather than the `<table>` element itself, which is the W3C recommended pattern for scrollable tables and also keeps the Svelte a11y linter happy.
- Added an sr-only `<caption>` so the table has a programmatic name for screen readers without changing the visual layout (the route name is already shown in the accordion header).
- The AM and PM rows are now `<th scope="colgroup">` instead of plain `<td>`, so screen readers announce them as section headers for the rows below them.
- Added scope="col" to the Hour and Minutes column headers to complete the header relationships, and removed the cursor-pointer class since those headers are not clickable.
- Added the new schedule_table_caption string to all 25 locale files.

## Screenshots
image

## Testing

- npm run lint passes
- Added src/components/schedule-for-stop/__tests__/RouteScheduleTable.test.js covering the focusable labelled region, the caption, the colgroup and col scopes, and the empty state. All pass.
- Manually verified with VoiceOver that the table is now reachable by keyboard and that headers are announced with each cell.

### DIFF
--- a/src/components/schedule-for-stop/RouteScheduleTable.svelte
+++ b/src/components/schedule-for-stop/RouteScheduleTable.svelte
@@ -4,6 +4,10 @@
 
 	let { schedule } = $props();
 
+	/** Unique per instance so multiple expanded route tables don't share one */
+	/** caption id (aria-labelledby would otherwise resolve to the first table). */
+	const captionId = `schedule-table-caption-${crypto.randomUUID()}`;
+
 	let scheduleData = $derived(renderScheduleTable(schedule));
 
 	function renderScheduleTable(schedule) {
@@ -29,13 +33,13 @@
 <div
 	role="region"
 	tabindex="0"
-	aria-labelledby="schedule-table-caption"
+	aria-labelledby={captionId}
 	class="overflow-x-auto dark:bg-black"
 >
 	<table
 		class="mt-4 w-full table-auto rounded-lg border border-gray-200 shadow-lg dark:border-gray-700 dark:bg-black"
 	>
-		<caption id="schedule-table-caption" class="sr-only">
+		<caption id={captionId} class="sr-only">
 			{$isLoading
 				? ''
 				: $t('schedule_for_stop.schedule_table_caption', {
--- a/src/components/schedule-for-stop/RouteScheduleTable.svelte
+++ b/src/components/schedule-for-stop/RouteScheduleTable.svelte
@@ -23,25 +23,46 @@
 	}
 </script>
 
-<div class="overflow-x-auto dark:bg-black">
+<!-- A focusable scroll container gives keyboard users an entry point to the
+	scrollable schedule table; role="region" + aria-label provide context. -->
+<!-- svelte-ignore a11y_no_noninteractive_tabindex -->
+<div
+	role="region"
+	tabindex="0"
+	aria-label={$isLoading
+		? undefined
+		: $t('schedule_for_stop.schedule_table_caption', {
+				values: { route: schedule.tripHeadsign }
+			})}
+	class="overflow-x-auto dark:bg-black"
+>
 	<table
 		class="mt-4 w-full table-auto rounded-lg border border-gray-200 shadow-lg dark:border-gray-700 dark:bg-black"
 	>
+		<caption class="sr-only">
+			{$isLoading
+				? ''
+				: $t('schedule_for_stop.schedule_table_caption', {
+						values: { route: schedule.tripHeadsign }
+					})}
+		</caption>
 		<thead class="bg-gray-100 text-gray-800 dark:bg-gray-900">
 			<tr>
-				<th class="cursor-pointer px-6 py-3 text-left dark:text-white"
+				<th scope="col" class="px-6 py-3 text-left dark:text-white"
 					>{$isLoading ? '' : $t('schedule_for_stop.hour')}</th
 				>
-				<th class="cursor-pointer px-6 py-3 text-left dark:text-white"
+				<th scope="col" class="px-6 py-3 text-left dark:text-white"
 					>{$isLoading ? '' : $t('schedule_for_stop.minutes')}</th
 				>
 			</tr>
 		</thead>
 		<tbody>
 			<tr class="bg-gray-50 hover:bg-gray-100 dark:hover:bg-gray-600">
-				<td
+				<th
 					colspan="2"
-					class="px-6 py-3 font-semibold text-gray-700 dark:bg-gray-800 dark:text-white">AM</td
+					scope="colgroup"
+					class="px-6 py-3 text-left font-semibold text-gray-700 dark:bg-gray-800 dark:text-white"
+					>AM</th
 				>
 			</tr>
 			{#if scheduleData.amTimes.length === 0}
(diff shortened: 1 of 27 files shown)


aaronbrethorst (changes requested):
Hi Tarun — thanks for this. Keyboard reachability of that schedule table is exactly the kind of quiet, high-impact fix that never shows up in a screenshot but makes the app usable for a whole class of riders. You picked the right primitive (a focusable `overflow-x-auto` region is the W3C-recommended pattern for scrollable tables), your VoiceOver verification shows in the details, and localizing the new string across all 25 locales — with the `{route}` placeholder preserved and genuine translations, not English placeholders — is the standard I wish every PR held to.

A few accessibility details are worth tightening before this lands. Because this is an a11y PR, getting the ARIA semantics exactly right is squarely within its own goalposts, so I'd like these addressed rather than deferred.

## Critical Issues (0 found)

None. No correctness bugs, no null-safety issues (`schedule` is always provided by the `{#each}` in `schedule/+page.svelte:188`), no CLAUDE.md convention violations.

## Important Issues (2 found)

1. **`scope="colgroup"` mislabels the AM/PM section dividers** [`RouteScheduleTable.svelte:63`, `:100`]. These full-width `colspan="2"` headers introduce a group of *rows* (the hourly departures beneath them), not a group of columns — and there's no `<colgroup>` for `colgroup` scope to reference. The semantically correct value is `scope="rowgroup"`. Note that `rowgroup` scope associates the header with the rest of its row group, so a fully correct fix also wraps each section (AM rows, PM rows) in its own `<tbody>`; with both sections in the single `<tbody>` at lines 59–133, a lone `rowgroup` header would scope the entire body. This also propagates into the test: `RouteScheduleTable.test.js:56–66` asserts `scope="colgroup"` and queries `getByRole('columnheader', {name:'AM'})`. A `rowgroup` header exposes the `rowheader` role, so the implementation and the test need to move together. It's a real improvement over the original `<td>` either way — but let's land the correct scope rather than codify the wrong one in a test.

2. **The focusable region loses its accessible name while i18n is loading** [`RouteScheduleTable.svelte:30–36`]. When `$isLoading` is true, `aria-label` evaluates to `undefined` and Svelte drops the attribute, leaving `<div role="region" tabindex="0">` with no accessible name — a keyboard user can tab into an unnamed, non-landmark container. English loads synchronously as the fallback so the window is narrow, but it's reachable while a non-English locale lazy-loads. The clean fix is Suggestion #1 below, which gives the region a stable name and removes the duplicated `$isLoading` branching at the same time.

## Suggestions (3 found)

1. **Consolidate the region name and the `<caption>`** [`RouteScheduleTable.svelte:32–36`, `:42–48`]. Both render the identical interpolated string, so a screen reader announces it once entering the region and again entering the table, and the `$isLoading` ternary is duplicated in two places. Give the `<caption>` an `id` and point the region at it via `aria-labelledby`. That removes the double announcement and resolves Important #2 as a side effect (one source of truth for the name).

2. **Cover the loading state** [`RouteScheduleTable.test.js`]. The `svelte-i18n` mock pins `isLoading` to `false`, so every one of this PR's new `$isLoading ? … : …` branches runs only on its false side. A regression that emitted an empty-string `aria-label` (an empty accessible name, which is itself an a11y bug) rather than omitting it would pass all five current tests. Add a `describe` block that re-mocks `isLoading` to emit `true` (the getter pattern from `LanguageSwitcher.test.js` per CLAUDE.md) and assert the region has *no* `aria-label` attribute and the caption is empty.

3. **Assert the rendered schedule content** [`RouteScheduleTable.test.js`]. The fixture already drives the populated `{#each}` rows, but nothing checks their output — so `convert24HourTo12Hour` (hour `15` → `3` PM), `extractMinutes` (the multi-time cell rendering `05` and `25`), and the `title="Full Time: {hour}:…"` attribute all execute unverified. A couple of assertions against the existing fixture would lock down the core rendering contract cheaply.

## Strengths

- Solves the actual bug from #514 — the table is now a keyboard tab stop — using the recommended scrollable-region pattern.
- Complete, authentic localization across all 25 locales with the `{route}` placeholder intact (verified Arabic RTL, Japanese, Chinese, Spanish, French).
- The accessibility tests are resilient and behavior-focused: `getByRole` with accessible names, asserting roles/scopes rather than CSS classes — exactly right for an a11y change.
- Clean single commit; `prettier`, `eslint`, and the new tests all pass locally. Accurate inline comment explaining *why* the scroll container is focusable.

## Recommended Action

1. Fix the scope semantics (Important #1) — `rowgroup` with per-section `<tbody>`, updating the test alongside.
2. Give the region a loading-safe accessible name (Important #2), ideally via the `aria-labelledby` consolidation in Suggestion #1.
3. Backfill the loading-state and content-rendering tests (Suggestions #2–#3).
4. Re-request review after the fixes.

**Verdict: Request Changes.** The direction is right and the work is close — these are focused a11y-correctness refinements, not a rethink.

aaronbrethorst:
### Code review

Found 1 issue:

1. The caption uses a hardcoded `id="schedule-table-caption"`, and the region references it via `aria-labelledby="schedule-table-caption"`. But `RouteScheduleTable` is rendered once per route inside `{#each schedules as schedule}` in `src/routes/stops/[stopID]/schedule/+page.svelte`, and the custom Accordion allows multiple items open at once (the "Show all routes" button calls `openAll()`, which activates every item). When two or more routes are expanded, the DOM contains duplicate `id="schedule-table-caption"` elements, so every region's `aria-labelledby` resolves to the *first* route's caption. Every schedule table after the first is then announced with the wrong route name, defeating the labelling this PR adds. Deriving a unique id per instance (e.g. `crypto.randomUUID()`, as `AccordionItem.svelte` already does) fixes it. (bug due to duplicate id in RouteScheduleTable.svelte)

https://github.com/OneBusAway/wayfinder/blob/770d639c344e68be0c962fa5b835743900fc82ef/src/components/schedule-for-stop/RouteScheduleTable.svelte#L29-L38

aaronbrethorst (changes requested):
Thanks for tackling this — the underlying accessibility work is genuinely good: `scope="col"`/`scope="rowgroup"` on the header cells, the focusable `role="region"`, the `sr-only` `<caption>`, and rolling the new `schedule_table_caption` key through all 26 locales. That's the right shape for a screen-reader-friendly schedule table, and it's well tested.

One blocker before this can merge:

**The caption id is hardcoded, but this component renders once per route.** `RouteScheduleTable` is instantiated inside an `{#each}` over routes, and the "Show all routes" button expands every accordion item at once. With two or more tables in the DOM you get duplicate `id="schedule-table-caption"`, and every region's `aria-labelledby` resolves to the *first* table's caption — so every table after the first gets the wrong accessible name. That defeats the labelling this PR is adding, precisely in the common multi-route case.

Please make the id unique per instance (e.g. a `crypto.randomUUID()`-derived id, the same pattern `AccordionItem.svelte` already uses) and point each region's `aria-labelledby` at its own caption. The tests assert the literal string `schedule-table-caption`, so they'll need to follow suit.

Happy to re-review as soon as that's sorted.

tarunsinghofficial:

aaronbrethorst (approved):
🔥 👍 :shipit:

PR #535, opened 2026-06-28
fix: handle null/invlaid input in colorUtils.js with tests

### LINKED ISSUE #387: `hexToRgb` and `getBrightness` don't handle null/invalid input
**Issues found:**

- `hexToRgb(hex) `crashes if hex is `null `or `undefined`

If this function gets called with a null or undefined value (which can happen when route/agency color data is missing from the API), it immediately throws because of .replace() being called on a non-string.

**Simple fix** is just adding an early return:
jsif (!hex || typeof hex !== 'string') return null;

- `getBrightness(rgb)` doesn't guard against `null`

This function is called right after hexToRgb, but doesn't check if the rgb object is null before accessing .r, .g, .b. Since hexToRgb can return null on invalid input, this can cascade into another crash.

**Fix:**
if (!rgb) return 0;

tarunsinghofficial:
Fixes #387 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved color handling so invalid or missing color values no longer cause errors.
  * Brightness calculations now safely return a default value when no color data is provided.

* **Tests**
  * Added coverage for invalid and empty inputs to confirm the updated color behavior.

### DIFF
--- a/src/lib/colorUtils.js
+++ b/src/lib/colorUtils.js
@@ -10,6 +10,7 @@
  * @returns {{r: number, g: number, b: number} | null} RGB object or null if invalid
  */
 export function hexToRgb(hex) {
+	if (!hex || typeof hex !== 'string') return null;
 	hex = hex.replace(/^#/, '');
 	// Expand 3-digit hex to 6-digit
 	if (hex.length === 3) {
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
nice work, ready to merge! 👍

Issue #533, opened 2026-06-28
Feat: Add shareable URLs for stops and trip plans


tarunsinghofficial:
**What is the problem**
Right now the browser URL does not reflect what you are looking at. If you open a stop or plan a trip and copy the link, the person you send it to just lands on the default map. Agencies also cannot link directly to a stop page from their own website or printed materials.

**What I want to do**
Reflect the current view in the URL using SvelteKit search params, so the active stop, map position, and trip context are all encoded in the link. Loading that URL should restore the same view. This also enables deep linking.

**Scope**
- Update the URL when a stop is opened and when a trip is planned
- Read params on load to restore the view
- Follow the same pattern already used in PR #42 for search

**Out of scope**
- Any backend changes, this is fully client-side

**Notes**
Part of my GSoC feature work. Happy to share a short before and after once a draft is ready.

tarunsinghofficial:
Closing as the child PR is merged!

PR #538, opened 2026-06-28
feat: add shareable trip plan URLs

tarunsinghofficial:
Part of #533 (stop/map-position work is still separate).

## Summary
Riders can now copy and share a link to a planned trip. When someone opens that link, the app restores the same from/to locations and runs the trip search automatically. For ex- http://localhost:5173/?from=47.44359%2C-122.29607&to=47.47969%2C-122.20792&fromName=SeaTac%2C+WA%2C+USA&toName=Renton%2C+WA%2C+USA.

## What changed

- Added `src/lib/urlState.js` to read and write trip params in the URL (`from`, `to`, `fromName`, `toName`)
- After a successful trip plan, the URL updates with `replaceState` so the link stays shareable without cluttering browser history
- On load, `SearchPane` detects trip params, opens the Plan a Trip tab, and restores the itinerary
- Fixed map behavior for shared links: stop markers no longer stick around when opening a trip URL (race in `batchAddMarkers` + skip initial region-center stop load when a shared trip is present)

## Repro

1. Open the app and go to Plan a Trip
2. Pick a from and to location and press Plan
3. Copy the URL from the address bar (should include `from`, `to`, and optional name params) (For ex: [http://localhost:5173/?from=47.44359%2C-122.29607&to=47.47969%2C-122.20792&fromName=SeaTac%2C+WA%2C+USA&toName=Renton%2C+WA%2C+USA](http://localhost:5173/?from=47.44359%2C-122.29607&to=47.47969%2C-122.20792&fromName=SeaTac%2C+WA%2C+USA&toName=Renton%2C+WA%2C+USA))
4. Open that URL in a new incognito window
5. Confirm the Plan tab opens, the trip runs, and stop markers are hidden on the map
6. Switch to Stops and Stations and confirm stops load for the trip area
7. Clear a from/to field and confirm trip params are removed from the URL

## Summary by CodeRabbit

- **New Features**
  - Added shareable trip links that preserve “from/to” coordinates and place names in the URL.
  - Shared links now restore trip inputs and open trip-planning mode automatically.
- **Bug Fixes**
  - Improved map marker timing and stability when switching between trip-planning and normal modes.
  - Added clearer feedback for invalid shared links and trip-planning request failures.
- **Tests**
  - Expanded automated coverage for URL parsing/sync, shared-link restore flows, and map/trip-planner behavior.

### DIFF
--- a/src/components/MapContainer.svelte
+++ b/src/components/MapContainer.svelte
@@ -15,6 +15,7 @@
 		mapProvider = $bindable(),
 		stop,
 		initialCoords = null,
+		startInTripPlanMode = false,
 		...restProps
 	} = $props();
 
@@ -30,7 +31,14 @@
 </script>
 
 {#if mapProvider}
-	<MapView {handleStopMarkerSelect} {mapProvider} {stop} {initialCoords} {...restProps} />
+	<MapView
+		{handleStopMarkerSelect}
+		{mapProvider}
+		{stop}
+		{initialCoords}
+		{startInTripPlanMode}
+		{...restProps}
+	/>
 {:else}
 	<FullPageLoadingSpinner />
 {/if}
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -27,7 +27,7 @@
 	import { analyticsDistanceToStop } from '$lib/Insights/insightsUtils';
 	import SurveyLauncher from '$components/surveys/SurveyLauncher.svelte';
 	import { parseInitialCoordinates, cleanUrlParams } from '$lib/urlParams';
-	import { parseTripParams } from '$lib/urlState';
+	import { hasTripParams } from '$lib/urlState';
 	import { env } from '$env/dynamic/public';
 	import TripOptionsModal from '$components/trip-planner/TripOptionsModal.svelte';
 	import { showTripOptionsModal } from '$stores/tripOptionsStore';
@@ -48,10 +48,10 @@
 				Number(PUBLIC_OBA_REGION_CENTER_LNG)
 			);
 
-	// Shared trip links should enter trip-plan mode before the map loads stops,
-	// so region-center markers are never painted and then stuck on the map.
+	// Any shared-trip URL (valid or broken) should enter trip-plan mode before the
+	// map loads stops, so region-center markers are never painted and then stuck.
 	const startInTripPlanMode =
-		browser && !!env.PUBLIC_OTP_SERVER_URL && !!parseTripParams(initialPage.url.searchParams);
+		browser && !!env.PUBLIC_OTP_SERVER_URL && hasTripParams(initialPage.url.searchParams);
 
 	let currentModal = $state(null);
 	let selectedTrip = $state(null);
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -27,6 +27,8 @@
 	import { analyticsDistanceToStop } from '$lib/Insights/insightsUtils';
 	import SurveyLauncher from '$components/surveys/SurveyLauncher.svelte';
 	import { parseInitialCoordinates, cleanUrlParams } from '$lib/urlParams';
+	import { parseTripParams } from '$lib/urlState';
+	import { env } from '$env/dynamic/public';
 	import TripOptionsModal from '$components/trip-planner/TripOptionsModal.svelte';
 	import { showTripOptionsModal } from '$stores/tripOptionsStore';
 	import { mapStopPath } from '$lib/mapStopUrl.js';
@@ -45,6 +47,11 @@
 				Number(PUBLIC_OBA_REGION_CENTER_LAT),
 				Number(PUBLIC_OBA_REGION_CENTER_LNG)
 			);
+
+	// Shared trip links should enter trip-plan mode before the map loads stops,
+	// so region-center markers are never painted and then stuck on the map.
+	const startInTripPlanMode =
+		browser && !!env.PUBLIC_OTP_SERVER_URL && !!parseTripParams(initialPage.url.searchParams);
 
 	let currentModal = $state(null);
 	let selectedTrip = $state(null);
@@ -567,6 +574,7 @@
 		{isRouteSelected}
 		{showRouteMap}
 		{initialCoords}
+		{startInTripPlanMode}
 		{activeRoutes}
 		{routeColors}
 		bind:mapProvider
--- a/src/components/map/MapView.svelte
+++ b/src/components/map/MapView.svelte
@@ -24,6 +24,7 @@
 	 * @property {any} [mapProvider]
 	 * @property {any} [stop] - Currently selected stop to preserve visual context
 	 * @property {{ lat: number, lng: number } | null} [initialCoords] - Optional initial coordinates from URL params
+	 * @property {boolean} [startInTripPlanMode] - Skip initial stop load when opening a shared trip link
 	 */
 
 	/** @type {Props} */
@@ -37,7 +38,8 @@
 		stop = null,
 		initialCoords = null,
 		activeRoutes = [],
-		routeColors = new Map()
+		routeColors = new Map(),
+		startInTripPlanMode = false
 	} = $props();
 
 	let routeStopIds = $state(new Map());
@@ -72,7 +74,7 @@
 		}
 	});
 
-	let isTripPlanModeActive = $state(false);
+	let isTripPlanModeActive = $state(startInTripPlanMode);
 	let mapInstance = $state(null);
 	let mapElement = $state();
 	let allStops = $state([]);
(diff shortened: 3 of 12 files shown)


aaronbrethorst (changes requested):
Hi Tarun — this is a genuinely nice feature, and the way you built it is the right way to build it. Pulling the whole URL contract into a dependency-free `urlState.js` and testing it in isolation is exactly the move; it keeps the "what does a shareable link look like" question in one small, well-tested place. The reactive rework in `MapView` is also careful — the `previousMapMode` handling, the `pendingMarkerBatch` cancellation, and the mode re-checks inside the animation frame all hold up under tracing, and your comments explaining *why* (especially the synchronous `allStops` read for effect reactivity) are the kind of thing that saves the next person an hour.

The event choreography across `SearchPane → +page → TripPlan → MapView` is correct: I verified that capturing the parsed trip before dispatching `planTripTabClicked` is load-bearing (that dispatch wipes the query string via `closePane()`'s `pushState('/')`), and `syncTripUrl()` correctly re-writes the params after the itinerary loads.

Where this needs another pass is the **failure paths of the shared-link flow** and **test coverage of the two integration points that make the feature work**. Because a shared link is opened by someone who didn't build it, a silent failure there is worse than in the normal flow — the recipient has no context to recover.

## Critical Issues (0 found)

None. The happy path is correct and the reactive logic is sound.

## Important Issues (3 found)

- **A failed shared link lands the recipient on a blank, dead-end map with no feedback** [`src/components/trip-planner/TripPlan.svelte:222-262`, `src/components/map/MapView.svelte:181`]. When OTP fails at the network/HTTP level (server down, 500, timeout), `fetchTripPlan` returns `null`, so `handleTripPlan` never runs and the trip-planner modal never opens. Normally that's just "nothing happens," but for a shared link `startInTripPlanMode` has *skipped the initial stop load*, so the map is also empty — the recipient sees a blank screen with no itinerary, no error, and no stops to browse. Note the asymmetry: an OTP "no route found" (HTTP 200 + `data.error`) *does* surface in the modal, but a real server failure shows nothing. The underlying swallow in `fetchTripPlan` predates this PR, but this PR introduces the automated caller and the skipped-stop-load that make it a dead end. At minimum, when a shared restore fails to produce an itinerary, surface an error state so the recipient knows the link didn't work.

- **The shared-trip restore path has no error handling and can throw unhandled rejections** [`src/components/trip-planner/TripPlan.svelte:302-311, 207-218`]. `handleLoadSharedTrip` does `await planTrip()` with no `catch`, and `planTrip` has a `try/finally` but no `catch`. Anything that throws outside `fetchTripPlan`'s own try — notably `syncTripUrl()`'s `replaceState`, which throws if called before the SvelteKit router is initialized ("Cannot call replaceState(...) before the router is initialized") — becomes an unhandled promise rejection and aborts the rest of the success path (e.g. the recent-trips save). Wrap the restore in try/catch, and guard `replaceState` in `syncTripUrl`/`clearTripUrl` so that failing to update the *shareable* URL never breaks the *actual* trip the user asked for.

- **No tests for the two integration points that make this feature work** [`src/components/search/SearchPane.svelte:262-280`, `src/components/trip-planner/TripPlan.svelte:302-311, 207-218`]. `urlState.js` is tested well, but the entry point (`maybeRestoreSharedTrip`) and the round trip (`handleLoadSharedTrip` / `syncTripUrl` / `clearTripUrl`) are unexercised on their meaningful paths — a regression in either silently breaks every shared link with no failing test. Both are cheaply testable with harnesses you already have: `SearchPane.test.js` mocks `$env/dynamic/public` and `$page`, and `TripPlan.test.js` already drives the component via `CustomEvent` dispatch and stubs `replaceState` (`test-utils.js:69`). A test that (a) mounts SearchPane with `?from=…&to=…` and asserts `planTripTabClicked` + `loadSharedTrip` are dispatched, and (b) dispatches `loadSharedTrip` into TripPlan and asserts `replaceState` is called with the params (and cleared on input-clear), would cover the gap and match the bar the rest of this PR sets.

## Suggestions (5 found)

- **`Fixes #533` overstates the scope** (PR description). Issue #533 asks for URL sync on stop-open, map position, *and* trip plans; this PR does only trip plans. Merging with `Fixes #533` will auto-close a still-incomplete issue. Consider `Part of #533` and tracking the stop/map-position work separately.

- **Malformed shared links degrade silently** [`src/lib/urlState.js:54-72`]. `parseTripParams` returns `null` for both "no trip params" and "params present but broken" (out-of-range, truncated by a chat client, one endpoint missing), and both callers treat that as "no shared trip." A recipient of a mangled link just gets the normal map. Graceful, but consider distinguishing "params present but invalid" so you can show a "this shared trip link looks broken" hint.

- **Lax coordinate parsing** [`src/lib/urlState.js:39-40`]. `parseFloat('47.6xyz')` returns `47.6`, so `?from=47.6junk,-122.3junk` parses as valid. Consider stricter numeric validation (`Number()` + `Number.isFinite`, or a regex) so corrupted-but-in-range links are rejected rather than silently reinterpreted — and pin the behavior with a test.

- **Provider abstraction leak** [`src/components/map/MapView.svelte:95`]. The return-to-NORMAL branch reads zoom via `mapInstance.map.getZoom()`, reaching past the provider interface into the raw Leaflet/Google object. It follows a pre-existing sibling at line 191, so it's not new debt, but the clean fix is a `getZoom()` method on both providers routing both call sites through it. Fine as a follow-up.

- **Round-trip test masks the rounding asymmetry** [`src/lib/__tests__/urlState.test.js:90-100`], and **brief marker gap on route close** [`src/components/map/MapView.svelte:96-98`]. The round-trip test feeds pre-rounded coords, so it asserts a losslessness that doesn't actually hold (`applyTripParams` rounds to 5 decimals on write); a 6-decimal input would not round-trip equal. Use un-rounded input asserting the rounded expectation. Separately, on *every* ROUTE→NORMAL return the effect now waits for a network round-trip before re-adding markers, so closing a route briefly shows an empty viewport; consider re-adding cached `allStops` synchronously first, then refreshing.

## Strengths

- `urlState.js` is clean, pure, well-documented, range-validated, and refuses to write a partial (`from` without `to`) link. Its test suite covers the primitives thoroughly.
- The `MapView` reactive rework is correct and terminating — no infinite-loop risk, and the `previousMapMode` refresh is what keeps the map from being permanently empty after a shared trip closes.
- Comments are accurate and explain the non-obvious "why" (verified the "planTripTabClicked resets the URL" claim by tracing the full chain). Two small JSDoc tweaks worth making: `MapView.svelte:25` understates `startInTripPlanMode` (it seeds the whole starting mode, not just the stop-load skip), and `SearchPane.svelte:277` should note the `tick()` also lets the Plan tab mount so TripPlan's `loadSharedTrip` listener registers before the event fires.
- Pin teardown is now null-guarded across `clearInput`, modal-close, and `onDestroy`, so overlapping teardown dispatches can't double-remove a marker.

## Recommended Action

1. Give a failed/broken shared link a visible error state instead of a blank dead-end map (Important #1).
2. Add error handling to the restore path and guard `replaceState` (Important #2).
3. Add integration tests for `maybeRestoreSharedTrip` and the `handleLoadSharedTrip`/`syncTripUrl`/`clearTripUrl` round trip (Important #3).
4. Fix up the PR's `Fixes #533` reference and consider the suggestions.
5. Re-request review after the above.

**Verdict: Request Changes.**

tarunsinghofficial:
Thanks @aaronbrethorst for the feedback. Here in this PR, i have addressed all three important items and the suggestions:

- Failed/broken shared links now open the itinerary modal with a visible error instead of a blank map
- Restore path is try/catch guarded; replaceState failures no longer break planning
- Added integration tests for maybeRestoreSharedTrip and the loadSharedTrip/syncTripUrl/clearTripUrl round trip
- Stricter coord parsing, hasTripParams for broken-link detection, getZoom() on both map providers, synchronous cached-stop repaint on mode return

Note: This PR is part of issue #533 (stop/map-position work is still separate) and will be created as a separate PR later.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Really nice piece of work. Pulling the URL serialization into a pure, fully-tested `urlState.js` helper (`formatCoord`/`parseCoord`/`parseTripParams`/`applyTripParams`/…) is the right seam, and the tests cover the cases that matter — round-trips, invalid/out-of-range params, trailing garbage, network failure, malformed payloads, and router-not-initialized.

I checked the parts most likely to bite and they hold up: the `{ error }` return shape from `fetchTripPlan` is consumed correctly in both `handleTripPlan` and `TripPlanModal`, so shared-link failures surface instead of failing silently; the `hasRestoredSharedTrip` re-entrancy guard is set synchronously before the first await; the marker-refresh `$effect` doesn't infinite-loop and re-checks `mapMode` inside the rAF; and `startInTripPlanMode` is browser-gated so there's no hydration mismatch. Svelte 5 runes, path aliases, and en.json strings all in order. Ready to merge.

aaronbrethorst:
Approved on the merits — the review above stands. Heads up though: now that some of the sibling PRs have landed on `develop` (notably #551, which also touched `GoogleMapProvider`/`OpenStreetMapProvider`, plus other changes to `SearchPane.svelte`), this branch has developed merge conflicts and GitHub now marks it `DIRTY`. Could you merge/rebase `develop` in and resolve the conflicts? Once it's mergeable again it's good to go — no further review needed on my end unless the conflict resolution changes behavior.

tarunsinghofficial:

aaronbrethorst (approved):
nice work! 👍

PR #539, opened 2026-06-28
Fix/load spinners

### LINKED ISSUE #526: a11y: mark decorative spinner SVG in TripPlan as aria-hidden
Follow-up from #509 / #510.

`src/components/trip-planner/TripPlan.svelte` (around line 416) renders a decorative spinner `<svg>` inside the "plan trip" button while `loading` is true. The button already exposes the visible localized text `{$t('trip-planner.planning')}...` alongside it, so the SVG is purely decorative and contributes nothing useful to assistive tech.

For consistency with the spinners fixed in #510, add `aria-hidden="true"` to that `<svg>`.

### Acceptance criteria
- [ ] `<svg>` in `src/components/trip-planner/TripPlan.svelte` (planTrip button) has `aria-hidden="true"`
- [ ] `npm run lint` passes

Low priority / fit-and-finish.

tarunsinghofficial:
The plan trip button already announces its loading state via visible text, so the spinner SVG should be ignored by assistive tech.

Fixes #526

### DIFF
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -438,6 +438,7 @@
 					xmlns="http://www.w3.org/2000/svg"
 					fill="none"
 					viewBox="0 0 24 24"
+					aria-hidden="true"
 				>
 					<circle class="opacity-25" cx="12" cy="12" r="10" stroke="currentColor" stroke-width="4"
 					></circle>


aaronbrethorst (commented):
nice work! this is a great addition.

aaronbrethorst (approved):
nice work, great addition :)

PR #547, opened 2026-07-04
fix(map): highlight selected vehicle and correct trip-details stop

Ahmedhossamdev:
Fixes: #520 
Fixes: #521 

Two fixes for the route/trip view:

- Trip details highlighted the wrong stop for the bus. Position was
  guessed from raw lat/lon ranges, which assumes stops are ordered
  monotonically by coordinate — wrong on most route directions. Now
  resolves the stop by the server's closestStop / nextStop id.
- Selecting an arrival showed every vehicle on the route with no way to
  tell which was yours. The clicked trip's vehicle now gets a soft glow
  and raised z-index; other route vehicles stay visible.

Tests: added regression coverage for the stop resolution, the highlight
flag threading, and the glow markup. Full suite green (1396 tests).

image

image

## Summary by CodeRabbit

* **New Features**
  * Selected trips are now visually highlighted on the map, making it easier to follow a specific vehicle.
  * Vehicle markers now render with a distinct emphasis and stay on top of other markers when highlighted.
  * Trip details now use a more reliable stop-position lookup, improving the bus position shown for a trip.

* **Tests**
  * Added coverage for highlighted vehicle markers and stop-position resolution to prevent regressions.

PR #549, opened 2026-07-05
fix(map): key vehicle markers by vehicleId to stop markers jumping

### LINKED ISSUE #548: Vehicle marker jumps back and forth on some routes
### Description

On some routes, a vehicle marker repeatedly jumps forward and then back to a
different location every refresh (~30s), instead of moving smoothly.

### Cause

The OBA `trips-for-route` response can list the **same `activeTripId` for two
different vehicles** — the vehicle actually serving the trip, plus a second one
parked at the base (which sits on a shared "depot" coordinate). We key vehicle
markers by `activeTripId`, so both entries collapse into a single marker that
gets set to one vehicle's position, then the other's, on every poll.

### Evidence

Same `activeTripId`, two different vehicles and positions in one response:

    veh=1_8094700  trip=1_727429170  @47.52330,-122.26822   (base)
    veh=1_4517     trip=1_727429170  @47.52210,-122.26445   (real, moving)

Both map to one marker lead to ->> it flips between the two spots each refresh.

### Expected

Each physical vehicle should have its own marker and move smoothly; two vehicles
sharing a trip id must not fight over one marker.

Ahmedhossamdev:
Fixes: #548 

Vehicle markers were keyed by `activeTripId`, but the OBA trips-for-route
response can report the same `activeTripId` for two different vehicles (the one
serving the trip plus a second parked at the base). Both collapsed into one
marker that flipped between their positions on every ~30s refresh — the vehicle
appeared to jump forward and back.

Key markers by the physical `vehicleId` instead (falling back to `activeTripId`
only when a status has no vehicle id), so each vehicle gets its own marker and
two vehicles sharing a trip no longer collide.

Added regression tests: two vehicles sharing an activeTripId now produce two
markers (not one add + one update), a vehicle reuses its marker across
refreshes, and the no-vehicleId fallback still works.

It looks great now after testing vs android app!

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vehicle marker behavior so markers stay associated with the correct vehicle, even when multiple vehicles share the same trip.
  * Prevented markers from being duplicated or reassigned incorrectly during refreshes.
  * Ensured markers are still handled correctly when vehicle details are incomplete.

### DIFF
--- a/src/lib/__tests__/vehicleUtils.test.js
+++ b/src/lib/__tests__/vehicleUtils.test.js
@@ -54,7 +54,7 @@ describe('buildVehiclePopupData', () => {
 	});
 });
 
-describe('updateVehicleMarkers highlighting', () => {
+describe('updateVehicleMarkers', () => {
 	function makeProvider() {
 		return {
 			addVehicleMarker: vi.fn(() => ({})),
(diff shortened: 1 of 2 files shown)


Ahmedhossamdev:
@coderabbitai review

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits — this is a clean, well-targeted fix. Keying the marker map by `vehicleId || activeTripId` is exactly right (the two namespaces can't collide, so the fallback is safe), it's applied consistently across the add/update/remove and cleanup paths, and the three regression tests genuinely exercise the split, reuse, and no-vehicleId cases.

Heads up: now that sibling #547 landed on `develop` and touched the same lines of `src/lib/vehicleUtils.js` (and `vehicleUtils.test.js`), this branch has merge conflicts. Please merge/rebase `develop` in and resolve. The two changes are compatible — keep your `markerKey` as the map key, and keep #547's `isHighlighted = highlightedTripId === activeTripId` alongside it (highlight is still driven by trip id, keying is by vehicle id). Once it's mergeable it's good to go — no re-review needed unless the resolution changes behavior.

Ahmedhossamdev:
Done @aaronbrethorst

PR #551, opened 2026-07-05
fix(a11y): polish marker accessibility follow-ups from #487

### LINKED ISSUE #542: Follow-ups from PR #487: marker a11y polish & test coverage
Minor, non-blocking follow-ups identified during review of #487 (marker keyboard accessibility). None of these blocked that PR; capturing them so they aren't lost.

### 1. Cover the GoogleMapProvider `title: stop.name` change with a test
`src/lib/Provider/GoogleMapProvider.svelte.js` now sets `title: stop.name` on stop route markers — this is the **only** accessible-name mechanism for stops on Google-backed regions (no `aria-label` path there), yet it has no test. A one-line assertion mirroring the OSM `addMarker` option test would guard against a silent regression that strips the accessible name for every stop.

### 2. Guard `aria-label` against a falsy `stop.name`
`OpenStreetMapProvider.svelte.js` (`addStopRouteMarker`) calls `el.setAttribute('aria-label', stop.name)`. `setAttribute` coerces `undefined`/`null` to the literal string `"undefined"` — worse than no label for a screen reader. OBA stops always carry a `name` today, so this is purely defensive, but `stop.name ?? ''` (or an `if (el && stop.name)` guard) makes it robust.

### 3. Make the `getElement() === null` path observable
Both `addStopRouteMarker` and `addVehicleMarker` do `const el = marker.getElement(); if (el) { …a11y… }`. If `getElement()` ever returns null (marker added mid-teardown / before render), the aria-label and keyboard handler are silently skipped, shipping an inaccessible marker with no diagnostic. A `console.warn` in the `else` would make this failure audible. Add a test exercising the null branch (the current fake marker always returns a live element, so this path is never hit).

### 4. (Pre-existing, not from #487) `removeVehicleMarker` doesn't prune `this.vehicleMarkers`
`removeVehicleMarker` calls `marker.remove()` but never removes the entry from the `vehicleMarkers` array, so removed markers (and now their detached DOM + keydown closures) are retained until `clearVehicleMarkers()`. Pre-dates #487; worth a cleanup pass.

_Filed from code review of #487._

tarunsinghofficial:
- Add GoogleMapProvider test that stop route markers set title to stop.name
- Guard OSM aria-label against falsy stop names via setupMarkerAccessibility
- Warn when Leaflet getElement() returns null during a11y setup
- Prune vehicleMarkers when removeVehicleMarker runs (OSM and Google)

Fixes #542

### DIFF
--- a/src/lib/Provider/GoogleMapProvider.svelte.js
+++ b/src/lib/Provider/GoogleMapProvider.svelte.js
@@ -406,8 +406,15 @@ export default class GoogleMapProvider {
 	}
 
 	removeVehicleMarker(marker) {
+		if (!marker) return;
+
 		cancelMarkerAnimation(marker);
 		marker.setMap(null);
+
+		const index = this.vehicleMarkers.indexOf(marker);
+		if (index > -1) {
+			this.vehicleMarkers.splice(index, 1);
+		}
 	}
 
 	clearVehicleMarkers() {
--- a/src/lib/Provider/OpenStreetMapProvider.svelte.js
+++ b/src/lib/Provider/OpenStreetMapProvider.svelte.js
@@ -223,6 +223,31 @@ export default class OpenStreetMapProvider {
 		});
 	}
 
+	/**
+	 * Apply aria-label and keyboard activation to a Leaflet marker element.
+	 * Logs a warning when getElement() returns null so inaccessible markers are
+	 * not silently shipped during teardown or before first render.
+	 *
+	 * @param {import('leaflet').Marker} marker
+	 * @param {{ label?: string, onActivate?: () => void, context: string }} options
+	 */
+	setupMarkerAccessibility(marker, { label, onActivate, context }) {
+		const el = marker.getElement();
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Clean, tightly-scoped follow-up that implements all four items from #542: the `removeVehicleMarker` null guard plus pruning the entry out of `vehicleMarkers` (fixing the leak where it only called `setMap(null)`), the falsy-name guard on `aria-label`, and the `console.warn` diagnostic when `getElement()` returns null so inaccessible markers aren't silently shipped. The `setupMarkerAccessibility` extraction reads well and the load-bearing `preventDefault` comment is preserved.

I confirmed the array pruning is safe — the production caller iterates the separate `vehicleMarkersMap`, not the `vehicleMarkers` array, so the in-place `splice` can't corrupt an active iteration. Tests cover the new branches. Ready to merge.

PR #552, opened 2026-07-05
Fix test coverage follow-ups for RecentTripsList a11y refactor

### LINKED ISSUE #541: Test coverage follow-ups for RecentTripsList a11y refactor (#531)
Follow-up from PR #531, which refactored the recent-trips cards to native buttons and fixed heading/contrast a11y issues. The code is correct, but a few worthwhile test-coverage enhancements were deferred to keep that PR small:

- **Assert delete does not trigger `onSelect`.** The delete button is now an absolutely-positioned sibling of the card button (no longer nested), and `handleDelete` still calls `e.stopPropagation()`. The existing test asserts `removeTrip` is called but never asserts `onSelect` was *not* — add `expect(mockOnSelect).not.toHaveBeenCalled()` so a future re-nesting can't silently regress into "remove navigates the user into a trip plan."
- **Add a keyboard-activation test.** The whole point of moving from `div[role=button]`+`onkeydown` to a native `<button>` is free keyboard semantics. A test that focuses the card and fires `{Enter}`/`Space` asserting `onSelect` fires would guard the exact behavior the refactor claims.
- **Loosen the contrast assertion.** `RecentTripsList.test.js` pins `toHaveClass('text-gray-600')`; the `.not.toHaveClass('text-gray-400')` half carries the real regression value. Consider keeping the negative assertion and adding a comment naming gray-600 as the AA threshold, so an equal-or-better change (e.g. gray-700) doesn't fail a correct edit.

Separately (pre-existing, not introduced by #531): the delete button uses `opacity-0 group-hover:opacity-100`, which leaves it clickable and keyboard-focusable while visually hidden. Worth gating behind `focus-within` / `pointer-events` in a future a11y pass.

tarunsinghofficial:
Fixes #541

### DIFF
--- a/src/components/trip-planner/__tests__/RecentTripsList.test.js
+++ b/src/components/trip-planner/__tests__/RecentTripsList.test.js
@@ -13,6 +13,7 @@ vi.mock('svelte-i18n', () => {
 	const translations = {
 		'trip-planner.remove_recent_trip': 'Remove recent trip',
 		'trip-planner.recent_searches': 'Recent Searches',
+		'trip-planner.recent_trip': 'Recent trip from {from} to {to}',
 		'trip-planner.clear_all': 'Clear All'
 	};
 	return {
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Test-only follow-up that pulls its weight: the keyboard-activation test genuinely exercises Enter/Space on the native button, and the delete test now asserts `onSelect` isn't called. The comment documenting why the "Clear All" color assertion was loosened (pin the regression shade, not the exact shade) is the right call and clearly explained. All tests pass. Ready to merge.

PR #553, opened 2026-07-05
test(a11y): strengthen skip-to-content layout test coverage

### LINKED ISSUE #502: Strengthen skip-to-content a11y test coverage (i18n + dark-mode focus)
Follow-up to #497, which internationalized the skip-to-main-content link and fixed its focus target and dark-mode focus styling.

The unit tests in `src/routes/__tests__/layout.test.js` cover the structural a11y contract well (link present, targets `#main-content`, link is not `tabindex=-1`, `<main>` is `tabindex=-1`). Two coverage improvements were deferred from that PR:

1. **i18n assertion is a near-tautology.** The test asserting `textContent === 'skip_to_main_content'` only passes because the global `svelte-i18n` mock returns the key verbatim. It guards against a hardcoded-English regression but does not verify the key resolves or exists. Consider asserting against real locale data (load `src/locales/en.json`, assert `skip_to_main_content` is present and non-empty) — or, more simply, assert the text is *not* the hardcoded `"Skip to main content"`.

2. **Dark-mode focus styling and actual focus movement are untested.** The `dark:focus:*` classes and the behavior of focus landing on `<main>` after activation can't be meaningfully verified under jsdom (Tailwind variants don't compute; fragment navigation isn't implemented). These belong in a Playwright/e2e a11y test rather than a brittle className assertion.

Low priority — the PR ships correct behavior; this is about hardening regression coverage.

tarunsinghofficial:
- Assert skip_to_main_content exists in en.json locale data
- Mock $t() with a distinct resolved label to prove the layout wires the key
- Document that dark-mode focus and focus movement belong in Playwright (#502)

Fixes #502

## Summary by CodeRabbit

* **Tests**
  * Improved layout coverage for the skip-to-main-content link.
  * Updated the localization mock to use the real translated label instead of the raw key.
  * Added checks to ensure the translated skip label exists and is not empty.
  * Clarified that focus behavior for the skip link should be verified in a browser-based test.

### DIFF
--- a/src/routes/__tests__/layout.test.js
+++ b/src/routes/__tests__/layout.test.js
@@ -1,6 +1,35 @@
 import { render } from '@testing-library/svelte';
 import { describe, expect, test, vi, beforeEach } from 'vitest';
 import Layout from '$src/routes/+layout.svelte';
+import en from '../../locales/en.json';
+
+// Distinct from en.skip_to_main_content so the render test proves $t() is called
+// rather than a hardcoded English literal that happens to match the locale file.
+const RESOLVED_SKIP_LABEL = 'Resolved skip-to-main label';
+
+vi.mock('svelte-i18n', () => ({
+	t: {
+		subscribe: vi.fn((fn) => {
+			fn((key) => {
+				if (key === 'skip_to_main_content') {
+					return RESOLVED_SKIP_LABEL;
+				}
+				return en[key] ?? key;
+			});
+			return { unsubscribe: () => {} };
+		})
+	},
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice strengthening of the skip-link coverage. Splitting into "key exists in en locale data" plus "text is resolved via \$t()" — with a distinct sentinel so the test fails if the component ever hardcodes the string or drops \$t() — is exactly the right way to lock this in. The file-local `svelte-i18n` mock correctly overrides the global one. Ready to merge.

PR #555, opened 2026-07-05
feat: surface route load and polyline decode failures via toast notifications

### LINKED ISSUE #530: Route load / polyline decode failures are silent to the user
### Summary

Surfaced during review of #518. When a route fails to load or a route shape fails to decode, the only artifact is a `console.error` — the user sees a near-empty map (or a route drawn with a missing segment) with no indication anything went wrong and no retry affordance.

This is **pre-existing** behavior, not introduced by #518; that PR's follow-up fixes actually improved the situation (decode failures now degrade gracefully via `createPolyline` returning `null` instead of aborting the whole load, and `RouteMap.loadRouteData` no longer throws unhandled rejections). But the underlying gap remains.

### Affected paths

- `src/components/map/RouteMap.svelte` — `loadRouteData` catch logs to console only.
- `src/components/search/SearchPane.svelte` — `handleRouteClick` outer catch logs to console only.
- `src/lib/Provider/*MapProvider.svelte.js` — `createPolyline` logs decode failures to console and returns `null`; a partial route renders as if correct.
- `src/components/trip-planner/TripPlanModal.svelte` — a leg whose geometry fails to decode is silently skipped.

### Why it's deferred

The codebase currently has no user-facing error primitive (no toast/snackbar/Sentry). `console.error` is the established house style. Addressing this well means introducing a lightweight notification primitive, which was out of scope for the #518 review fixes.

### Suggested direction

Introduce a small toast/notification primitive and surface a "Couldn't load this route — tap to retry" affordance on total load failure, and (optionally) a subtler indicator when a route renders with a missing shape segment.

tarunsinghofficial:
Fixes #530 

Introduces a lightweight toast notification primitive and surfaces route load/polyline decode failures that were previously console-only.

- Total failures: red toast with Retry (SearchPane route click, RouteMap trip overlay)
- Partial decode failures: amber warning when some shape segments fail to draw
  (SearchPane, RouteMap, TripPlanModal)

Screenshot:
Screenshot 2026-07-05 at 9 33 46 PM

## Summary by CodeRabbit

* **New Features**
  * Added global toast notifications for route loading and rendering issues, including tap-to-retry (when available) and dismiss controls.
  * Introduced route notification helpers and new localized notification text.
* **Bug Fixes**
  * Route and trip-detail loading now surface failures to users (with retry) instead of only logging.
  * Users are warned when only part of the route shape/geometry can be drawn.
  * Notifications are cleared on teardown to prevent stale messages across navigation/retries.
* **Tests**
  * Added coverage for the notification store and route notification helpers, including timing behavior.

### DIFF
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -28,6 +28,7 @@
 	import SurveyLauncher from '$components/surveys/SurveyLauncher.svelte';
 	import { parseInitialCoordinates, cleanUrlParams } from '$lib/urlParams';
 	import { hasTripParams } from '$lib/urlState';
+	import { notifications } from '$stores/notificationStore';
 	import { env } from '$env/dynamic/public';
 	import TripOptionsModal from '$components/trip-planner/TripOptionsModal.svelte';
 	import { showTripOptionsModal } from '$stores/tripOptionsStore';
@@ -179,6 +180,10 @@
 				tripPlanError = null;
 			}
 			currentModal = null;
+			// The stop now owns the map, so any route/trip toast still on screen is
+			// stale — and tapping its Retry would tear down the stop's own markers
+			// and polylines. Unscoped: this supersedes whichever component raised it.
+			notifications.dismiss();
 
 			searchCollapsed = true;
 			if (browser && window.innerWidth >= 768) sheetSnap = 'full';
--- a/src/components/map/RouteMap.svelte
+++ b/src/components/map/RouteMap.svelte
@@ -3,6 +3,7 @@
 	import { clearVehicleMarkersMap, fetchAndUpdateVehicles } from '$lib/vehicleUtils';
 	import { mapContrastColor } from '$lib/colorUtils';
 	import { onMount, onDestroy } from 'svelte';
+	import { notifyRouteLoadFailed, notifyPartialRouteShape } from '$lib/routeNotifications';
 	let { mapProvider, tripId, currentSelectedStop = null } = $props();
 	let shapeId = null;
 	let tripData = null;
@@ -59,6 +60,9 @@
 			mapProvider.removeStopMarkers();
 
 			const tripResponse = await fetch(`/api/oba/trip-details/${tripId}`);
+			if (!tripResponse.ok) {
+				throw new Error(`Trip details request failed: ${tripResponse.status}`);
+			}
 			tripData = await tripResponse.json();
 
 			const tripReferences = tripData?.data?.references?.trips;
@@ -76,7 +80,10 @@
 				shapeData = await shapeResponse.json();
 				const shapePoints = shapeData?.data?.entry?.points;
 				if (shapePoints && isMounted) {
-					await mapProvider.createPolyline(shapePoints, { color: routeColor });
+					const polyline = await mapProvider.createPolyline(shapePoints, { color: routeColor });
+					if (!polyline) {
+						notifyPartialRouteShape();
+					}
 				}
 			}
 
@@ -122,6 +129,9 @@
 			}
 		} catch (error) {
 			console.error(`Error loading route data for trip ${tripId}:`, error);
+			if (isMounted) {
+				notifyRouteLoadFailed(() => loadRouteData());
+			}
 		}
 	}
 </script>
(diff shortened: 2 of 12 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
The core of this is good and genuinely useful: failures are surfaced rather than swallowed, every catch keeps its `console.error` and adds a toast, and the partial-shape detection leans on the providers' documented `null`-on-decode-failure contract so it won't fire spurious warnings. The store and helper are well tested.

One thing I want fixed before merge, and it's a real bug rather than a nitpick:

**`RouteMap` never dismisses its notification on unmount, and its Retry closure outlives the component.** Retriable error toasts intentionally have no auto-dismiss (`keeps retriable errors visible until dismissed`), and the error toast is a global singleton. So when a route load fails and the user navigates away, the toast lingers on top of whatever view they moved to. Worse, the retry callback is `() => loadRouteData()` bound to the destroyed `RouteMap` instance — tapping Retry then runs `loadRouteData()`, which starts with an unconditional `mapProvider.removeStopMarkers()` (and clears polylines), wiping the map content the current view owns, while the inner draw steps are gated on `isMounted === false` and redraw nothing. Net result: a stale toast can blank the shared map.

The fix is small — dismiss the notification in `RouteMap`'s `onDestroy` (you already import it), so the toast and its stale closure don't outlive the component. Please also give the two `SearchPane`/`TripPlanModal` retry paths a quick look for the same lifecycle concern.

Everything else looks ready; happy to re-review right after.

tarunsinghofficial:

aaronbrethorst (approved):
I rebased this onto current `develop` and pushed a follow-up commit (`7adf0f2`) fixing what a full review turned up. Two rebase conflicts resolved by hand: the import block in `SearchPane.svelte`, and the `createPolyline` call in `RouteMap.svelte`, which now carries both develop's `{ color: routeColor }` and your null check.

The primitive is the right shape and it's wired end-to-end — `<Toast />` really is mounted in `+layout.svelte`, strings really do go through `t()`. Nice. What I fixed falls into three buckets:

**The store had no ownership model.** A single global slot, an argument-less `dismiss()`, and a `show()` that overwrites unconditionally. Three separate bugs fell out of that: a transient `RouteMap` mount (the `tripId`-less one your own comment describes) wiped `SearchPane`'s retry toast on unmount; a 6-second partial-shape warning from the trip planner permanently destroyed a persistent route-load error and its retry callback; and a route-failure toast survived a stop selection, so tapping Retry blanked the map underneath an open stop sheet. `show()` now returns an id, `dismiss(id)` only clears a match, and a retriable error outranks a plain one. Stop selection in `MapExperience` dismisses unscoped, since it genuinely supersedes whatever's showing.

**Two of the four paths #530 names were still silent.** In `TripPlanModal`, `legCount++` sits below `if (!shape) continue`, so a leg with no geometry at all is excluded from the denominator — the one case the issue calls out for that file never fires a warning. And `RouteMap` checks `tripResponse.ok` but not `shapeResponse.ok`, so a 500 from `/api/oba/shape/:id` yields `shapePoints === undefined`, no throw, no toast, and a route-less map. Related: `RouteMap` draws *one* polyline for the whole trip, so a null return there means total failure, not a missing segment — it was showing the auto-dismissing "part of this route" warning and forfeiting the retry. That's now a distinct `route_shape_failed` error with retry, and stops/vehicles still render.

**Two lifecycle issues.** The retry callback called `loadRouteData()` without assigning `loadRouteDataPromise`, which silently broke the invariant that `onDestroy` awaits every in-flight load — an orphaned retry could fly the camera off a stop the user had just opened. And `isMounted` was checked *before* the awaited `createPolyline`, not after, so a slow Google decode could raise a toast for an already-unmounted view with nothing left to dismiss it.

Also: the error toast never auto-dismissed and is `fixed bottom-6 z-[9999] pointer-events-auto`, so it permanently covered the bottom sheet's controls. It now expires after 20s, Escape dismisses it, and the wrapper is `pointer-events-none` so only the card intercepts taps. Added a `Toast.svelte` component test plus store coverage for scoped dismiss, priority, and expiry — the render path had none, which is exactly the failure mode #530 is about.

PR #556, opened 2026-07-05
fix(a11y): use brand-accent for resting CTA contrast

### LINKED ISSUE #528: a11y: remaining brand CTAs fail WCAG AA contrast at rest (bg-brand)
Follow-up from #500 / #513.

PR #513 fixed the hover-state contrast on the stop-page CTAs by introducing a `brand-accent-dark` hover token. While reviewing it, I found several other brand CTAs that have the *inverse* problem: they rest on `bg-brand` (`#78aa36`) with white / `text-brand-foreground` text, which is only **2.76:1** — below the 4.5:1 AA threshold for normal text. The resting state is the persistent state, so this is more visible than a hover failure.

Affected:
- `src/components/map/ContextMenuPopup.svelte:11,18` (`text-brand-foreground` on `bg-brand`)
- `src/components/map/PopupContent.svelte:19` (`text-white` on `bg-brand`)
- `src/components/trip-planner/TripPlan.svelte:412` (`text-brand-foreground` on `bg-brand`)

These all hover *to* `bg-brand-accent` (which passes), so the fix is likely the same shape as #494/#500: move the resting state to `bg-brand-accent` and hover to the `brand-accent-dark` token introduced in #513.

Not a regression from #513 — pre-existing, just surfaced during that review.

tarunsinghofficial:
Fixes #528

Several brand CTAs rested on `bg-brand` (#78aa36) with white text, which is only 2.76:1 contrast (below WCAG 2.1 AA 4.5:1 for normal text). PR #513 fixed hover on stop-page CTAs; this applies the same resting-state pattern everywhere else that had the inverse problem.

Resting state is now `bg-brand-accent` + `text-white`; hover is `hover:bg-brand-accent-dark` (token from #513).

## Changes

- **ContextMenuPopup** - Start here / End here buttons
- **PopupContent** - View stop information button on route stop popups
- **TripPlan** - Plan your trip button (and loading spinner color)
- **AlertsModal** - More info button dark mode (was still `dark:bg-brand`)
- **app.css** - Active itinerary tab and `.button--primary` hover aligned to the same pattern

## Tests
All tests passed. And I have manually verified all contrast checks for the buttons using the AxeTools extension (premium version).

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -56,9 +56,9 @@
 	}
 
 	.itinerary-tab--active {
-		@apply bg-brand text-brand-foreground shadow-md;
-		@apply hover:bg-brand;
-		@apply dark:bg-brand dark:text-brand-foreground dark:hover:bg-brand;
+		@apply bg-brand-accent text-white shadow-md;
+		@apply hover:bg-brand-accent-dark;
+		@apply dark:bg-brand-accent dark:text-white dark:hover:bg-brand-accent-dark;
 	}
 
 	.close-button {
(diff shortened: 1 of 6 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Solid contrast fix for #528. `brand-accent` (#486621) on white computes to ~6.6:1, clearing AA, so this genuinely resolves the ~2.8:1 problem rather than moving it around. Good catch on the AlertsModal dark-mode variant that was still `dark:bg-brand`, `brand-accent-dark` is a real token so the hover states resolve, and the test assertions match the new class lists. Ready to merge.

PR #557, opened 2026-07-09
test: harden error.test.js cleanup and assertions

### LINKED ISSUE #498: Harden error.test.js mock cleanup against assertion-failure leaks
Follow-up from PR #489.

`src/routes/__tests__/error.test.js` restores its spies and the `Object.defineProperty(window, 'location', ...)` stub *inline* at the end of each test body, and the file has no top-level `afterEach`. If an assertion throws before the inline restore runs, `window.location` / `window.history.back` / `window.history.length` stay mocked for every subsequent test in the run, producing confusing cascade failures.

The sibling `LanguageSwitcher/__tests__/LanguageSwitcher.test.js` already guards against this with `afterEach(() => vi.restoreAllMocks())`. Since error.test.js is intentionally modeled on that file's conventions, it should match.

Suggested fixes:
1. Add `afterEach(() => { vi.restoreAllMocks(); })` to the top-level `ErrorPage` describe.
2. Wrap the `window.location` redefine/restore in `try/finally` (or capture `originalLocation` once and restore in `afterEach`).
3. Minor parity: the error-page "$_ throws" test could spy on and assert the `[i18n fallback]` `console.warn`, matching the LanguageSwitcher "throws" test — otherwise removing the warn line in `+error.svelte` would not fail any error-page test.

Low priority — these only manifest on an already-failing assertion and are test-hardening, not correctness fixes.

antriksh-9:
### Description
- automatic mock cleanup (afterEach)
- safe `window.location` restoration (try/finally)
- additional `console.warn` assertion
- verified using `npm run test -- src/routes/__tests__/error.test.js`

Closes #498 

Screenshot from 2026-07-09 23-20-40

## Summary by CodeRabbit

* **Tests**
  * Improved mock lifecycle management to reset spies between test cases
  * Strengthened error handling and fallback behavior validation

### DIFF
--- a/src/routes/__tests__/error.test.js
+++ b/src/routes/__tests__/error.test.js
@@ -1,5 +1,5 @@
 import { render, screen, waitFor, fireEvent } from '@testing-library/svelte';
-import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi, afterEach } from 'vitest';
 import ErrorPage from '$src/routes/+error.svelte';
 import { locale, __setTranslatorMode } from 'svelte-i18n';
 import { page } from '$app/stores';
@@ -128,6 +128,10 @@ describe('ErrorPage', () => {
 		page.__setError({ message: 'Not Found' });
 	});
 
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
 	test('renders the 404 status code', () => {
 		render(ErrorPage);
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a clean piece of test hardening — thanks.

The part I like most is that the new `console.warn` assertion is load-bearing rather than decorative. `+error.svelte:57` warns with `` `[i18n fallback] Error translating ${key}:` `` and `e.message`, and your `expect.stringContaining(...)` / `'Translator error'` pair matches that exact two-argument shape, so deleting the warn line would actually fail the test. That was the real gap.

The `try/finally` around the `window.location` stub is the right call too — `vi.restoreAllMocks()` can't undo an `Object.defineProperty` stub, so that one genuinely needed manual restoration on the failure path.

One thing worth knowing for later, not a blocker: `vi.restoreAllMocks()` in `afterEach` also resets the `vi.fn()` mocks registered in `vitest-setup.js` (`ResizeObserver`, `IntersectionObserver`, `matchMedia`), degrading them to no-ops for the rest of the file. Harmless here since `+error.svelte` touches none of them, and it matches the `LanguageSwitcher` precedent — but it's a trap if this file ever renders something that observes resize.

Merging.

PR #564, opened 2026-07-22
Refine load-more empty-state copy and add test coverage

### LINKED ISSUE #561: Load more arrivals: refine empty-state copy and add load-more test coverage
Follow-up from #546 (Load more arrivals). Two minor, non-blocking items surfaced in review:

**1. Empty-state copy reads oddly below a populated list.** When "Load more arrivals" widens the window but the server returns no additional arrivals, `noMoreArrivals` becomes true and the list renders:

> No arrivals found in the next 65 minutes

directly *above* a populated list of arrivals. The wording is confusing in that position — it should read as "No *more* arrivals in the next N minutes" (or similar) when there are already results shown, and keep the current phrasing only for the truly-empty case. See the `loadMoreButton` snippet in `src/components/stops/StopPane.svelte`.

**2. Test coverage for the load-more path.** The current tests cover the new message text and the `?minutesAfter=35` URL, but not the button click, `loadMoreArrivals()`, or the `noMoreArrivals` transition. Worth adding a click-through test so the pagination behavior is guarded.

Neither is a functional bug — the feature works — so this is polish, not a fix.

Ahmedhossamdev:
Fixes #561 

- Add no_more_arrivals_in_next_minutes i18n key for when results exist but load-more returns nothing new
- Use differentiating copy: 'No arrivals found' for truly empty stops, 'No more arrivals' when widening the window yields nothing
- Add 4 tests covering load-more click, API call with incremented minutesAfter, noMoreArrivals message, and truly empty stop message

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved arrival loading so the app accurately detects whether additional arrivals are available after expanding the time window.
  - Prevented outdated “no more arrivals” messages from persisting when later arrivals are found.
  - Distinguishes between stops with no arrivals and stops with no further arrivals.

- **Localization**
  - Added translated “No more arrivals” messaging across supported languages.

- **Tests**
  - Added coverage for expanded arrival searches, empty results, and accurate status messages.

### DIFF
--- a/src/components/stops/StopPane.svelte
+++ b/src/components/stops/StopPane.svelte
@@ -53,7 +53,28 @@
 	let currentStopSurvey = $state(null);
 	let remainingSurveyQuestions = $state([]);
 
+	// Furthest arrival time (ms) observed when we last concluded "no more arrivals".
+	// A later poll that surfaces an arrival beyond this clears the stale hint.
+	let noMoreArrivalsBoundary = 0;
+
 	let abortController = null;
+
+	/**
+	 * Returns the furthest-out arrival time (ms) in the list, preferring the
+	 * predicted time and falling back to the scheduled time. Comparing this
+	 * across window widenings is more robust than a raw count: an arrival
+	 * departing the front of the list no longer masks a genuinely later one.
+	 * @param {any[]} [arrivals]
+	 * @returns {number}
+	 */
+	function furthestArrivalTime(arrivals) {
+		let furthest = 0;
+		for (const arrival of arrivals ?? []) {
+			const time = arrival.predictedArrivalTime || arrival.scheduledArrivalTime || 0;
+			if (time > furthest) furthest = time;
+		}
+		return furthest;
+	}
 	/**
 	 * Fetches arrivals for the stop within the current `minutesAfter` window.
 	 * @returns {Promise<number|null>} the number of arrivals fetched, or `null`
@@ -86,9 +107,13 @@
 			serviceAlerts = filterActiveAlerts(data.data.references.situations || []);
 			error = null; // Clear previous errors if successful
 			const count = arrivalsAndDepartures?.arrivalsAndDepartures?.length ?? 0;
-			// A refresh that returns arrivals clears a stale "no more arrivals" hint
-			// (e.g. after a 30s poll surfaces new departures).
-			if (count > 0) {
+			// Clear a stale "no more arrivals" hint only when a refresh actually
+			// surfaces an arrival further out than the window we'd exhausted — a
+			// poll returning the same (or nearer) arrivals must leave the hint up.
+			if (
+				noMoreArrivals &&
+				furthestArrivalTime(arrivalsAndDepartures?.arrivalsAndDepartures) > noMoreArrivalsBoundary
+			) {
 				noMoreArrivals = false;
 			}
 			return count;
@@ -119,7 +144,7 @@
 	// Widen the arrivals window and refetch. If the count doesn't grow, surface a
 	// "no more arrivals found" hint (the server has nothing further to show).
 	async function loadMoreArrivals() {
-		const previousCount = arrivalsAndDepartures?.arrivalsAndDepartures?.length ?? 0;
+		const previousFurthest = furthestArrivalTime(arrivalsAndDepartures?.arrivalsAndDepartures);
 
 		loadingMore = true;
 		noMoreArrivals = false;
@@ -132,7 +157,10 @@
 		loadingMore = false;
 		// Only conclude "no more arrivals" from a request that actually completed.
 		if (count !== null) {
-			noMoreArrivals = count === previousCount;
+			const newFurthest = furthestArrivalTime(arrivalsAndDepartures?.arrivalsAndDepartures);
+			// Widening the window surfaced nothing further out → nothing more to show.
+			noMoreArrivals = newFurthest <= previousFurthest;
+			if (noMoreArrivals) noMoreArrivalsBoundary = newFurthest;
 		}
 		analytics.reportArrivalClicked('Loaded more arrivals');
 	}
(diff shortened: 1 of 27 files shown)


src/components/stops/StopPane.svelte
```diff
@@ -293,10 +293,14 @@
 
 				{#snippet loadMoreButton(emptyResults = false)}
 					<div class="flex flex-col items-center gap-2">
-						{#if emptyResults || noMoreArrivals}
+						{#if emptyResults}
```
aaronbrethorst, line 324: **Correctness — the new `No more arrivals` copy is unreachable in the empty-stop path.**

Calling `loadMoreButton(true)` forces `emptyResults=true`, so this `{#if emptyResults}` branch always wins and the `{:else if noMoreArrivals}` copy can never render on a stop that started empty. A user at a zero-arrival stop only ever sees *"No arrivals found in the next N minutes"*, even after clicking Load more and widening the window with still nothing further out. If that's intended, great — but the two messages are mutually exclusive by construction here, so if anyone expected "No more arrivals" to appear after exhausting an initially-empty stop, it never will.

src/components/stops/__tests__/StopPane.test.js
```diff
@@ -507,6 +509,93 @@ describe('StopPane', () => {
...
+			expect(global.fetch).toHaveBeenCalledWith(
+				'/api/oba/arrivals-and-departures-for-stop/1_75403?minutesAfter=65',
+				expect.objectContaining({
+					signal: expect.any(AbortSignal)
+				})
+			);
+		});
+	});
+
+	test('shows "No more arrivals" message when load-more returns same count', async () => {
```
aaronbrethorst: **Cleanup — this test duplicates `shows "No more arrivals" message when load-more returns no new results` (line 512).**

Both tests render, click *Load more arrivals* once with the mock returning the same `mockArrivalsAndDeparturesResponse`, and assert *"No more arrivals in the next 65 minutes"*. The only difference is `mockResolvedValueOnce` vs `mockResolvedValue`, which exercises the same `count === previousCount → noMoreArrivals` path. It adds ~20 lines to maintain without covering new behavior — if the no-more logic changes, both tests must be updated in lockstep for no added confidence. Consider dropping this one (or, if the intent was to cover the *repeated-poll* case, make it actually poll twice so it tests something distinct).

aaronbrethorst:
🤖 **Code review — 2 additional findings on pre-existing `loadData` logic**

These land on lines outside this PR's diff (so they can't be inline comments), but they directly affect whether the new *"No more arrivals"* message behaves as intended:

**1. Correctness — the 30s background poll silently clears the new message (`StopPane.svelte` ~line 92).**
After Load more sets `noMoreArrivals = true` and the new copy renders, the 30s refresh interval left running by `resetDataFetchInterval()` fires, `loadData()` refetches the same arrivals, and `if (count > 0) noMoreArrivals = false` flips it back off. The distinctive message the PR adds is therefore transient — it usually vanishes ~30s after appearing, before users can read it.

**2. Correctness — inferring exhaustion from count equality gives false positives (`StopPane.svelte` ~line 135).**
`noMoreArrivals` is derived purely from `count === previousCount`. At a busy stop, if a bus departs the front of the list (3 → 2 in view) just as the user clicks Load more, the widened window can pull in one further-out arrival, landing back at count 3 == previousCount. The code then shows *"No more arrivals in the next 65 minutes"* even though more arrivals exist further out. Comparing on a stable identity (e.g. the set of trip IDs, or the furthest arrival time) would be more robust than a raw count.

_Findings from a high-effort multi-agent review; verified before posting._

aaronbrethorst (changes requested):
two findings to fix.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new `shows "No arrivals found" for truly empty stop` test is a byte-for-byte duplicate of the pre-existing `displays empty arrivals message when no arrivals` test (same `mockEmptyArrivalsAndDeparturesResponse` mock, same `render`, same `getByText('No arrivals found in the next 35 minutes')` assertion, only the test name differs). It guards nothing the existing test doesn't already guard — this is the same duplication that was flagged earlier in this PR. To make it earn its place, have it assert the *differentiation* the PR is actually about: click Load more on the empty stop and assert `No more arrivals in the next 65 minutes` is **absent** while `No arrivals found in the next 65 minutes` is present. Otherwise, drop it. (duplicate of `src/components/stops/__tests__/StopPane.test.js#L245-L258`)

https://github.com/OneBusAway/wayfinder/blob/1c422a7f94466b05802e2a8c726c3e1fbb2849d0/src/components/stops/__tests__/StopPane.test.js#L625-L639

aaronbrethorst (approved):
Both of the things I blocked on are properly fixed, so I'm clearing my change request and merging this.

Specifically, I re-checked both at the current head:

- **The 30s poll no longer wipes the hint.** `loadData` now clears `noMoreArrivals` only when a refresh surfaces an arrival further out than `noMoreArrivalsBoundary`, instead of the old unconditional `if (count > 0)`. And you set the boundary in the same block that sets `noMoreArrivals = true`, so it can't be read stale.
- **The count-equality false positive is gone.** `noMoreArrivals = newFurthest <= previousFurthest` is the right mechanism, and the new test at `StopPane.test.js:540` exercises exactly the case I described — front arrival departs, a later one slides in, count unchanged. I checked the fixture arithmetic and it would genuinely fail under the old logic.

Locale coverage is complete at 25/25, and CI is green.

Two follow-up nits, neither worth another round trip — I'll pick them up separately unless you'd rather:

1. `StopPane.test.js:626` — `shows "No arrivals found" for truly empty stop` is the same test as `displays empty arrivals message when no arrivals` at line 245: same mock, same render, same assertion. Same duplication class I flagged before; it just moved.
2. The comment above `loadMoreArrivals` still says *"If the count doesn't grow, surface a 'no more arrivals found' hint"*. The logic underneath no longer uses count, and the hint is now the distinct new string, so that comment now misdescribes the function it sits on.

On my earlier note about the "No more arrivals" copy being unreachable in the empty-stop path — that's the behavior issue #561 asked for, so consider it answered. No change needed.

Nice work on the iteration here.

PR #575, opened 2026-07-26
feat: Fit map smoothly to trip itinerary after planning

### LINKED ISSUE #537: Feat: Fit map smoothly to trip itinerary after planning
**What is the problem**

When I plan a trip, the itinerary polylines show up on the map but the camera does not move to frame them. The map often stays on the from/to pins or the default region view. Google Maps and similar apps pan and zoom smoothly so the full route is visible as soon as results load.

**What I want**

After a successful trip plan (and when switching between itinerary tabs), the map should smoothly fit the visible route polylines, similar to how route selection already works elsewhere in the app.

**Current behavior**

- TripPlanModal draws polylines for each leg in drawRoute()
- The map is not fitted to those polylines
- Route selection already uses fitToPolylines on the map providers, so we likely have the building blocks

**Expected behavior**

- User plans a trip and sees results
- Map animates (flyTo / fitBounds) to show the full active itinerary
- Switching itinerary tabs re-fits the map to that option's route
- Works on both OSM and Google map providers

Scope

- Call fitToPolylines (or equivalent) after polylines are drawn in TripPlanModal
- Re-fit when the active itinerary tab changes
- Keep animation smooth, not a hard jump

**Notes**
This will be a Follow up from shareable trip URL work issue #533.

tarunsinghofficial:
## Summary
- After a successful trip plan, the map now smoothly fits to the active itinerary’s polylines (same building block route selection already uses).
- Switching itinerary tabs re-fits the camera to that option’s route.
- Padding accounts for the mobile bottom sheet and the desktop side panel so the route lands in the visible map, not under the UI. Works on both OSM and Google.

## What changed
- **`TripPlanModal`:** after drawing an itinerary’s leg polylines, call `fitToPolylines` so the map frames the route. Same path runs on new results and on tab switches. A draw token cleans up orphan polylines if you switch tabs mid-draw; if fit fails (e.g. missing geometry), fall back to `flyTo` on the itinerary midpoint.
- **`mapFitPadding`:** new helper that turns the sheet/panel’s bounding rect into per-side padding so the route lands in the visible map (above the mobile bottom sheet, beside the desktop side column) instead of under the UI. Caps any side at 60% of the viewport.
- **`BottomSheet`:** expose a bindable `element` so the modal can measure the sheet without poking at test ids.
- **Map providers:** OSM `fitToPolylines` now accepts `{ top, right, bottom, left }` padding (via `toLeafletPadding`) in addition to the existing number / `[x, y]` forms. Google already forwarded object padding to `fitBounds` — JSDoc only. Route selection callers are unchanged.
- **Tests:** unit coverage for padding math, provider padding translation, and TripPlanModal fit / re-fit / stale-draw / fallback behavior.

## Screenshot
ScreenRecording2026-07-26at1 30 11PM-ezgif com-video-to-gif-converter

Closes #537

## Summary by CodeRabbit

- **New Features**
  - Added map-fit padding computation (`panelFitPadding`) for bottom sheets and side panels.
  - Exposed a bindable `BottomSheet` root element reference to support accurate viewport padding.
  - Added padding normalization for OpenStreetMap map fitting (`toLeafletPadding`).

- **Bug Fixes**
  - Improved itinerary route framing using per-panel padding.
  - Prevented stale/overlapping route polylines and race-condition issues during rapid itinerary switching.
  - Added safer fallback behavior when automatic fitting can’t be performed, including partial-geometry warnings.

- **Tests**
  - Expanded map fitting, padding normalization, itinerary switching, stale draw cleanup, and warning/toast lifecycle coverage.

### DIFF
--- a/src/components/navigation/BottomSheet.svelte
+++ b/src/components/navigation/BottomSheet.svelte
@@ -7,6 +7,7 @@
     the extremes.
 
     @prop {('peek'|'half'|'full')} snap - Bindable current snap point
+    @prop {HTMLElement | null} element - Bindable reference to the sheet DOM node (for measuring occlusion)
     @prop {import('svelte').Snippet} header - Condensed header rendered inside the drag-handle row
     @prop {import('svelte').Snippet} children - Scrollable sheet body
 -->
@@ -15,7 +16,7 @@
 	import '$lib/i18n.js';
 	import { t } from 'svelte-i18n';
 
-	let { snap = $bindable('half'), header, children } = $props();
+	let { snap = $bindable('half'), element = $bindable(null), header, children } = $props();
 
 	const SNAP_ORDER = ['peek', 'half', 'full'];
 	const PEEK_HEIGHT = 150;
@@ -88,6 +89,7 @@
 
 <div class="pointer-events-none absolute inset-0" bind:clientHeight={containerHeight}>
 	<div
+		bind:this={element}
 		class="pointer-events-auto absolute inset-x-0 bottom-0 flex flex-col overflow-hidden rounded-t-[14px] border border-b-0 border-gray-400 bg-surface/95 shadow-[0_-8px_24px_rgba(0,0,0,.18)] backdrop-blur dark:border-gray-600 dark:bg-surface-dark/95 dark:text-surface-foreground-dark"
 		style:height="{sheetHeight}px"
 		style:transition={dragging ? 'none' : 'height .28s cubic-bezier(0,0,.2,1)'}
--- a/src/components/trip-planner/TripPlanModal.svelte
+++ b/src/components/trip-planner/TripPlanModal.svelte
@@ -103,7 +103,15 @@
 			const style = getLegPolylineStyle(leg);
 			// Await: the Google provider's createPolyline is async and returns
 			// null on decode failure — skip nulls instead of tracking a Promise.
-			const polyline = await mapProvider.createPolyline(shape, style);
+			// A provider that throws instead of returning null must not reject
+			// this fire-and-forget draw: that would skip the fit *and* the
+			// midpoint fallback, stranding the camera with no route and no toast.
+			let polyline = null;
+			try {
+				polyline = await mapProvider.createPolyline(shape, style);
+			} catch (error) {
+				console.error('Error creating itinerary leg polyline:', error);
+			}
 			// A newer draw (tab switch / new results) took over while we were
 			// awaiting — remove the orphan so it can't strand on the map, then bail.
 			if (token !== drawToken) {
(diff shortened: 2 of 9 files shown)


aaronbrethorst (approved):
Approving. Nice feature — framing the itinerary after planning is a real improvement over leaving the camera wherever the user left it, and the provider-parity work (`fitToPolylines` on both OSM and Google, with `toLeafletPadding` normalizing the shared `{top,right,bottom,left}` shape) is the right way to do it. Verified against the Maps JS API reference that `fitBounds` takes `number|Padding` natively, so passing the object straight through on the Google side is correct.

I rebased this onto `develop` and resolved a conflict in `TripPlanModal.svelte`: the partial-route-shape toasts from #555 had landed in the same `drawRoute`/`onDestroy` code this branch rewrote. The two changes are complementary, so I merged them rather than picking a side — `drawnCount`/`legCount` accounting now increments off the local `drawn` array so the stale-draw token check still works, and `onDestroy` keeps both `drawToken++` and the scoped `notifications.dismiss()`.

I also pushed four fixes on top:

1. **The fit fallback could crash.** An itinerary with an empty `legs` array clears the `!legs` guard (an empty array is truthy), so `legs[0].from` threw a `TypeError` that escaped as an unhandled rejection.
2. **…and could pan the map to NaN.** A leg endpoint without `lat`/`lon` made `calculateMidpoint` average `undefined` into `NaN` and return a truthy `{lat: NaN, lon: NaN}`, so `flyTo(NaN, NaN, 13)` sailed past the `if (midpoint)` guard. Both are fixed by filtering endpoints to finite coordinates first.
3. **A throwing `createPolyline` stranded the camera** (thanks @coderabbitai for spotting the unguarded `await`). Because `drawRoute` is fired and forgotten from an `$effect`, a provider that throws instead of returning `null` skipped the fit *and* the fallback — no route, no toast, camera untouched. Now caught per leg and logged, so the leg counts as undrawn and the existing partial-shape warning covers it.
4. Dropped the vestigial `await` in `onDestroy` and added the missing `panelFitPadding` top-branch test.

Every fix has a regression test, and I mutation-tested each one — reverting the fix makes the corresponding test fail, so none of them are vacuous. That caught one of my own tests being a no-op: an unhandled rejection surfaces on the process, not as a jsdom `window` event, so the first version of the empty-legs test passed against the broken code.

Local: 1704 tests green, `svelte-check` clean, lint clean.

One thing I left alone: the `isSidePanel` branch in `mapFitPadding.js` is unreachable from its only caller, since `BottomSheet`'s root is `inset-x-0` and always spans the full map container. Harmless as generic helper behavior, and not worth narrowing the utility's contract in this PR — but worth knowing it's untriggered today if you build on it.

tarunsinghofficial:
Thanks for the detailed review @aaronbrethorst and pushing the code changes as part of review :)

PR #576, opened 2026-07-26
Feat: Filter service alerts by stop/route and add severity styling

### LINKED ISSUE #534: Feat: Filter service alerts by affected route and stop, add severity styling
**What is the problem**
The service alerts panel shows every agency wide alert with no way to see only what affects the stop or routes you actually care about. There is also no visual difference between a minor notice and a severe disruption.

**What I want to do**
- Filter alerts using the allAffects data already in the OBA response (routeId and stopId), so riders see relevant alerts first while agency wide alerts still show
- Add severity differentiation for INFO, WARNING, and SEVERE
- Surface active date ranges and cause or effect fields that the API already returns

**Scope**
- Add a filter helper that reads allAffects
- Update ServiceAlerts.svelte and ServiceAlertItem.svelte for filtering and severity
- Keep the existing pagination

**Out of scope**
- No new API endpoints, the data is already in the response

**Notes**
Part of my GSoC feature work. Not open for external contributors. Please pick any other **good first issues** :)

tarunsinghofficial:
## Summary
Stop service alerts used to show every agency-wide notice with the same green icon and no sense of what actually affects the stop you’re looking at. This orders alerts by relevance using OBA `allAffects`, styles them by severity (INFO / WARNING / SEVERE), and surfaces active windows plus cause/effect in the detail modal.

## What changed
- **Helpers:** `normalizeSeverity`, `alertAffects`, `orderAlertsByRelevance`, `activeWindowRange`, etc. in `serviceAlertsHelper.js` — relevant (this stop / its routes) first, then agency-wide; severity breaks ties within each group.
- **UI:** `ServiceAlerts` / `ServiceAlertItem` — group headings when both groups exist, severity badge + icon (not color-only), date range on the row, cause/effect/dates in the modal; rows are real `<button>`s (Enter/Space); Escape closes via `svelte:window`. Pagination stays at 3 per page over the ordered list.
- **Wiring:** `StopPane` passes `stopId` and `routeIds` into `ServiceAlerts`.
- **i18n:** New `service_alerts.*` keys in `en.json` (other locales fall back).
- **Fixtures/tests:** OBA Situation shape (`allAffects`, `{ lang, value }`); unit + component coverage for ordering, severity, grouping, pagination, and keyboard access.

Closes #534

## Summary by CodeRabbit

* **New Features**
  * Service alerts now display severity badges, clearer summaries, active time ranges, causes, and effects.
  * Alerts are prioritized by relevance to the selected stop or route and grouped accordingly.
  * Alert details open in an accessible modal with expanded information.
  * Added support for open-ended alert time windows and richer alert data.

* **Bug Fixes**
  * Improved keyboard activation and modal keyboard handling for service alerts.

* **Tests**
  * Added comprehensive coverage for severity, relevance sorting, pagination, formatting, and alert interactions.


aaronbrethorst:
### Code review

Found 1 issue:

1. The new "Cause" / "Effect" rows render raw API enum values, and the five `reason_*` keys added to `en.json` never match what OBA actually returns. The default server (`PUBLIC_OBA_SERVER_URL="https://api.pugetsound.onebusaway.org/"` in `.env.example`) emits GTFS-style causes — a live situation on stop `1_431` right now has `"reason": "CONSTRUCTION"`, not the TPEG `equipmentReason`/`environmentReason` spelling from the SDK typings. So `reasonLabel()` falls through its `translated === key` guard and prints `Cause: CONSTRUCTION` (untranslated, ALL CAPS, plus a `svelte-i18n` "message not found" warning on every render). `modalEffect` has no i18n mapping at all and prints `consequences[0].condition` verbatim (`diversion`, `stopMoved`, …). This PR also rewrote the fixtures from `reason: 'CONSTRUCTION'` to `reason: 'environmentReason'` in both `src/tests/fixtures/obaData.js` and `src/tests/mocks/handlers.js`, so the tests no longer exercise the value real deployments send. Consider handling the uppercase GTFS causes too (normalizing both vocabularies the way `normalizeSeverity` does), and either mapping the effect conditions or dropping that row.

https://github.com/OneBusAway/wayfinder/blob/a4be36932bba4fecd8ec3a6118084fa9154e8042/src/components/service-alerts/ServiceAlerts.svelte#L226-L244

aaronbrethorst (changes requested):
There's a lot to like here. The relevance ordering is genuinely correct — I checked `allAffects` against live Puget Sound data and the predicate groups it the way #534 asks. `normalizeSeverity` has a proper safe `info` default instead of trusting the API vocabulary. And converting the `<div role="button">` into a real `<button>` is a real accessibility improvement over the Enter-only handler we shipped in #364.

One thing blocks it, and it's the fixture rewrite rather than the feature logic.

**The new Cause/Effect rows don't match what OBA servers actually send.**

The five new i18n keys use the TPEG vocabulary (`reason_equipmentReason`, `reason_environmentReason`, `reason_personnelReason`, …). I queried the default server from our own `.env.example` — `https://api.pugetsound.onebusaway.org/`, stop `1_431` — and the live situation comes back as:

```json
{ "reason": "CONSTRUCTION", "severity": "noImpact", "consequences": [{ "condition": "" }] }
```

Uppercase GTFS vocabulary, not TPEG. So `reasonLabel('CONSTRUCTION')` looks up `service_alerts.reason_CONSTRUCTION`, misses, and falls through to printing the raw enum — riders see **CONSTRUCTION**, and svelte-i18n logs a "message not found" warning on every render. `modalEffect` is worse: `consequences?.[0]?.condition` is printed with no mapping at all, so any non-empty condition reaches the rider as a raw code.

What makes this blocking rather than a follow-up is that the PR rewrites both fixtures *away from* the real values — `reason: 'CONSTRUCTION'` → `reason: 'environmentReason'`, `effect: 'DETOUR'` → `consequences: [{ condition: 'diversion' }]`. Those fixtures had the right values before, so the test suite now certifies an API shape the server doesn't produce. Please restore the fixtures to the real GTFS values and map against those (keeping the graceful raw-text fallback for anything unrecognized).

**Non-blocking, worth a look while you're in there:**

- `activeWindowRange` coerces a missing `from` to `0`, and `formatAlertDate(0)` is finite — so an open-start window renders "Jan 1, 12:00 AM – <to>" instead of taking the "Until <to>" branch, which is currently dead code.
- `isAgencyWideAlert` is exported and unit-tested but no component calls it.
- `formatAlertDate` and the three-branch active-label logic are duplicated verbatim in `ServiceAlerts.svelte` and `ServiceAlertItem.svelte` — that belongs in the helper module you already added.
- The "includes severity in the accessible name" test only asserts `name: 'service_alerts.open_alert'`; the global i18n mock returns the key and drops `values`, so the severity interpolation isn't actually verified.
- Leftover unused `service_alerts.alert_modal` key in `en.json`.

Fix the cause/effect mapping and this is good to go.

tarunsinghofficial:
Thanks for the thorough review -especially checking against live Puget Sound data. That helped a lot.

I pushed a follow-up that should address everything you called out:

- Restored fixtures to the real GTFS values (`CONSTRUCTION`, `DETOUR`, `noImpact`)
- Cause/effect now map through alias tables instead of raw enum lookups, with a humanized fallback for unknown values
- Empty `condition` is omitted instead of shown raw
- Fixed the open-start date window (`Until …` branch works now)
- Moved shared date formatting into the helper
- Removed unused `isAgencyWideAlert` and `alert_modal`
- Tightened the a11y test so severity is actually asserted

Lint is clean and all tests pass. Ready for another look when you have a minute.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a thorough response to the review, and the blocking issue is fixed at the right layer rather than patched over.

I re-verified the cause/effect work myself. The fixtures are back on the real GTFS values — `reason: 'CONSTRUCTION'` / `'MAINTENANCE'`, `consequences: [{ condition: 'DETOUR' }]` / `'STOP_MOVED'`, `severity: 'noImpact'` — so the suite once again certifies the shape our servers actually send. `enumToken` normalizing case and separators means one alias table covers both the GTFS uppercase spelling and the TPEG one from the SDK typings, and routing unknown values through `humanizeEnum` instead of `$t()` means svelte-i18n is never asked for a key that doesn't exist, which was the "message not found" warning I was worried about. I checked all 28 `cause_*`/`effect_*` suffixes in your alias tables against `en.json` at head — every one resolves, no gaps. Returning `null` for an empty `condition` so the row is omitted handles the live Puget Sound case exactly.

The non-blocking items all got picked up too: `normalizeBound` makes the "Until {to}" branch reachable instead of dead, `isAgencyWideAlert` is gone, `formatActiveWindowLabel` and `formatAlertDate` are deduplicated into the helper, the accessible-name test now interpolates values and asserts the full string, and the leftover `alert_modal` key is removed.

One thing for a follow-up, not worth holding this on:

The group headings are keyed to absolute index — `absoluteIndex === 0` for "Affects this stop" and `absoluteIndex === relevantCount` for "Other alerts" — so they only render on whichever page the boundary happens to land on. With `alertsPerPage = 3` and, say, 4 relevant alerts out of 6, page 2 opens with a relevant alert and no heading above it. Ordering and content are unaffected, so it's cosmetic, but a rider paging through a busy stop will see unlabeled groups. Deriving the heading from position-within-page rather than absolute index would fix it.

Nice work on this one.

PR #580, opened 2026-07-28
Feat[Mobile-UX]: Improve trip planner layout on mobile and desktop

### LINKED ISSUE #577: Trip planner UX: map-first layout, relocate Recent Trips, show itinerary details without zooming
## Summary

Improve trip planner UX on mobile and desktop so users can plan trips, reopen recent trips, and read itinerary details without the map/form crowding each other out.

Today, after planning a trip, the Plan form and inline Recent Searches stay fully expanded while Trip Itineraries open in a bottom sheet underneath. On typical phone/laptop heights the sheet is too short to show leg details comfortably, the map is almost hidden, and zooming the browser out is what makes details visible. The route is drawn on the map but the camera does not auto-fit to it (unlike OBA route views).

This is a **layout / space-allocation** issue (and missing map fit), not missing OTP data. Fits under GSoC **mobile UX improvements**.

## Proposed design (rough sketch direction)

- **Map-first:** more map visible; plan / stops live in a compact bottom sheet (mobile) or left rail (desktop).
- **Move Recent Trips** out of the tall list under From/To into a small **icon / button** (top-right on mobile; inside the left rail on desktop) that opens recent trips when needed.
- **Remove inline trip history** from the main Plan form so itineraries + leg details have room.
- Keep From / To / Options / Plan Your Trip compact; after results load, form stays short and results/details own the remaining space.
- **Fit the map** to the active itinerary polylines (with padding for sheet/sidebar).

Optional / later: tab order or icon tabs for Stops vs Plan — not required for the first pass.

### Mobile
- Editing a plan: compact bottom sheet with From, To, Options, Plan.
- Recent Trips available via icon, not a long inline list.
- After plan: sheet shows itinerary pills + scrollable details; map remains usable above the sheet.
- Map auto-fits the selected itinerary with bottom padding for the sheet.

### Desktop
- Compact plan controls + scrollable itineraries/details in one left rail (avoid two tall stacked panels competing for height).
- Recent Trips control in the rail (not floating over map controls).
- Map auto-fits with left padding for the sidebar.

## Acceptance criteria

- [ ] Inline Recent Searches list is removed from the main Plan form; Recent Trips is reachable from a separate compact control.
- [ ] After planning a trip, itinerary summary + full leg timeline are readable by scrolling the results panel — no browser zoom required.
- [ ] Plan form stays compact while results are open (mobile + desktop).
- [ ] Active itinerary is drawn and the map fits bounds to it (including when switching itinerary tabs), with overlay padding.
- [ ] Closing / clearing results restores a usable plan-editing state.
- [ ] Desktop Recent Trips control does not collide with map zoom/locate controls.

## Notes

- Do **not** delete Recent Trips — relocate the entry point.
- Stops search already has a bottom-sheet pattern; prefer aligning Plan with that rather than redesigning Stops in the same PR.

tarunsinghofficial:
## Summary

This PR improves the trip planner layout so it works better on smaller screens and feels more intentional on desktop.

The main problem was that after planning a trip, the search form and recent trip history were still taking up too much vertical space while the itinerary results opened below. On mobile this made the UI feel cramped, covered too much of the map, and made trip details harder to scan. On desktop, the route would render on the map, but the layout still felt awkward and the available space was not being used very well.

This update changes the trip planner so the experience is more map-aware and gives the itinerary results room to breathe.

### 1. Fit the map to the active itinerary

After a trip is planned, the map now fits to the currently active itinerary instead of leaving the user to manually zoom or pan to find the route.

This also works when switching between itinerary options.

To make that framing feel correct across layouts:
- mobile uses sheet-aware padding so the route is not hidden behind the bottom sheet
- desktop keeps enough left-side padding for the trip planner rail

### 2. Reduce form clutter by hiding recent trips behind a toggle

Previously, recent trip history always appeared inline under the trip planner form, which took up a lot of space before the user even started reading trip results.

Now, recent trips are hidden behind a compact toggle:
- on mobile, the button is icon-only to save horizontal space
- on desktop, the label is shortened to `Recents`

This keeps the feature available without letting it dominate the layout.

### 3. Move the mobile trip planning flow into a single bottom sheet

On mobile, the trip planning flow now happens inside one bottom sheet:
- the form lives in the sheet
- recent trips are accessed from the same sheet
- itinerary results and leg details appear in the same sheet after planning

This makes the flow feel much more cohesive and avoids the old stacked layout where the form and results were competing for space.

### 4. Tune bottom-sheet behavior after planning

The sheet behavior was also adjusted so the transition between editing and viewing results feels more natural.

Current behavior:
- when opening trip planning on mobile, the sheet starts at a half-height state so the map is still visible
- after results load, the sheet stays at a usable half-height state instead of taking over the full screen immediately
- if the rider wants more detail, they can scroll inside the sheet or drag it higher

This gives a better default balance between map context and itinerary readability.

### 5. Keep desktop familiar while still improving results space

During iteration, one approach moved the full desktop plan flow into the lower sheet as well, but that made the first-open desktop state feel awkward.

The final desktop behavior keeps the more familiar structure:
- the plan form stays in the top search pane
- the itinerary sheet appears only after planning a trip

That keeps desktop more predictable while still benefiting from the improved itinerary presentation and map fitting.

## Why this helps

This PR mainly improves:
- **readability**, because itinerary content has more room
- **map visibility**, because the route is framed automatically and the UI does not overtake the screen
- **task flow**, because trip planning and trip reading now feel like part of the same interaction, especially on mobile
- **space efficiency**, because recent trips no longer permanently occupy valuable vertical space

## Notes

A small mobile-only compacting pass for the From/To fields was considered, but left out of this PR on purpose so this change stays focused on the larger trip planner flow and layout improvements.

That can be handled separately as a follow-up polish PR if needed.

## Testing

I verified the following manually:
- planning a trip on mobile opens the trip planner in the bottom sheet
- itinerary results render in the same sheet after planning
- switching itinerary tabs re-fits the map correctly
- recent trips are hidden by default and only shown when toggled
- desktop keeps the plan form in the search pane and only opens the itinerary sheet after planning
- mobile results keep enough map visible by default, while still allowing the user to drag or scroll for more details

## Screenshots
image
image
image


aaronbrethorst:
### Code review

Found 2 issues:

1. Mobile "Start here / End here" from the map context menu drops the picked location. `handleContextMenuTripPlan` now dispatches `planTripTabClicked` and then synchronously dispatches `setTripPlanLocation` in the same task. On a narrow viewport the first event flips `planTabActive`, which unmounts the `TripPlan` inside `SearchPane` (`embedTripPlan={!mobilePlanSheetOpen}`) and mounts a fresh one inside the plan sheet — but Svelte batches that state write, so `setTripPlanLocation` is still delivered to the *outgoing* instance. That instance sets `fromPlace`/`selectedFrom` and drops the pin, then its `onDestroy` runs `clearTripPlanPins`, so the sheet opens with an empty form and no pin. The added comment ("Dispatched before setTripPlanLocation so listeners are ready") describes the opposite of what happens; `maybeRestoreSharedTrip` in the same file guards this exact race with an extra `await tick()` a few lines below, which this path is missing.

https://github.com/OneBusAway/wayfinder/blob/47c2e2c91aa8e6f883cb487f25c2fa58f4cc6d73/src/components/search/SearchPane.svelte#L282-L288

2. On the mobile plan sheet, a trip that returns zero itineraries with no error now renders nothing at all. `showEmptyState` uses `error` as the proxy for "a plan attempt happened", so with `showForm` true the "No itineraries found" branch is suppressed whenever `error` is null — including after a completed plan. `mapGraphQLResponse` returns `{ plan: { itineraries: [] } }` with no `error` when OTP 2.x finds no path (`src/lib/otp/graphql.js`), and `handleTripPlan` maps that to `tripItineraries = []`, `tripPlanError = null`. The rider taps Plan Trip, the spinner stops, and the sheet is silent. The comment above the line states the intent correctly ("only after a plan attempt") but the condition doesn't track whether a plan was attempted. Desktop is unaffected since `showForm` is false there.

https://github.com/OneBusAway/wayfinder/blob/47c2e2c91aa8e6f883cb487f25c2fa58f4cc6d73/src/components/trip-planner/TripPlanModal.svelte#L218-L221

aaronbrethorst (changes requested):
The layout work here is good — moving the From/To form into the bottom sheet on narrow viewports is the right call, and I checked the pieces I expected to be fragile (RTL, dark mode contrast on the new toggle, the `sr-only md:hidden` / `hidden md:inline` label pair, duplicate DOM ids from two `TripPlan` mount points) and they all hold up.

But two mobile-only functional regressions came along with it, and both need fixing before this lands.

**1. "Start here / End here" from the map context menu loses the tapped location on mobile.**

In `SearchPane.handleContextMenuTripPlan`:

```js
window.dispatchEvent(new CustomEvent('planTripTabClicked'));
window.dispatchEvent(new CustomEvent('setTripPlanLocation', { detail: e.detail }));
```

`planTripTabClicked` sets `planTabActive = true` in `MapExperience`, which flips `mobilePlanSheetOpen`, which flips `embedTripPlan={!mobilePlanSheetOpen}` — tearing down the `TripPlan` in `SearchPane` and mounting a new one in the sheet. Both dispatches are synchronous in the same task, so `setTripPlanLocation` lands on the instance that's about to be destroyed, and its `onDestroy` calls `clearTripPlanPins`. The rider long-presses the map, picks "Start here", and the sheet opens with an empty form and no pin.

You actually diagnosed this exact race elsewhere in this same PR — `maybeRestoreSharedTrip` got an added `await tick()` with the comment "On mobile, TripPlan remounts inside the plan sheet after this event — wait one more tick so its listeners exist." The context-menu path is the same race and needs the same tick.

Also worth fixing while you're in there: the comment on those lines says "Dispatched before setTripPlanLocation so listeners are ready," which is the opposite of what the ordering actually achieves.

**2. A plan with zero itineraries and no error renders nothing on mobile.**

```js
let showEmptyState = $derived(!loading && !hasResults && (!showForm || error));
```

With `showForm` true (the mobile sheet), the "No itineraries found" branch requires `error` to be truthy. But `mapGraphQLResponse` returns `{ plan: { itineraries: [] } }` with no error when OTP finds no path, so `tripPlanError` stays null — the rider taps Plan Trip and gets zero feedback. The comment on that line ("only after a plan attempt") describes the right intent; the condition just doesn't track whether a plan was actually attempted. A `hasPlanned` flag would do it.

**One more worth considering, not blocking:** rotating a phone mid-plan crosses the 768px breakpoint (iPhone landscape is 844/852px), which flips `mobilePlanSheetOpen`, remounts `TripPlan`, and wipes `fromPlace`/`toPlace` plus the pins — they're component-local `$state`. Before this PR the component was always mounted in `SearchPane`, so nothing was lost. Lifting that state to the parent would fix it properly, but I don't want to expand this PR's scope; a follow-up is fine.

Ping me when the two blockers are pushed and I'll take another look.

tarunsinghofficial:
Hi Aaron, nice catch on both bugs. I have pushed fixes for both blockers:

- Context menu path now awaits a tick after planTripTabClicked before setTripPlanLocation (same pattern as maybeRestoreSharedTrip).
- Mobile empty state now uses a hasPlanned flag instead of error as the "plan attempted" proxy.

Added regression tests for both. Ready for another look when you have a minute.

aaronbrethorst:
### Code review

Found 1 issue:

1. The zero-result fix renders the empty state but places it off-screen on the mobile plan sheet. `hasPlanned` now correctly lets the `{:else if showEmptyState}` branch run, but that branch's container is `h-full` — 100% of the sheet's scroll viewport — and it is rendered *after* the embedded `TripPlan` form, so its vertically centered message starts a full viewport-height below the form. Concretely, at the default `half` detent the sheet is `0.55 × container`; on a ~700px-tall phone viewport that's a ~376px sheet, ~302px of scrollable body, while the From/To form (two labeled inputs + the button row) is ~226px. The centered message lands at ~377px — roughly 75px past the fold. It only becomes visible if the rider drags the sheet to `full`, and `handleTripPlan` deliberately does not snap for the zero-itinerary case (`else if (tripItineraries.length > 0)`). The same applies to the error box nested inside that container, so a `REQUEST_FAILED` plan is equally silent at the default detent. The desktop sheet is unaffected because it has no form above the message. Dropping `h-full` (or swapping it for `min-h-0`/no height) when `showForm` is true, or snapping to `full` when a plan attempt yields no results, would make the fix actually land.

https://github.com/OneBusAway/wayfinder/blob/d1ce61f7a4b27da09a6614b02cd4087771b7a628/src/components/trip-planner/TripPlanModal.svelte#L272-L278

aaronbrethorst (changes requested):
The first blocker is properly fixed. `handleContextMenuTripPlan` now puts `await tick()` between `planTripTabClicked` and `setTripPlanLocation`, so the sheet's `TripPlan` is mounted and listening before the location lands, and the outgoing instance's `onDestroy`/`clearTripPlanPins` can't take the pin with it. The comment is accurate now too, and you added a regression test for it. That was the subtler of the two and you got it right.

The second one is half fixed, and I want to be precise about which half. The state logic is exactly what I asked for: `tripPlanAttempted` set in `handleTripPlan`, cleared in `clearTripItineraries`, passed down as `hasPlanned`, and `showEmptyState` no longer abuses `error` as the proxy for "a plan happened." That part is right and tested.

**But the rider still sees nothing at the default detent.**

`BottomSheet` renders the sheet with an explicit `style:height="{sheetHeight}px"` and the content area is `flex-1 overflow-y-auto` inside it, so that body has a definite computed height. The empty state is:

```svelte
<div class="flex h-full flex-col items-center justify-center gap-3 py-12">
```

`h-full` therefore resolves to 100% of the body — and on mobile this block renders *after* the From/To form, not instead of it. So the vertically centered "No itineraries found" text sits roughly `formHeight + bodyHeight / 2` down the scroll content while only `bodyHeight` is visible. It's on screen only when the form is shorter than half the body, and at the `half` snap (`0.55 × containerHeight`, minus the grabber and header) that isn't true at phone viewport sizes — the message lands below the fold and needs a scroll to find. Same for the error box, since it's nested in that block.

So the symptom from my last review survives: the rider taps Plan Trip, the spinner stops, and the sheet looks unchanged. `handleTripPlan` also doesn't snap to `full` for the zero-result case, so nothing brings it into view.

Dropping `h-full` when `showForm` is true is the smallest fix — the block then sizes to its content and sits directly under the form where the rider is already looking. Snapping to `full` on a zero-result plan would also work, but I'd rather not move the sheet under them.

**Two non-blocking things while you're in there:**

- `tripPlanAttempted` isn't cleared when the rider edits From/To, only in `clearTripItineraries`. So "No itineraries found" lingers under the form while they're retyping a new trip, until they hit a field's X or leave the tab.
- The breakpoint has two sources of truth: `matchMedia('(max-width: 767px)')` drives the layout but `handleTripPlan` checks `window.innerWidth >= 768`. At fractional widths between them — browser zoom — those disagree and a desktop-layout sheet gets left at `half`.

The layout work itself is good and the race fix was the hard part. One CSS change and I'll merge it.

tarunsinghofficial:
@aaronbrethorst Thanks for the thorough review! 

I have fixed the below-the-fold empty state- dropped h-full when showForm is true, so on mobile the message sizes to its content and sits right under the form instead of centering past the visible body. Also handled both non-blocking notes: retyping From/To now clears the stale empty-state via clearTripItineraries, and handleTripPlan keys off isNarrowViewport (same matchMedia as the layout) instead of window.innerWidth so browser zoom can't split the breakpoint. 

Ready for another look :)

aaronbrethorst (approved):
Both blockers from the last two rounds are resolved at head, and I verified each one in the code rather than taking the commit messages for it.

The context-menu race is genuinely fixed — `SearchPane.svelte:282-289` now awaits a tick between `planTripTabClicked` and `setTripPlanLocation`, so the event lands on the remounted `TripPlan` inside the plan sheet instead of the outgoing instance, and the comment above it now describes what actually happens. The ordering test that asserts `setTripPlanLocation` isn't dispatched before the tick would fail if this regressed, which is the right kind of test.

The empty-state fix is exactly the one I asked for: `TripPlanModal.svelte:278` drops `h-full` when `showForm` is true and keeps full-height centering for the standalone desktop results sheet, with a comment explaining the asymmetry and a regression test asserting the class is absent. You also left the detent alone on a zero-result plan, which is what I preferred — no moving the sheet under the rider.

Nice catch consolidating the breakpoint too. `handleTripPlan` keying off `isNarrowViewport` instead of a second `window.innerWidth >= 768` check removes the disagreement under browser zoom.

Two things I noticed while verifying, neither worth holding this up:

The `clearTripItineraries()` you added at the top of `handleSearchInput` does more than clear the empty state — the parent's version also nulls `currentModal` and calls `clearAllPolylines()`. So on desktop, typing one character into From or To now tears down the results sheet and erases the drawn route mid-edit. That's consistent with what `clearInput` already does and it's a defensible read of "retyping starts a new trip," so I'm taking it as intended. One inconsistency if you ever revisit it: `clearInput` pairs that call with `clearTripUrl()` and `handleSearchInput` doesn't, so the shareable URL keeps the stale trip params after the results are gone.

The `py-12` on the empty state stayed, so on the shortest phones (~667px viewport) the message lands right about at the fold rather than comfortably above it. It clears fine on larger phones. Dropping the vertical padding when `showForm` is true would make it robust rather than marginal — worth a follow-up, not a blocker.

The rotation state loss from the first round is still open, as we agreed — that's a follow-up, not this PR's job.

Merging. Thanks for the patient turnaround on both rounds.

PR #583, opened 2026-07-29
fix:StopRoutesLayer partial route rendering failures via toast notifications

### LINKED ISSUE #579: StopRoutesLayer route shape failures are silent to the user
### Summary

Surfaced during review of #555, which implemented #530. That PR introduced a notification primitive (`src/stores/notificationStore.js` + `src/components/notification/Toast.svelte`) and wired it into the four paths #530 named. `StopRoutesLayer` is the one remaining `createPolyline` caller that never got the treatment — every failure there is still `console.error`-or-nothing.

This layer draws the routes a rider can board from the selected stop. When a route's shape can't be fetched or can't be drawn, that route silently vanishes from the map.

### Affected path

`src/components/map/StopRoutesLayer.svelte`, all three inside the per-route `Promise.all` in `drawRoutes`:

- **line 148** — `fetchRouteShape` throws; logged to console, `return`.
- **line 151** — `!shape.points`; bare `return`, not even a log.
- **line 175** — `createPolyline` returned `null` (undecodable shape); bare `return`, not even a log.

### Why this is lower severity than #530's cases

The degradation here is self-consistent, and deliberately so — see the comment at lines 145-147. A route that fails drops out of the lines, the legend, *and* the ring dots together, so the rider never sees a legend entry pointing at a line that isn't on the map. Contrast the `RouteMap` case fixed in #555, where the stops and vehicles stayed on screen and only the route line disappeared, which reads as a rendering bug.

So this is "the rider is quietly shown fewer routes than they can actually board" rather than "the map looks broken." Still worth surfacing — a rider comparing the map against the arrivals list below it has no way to know why a route is missing — but it didn't justify widening #555.

### Suggested direction

The primitive from #555 is already available and does the right thing here:

- `notifyPartialRouteShape()` (`src/lib/routeNotifications.js`) is the natural fit — some routes drew, some didn't, which is exactly the "partial" case. It's a `warning`, auto-dismisses after 6s, and won't clobber a retriable error thanks to the priority rule in `notificationStore.show`.
- Count attempted vs. drawn routes across the `Promise.all` and notify once at the end, rather than once per failed route — the store is a single slot, so N failures should still be one toast.
- Capture the id it returns into a component-scoped `notificationId` and pass it to `notifications.dismiss(notificationId)` on teardown, matching what `RouteMap`/`SearchPane`/`TripPlanModal` do after #555. Do **not** call the unscoped `notifications.dismiss()` — that would clear toasts owned by other components.
- The bare `return` at line 151 should at minimum get a `console.error` alongside the counter bump, for parity with line 148.

Worth checking whether the count should exclude routes dropped by the `token !== loadToken` supersede checks (lines 151, 167) — those aren't failures, they're a newer stop selection taking over, and counting them would fire a spurious warning every time the rider taps a second stop quickly.

antriksh-9:
### Summary
Surface partial route rendering failures in `StopRoutesLayer` using the existing notification infrastructure.

Previously, when a route shape failed to load, contained no points, or could not be decoded into a polyline, the affected route was silently omitted from the map (or only logged to the console). This change notifies riders when one or more routes cannot be rendered while preserving the existing behaviour of drawing all successfully rendered routes.

### Changes 

- Added route rendering counters to track attempted vs. successfully drawn routes.
- Log missing `shape.points` and polyline creation failures for consistency with shape fetch failures.
- Display a single `notifyPartialRouteShape()` warning when one or more routes fail to render.
- Store the returned `notificationId` and dismiss only that notification during component teardown.
- Exclude superseded route loads (token !== loadToken) from notification counts to avoid false warnings when users rapidly switch between stops. 

Testing

- ✅ npm run format
- ✅ npm run lint

---
<a href="https://app.blacksmith.sh/OneBusAway/codesmith/wayfinder/pr/583"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787945831&installation_model_id=429412&pr_number=583&repository=OneBusAway%2Fwayfinder&return_to=https%3A%2F%2Fgithub.com%2FOneBusAway%2Fwayfinder%2Fpull%2F583&signature=ae481f9e30381626406003b904b0ef5c5c9370cd58d58f1df6b74190a676b7b3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added partial-route notifications when some route shapes fail to load or render.
* **Bug Fixes**
  * Prevented partial notifications from being emitted after the map layer is torn down/destroyed.
  * Improved handling when polyline creation fails or completes after a newer selection, avoiding late/off-target updates.
* **Tests**
  * Extended coverage to assert partial-route notifications on shape-fetch failures and to verify error logging when polyline creation throws.

### DIFF
--- a/src/components/map/StopRoutesLayer.svelte
+++ b/src/components/map/StopRoutesLayer.svelte
@@ -114,6 +114,7 @@
 	}
 
 	let notificationId = null;
+	let isMounted = true;
 
 	async function drawRoutes(routes, colors, token) {
 		if (token !== loadToken) return;
@@ -171,20 +172,27 @@
 				// turn trip-expansion into a full redraw.
 				const promoted = untrack(() => promotedRouteId);
 				const isPromoted = promoted != null && route.id === promoted;
-				const polyline = await mapProvider.createPolyline(shape.points, {
-					color,
-					casing: true,
-					weight: weightFor(index),
-					pane: isPromoted ? ROUTE_PANE.PROMOTED : ROUTE_PANE.LINE,
-					casingPane: ROUTE_PANE.CASING
-				});
+				let polyline = null;
+
+				try {
+					polyline = await mapProvider.createPolyline(shape.points, {
+						color,
+						casing: true,
+						weight: weightFor(index),
+						pane: isPromoted ? ROUTE_PANE.PROMOTED : ROUTE_PANE.LINE,
+						casingPane: ROUTE_PANE.CASING
+					});
+				} catch (error) {
+					console.error('StopRoutesLayer: could not create polyline', route.id, error);
+					return;
+				}
+
 				if (token !== loadToken) {
 					// A supersede ran clearAllPolylines() before this create resolved
 					// (Google's createPolyline is async, awaiting importLibrary), so
 					// the polyline this call just attached to the map is an orphan of
 					// the selection that's already gone — take it back off.
 					mapProvider.removePolyline(polyline);
-					attemptedRoutes--;
 					return;
 				}
 				if (!polyline) {
@@ -241,6 +249,8 @@
 				routeStopIds = new Map(nextStopIds);
 			})
 		);
+
+		if (token !== loadToken || !isMounted) return;
 
 		if (attemptedRoutes > 0 && drawnRoutes < attemptedRoutes) {
 			notificationId = notifyPartialRouteShape();
@@ -470,6 +480,7 @@
 
 	onDestroy(() => {
 		loadToken++;
+		isMounted = false;
 		teardown();
 	});
 </script>
--- a/src/components/map/StopRoutesLayer.svelte
+++ b/src/components/map/StopRoutesLayer.svelte
@@ -26,6 +26,8 @@
 	// provider here would pull its whole map stack into the bundle regardless of
 	// which one PUBLIC_OBA_MAP_PROVIDER selects.
 	import { ROUTE_PANE } from '$lib/mapPanes.js';
+	import { notifyPartialRouteShape } from '$lib/routeNotifications';
+	import { notifications } from '$stores/notificationStore';
 
 	let {
 		mapProvider,
@@ -111,6 +113,8 @@
 		return { points: shapeData?.data?.entry?.points, stopIds };
 	}
 
+	let notificationId = null;
+
 	async function drawRoutes(routes, colors, token) {
 		if (token !== loadToken) return;
 
@@ -134,6 +138,8 @@
 		// order createPolyline resolved, which is shape-fetch race order, not
 		// activeRoutes order.
 		const drawnPolylines = [];
+		let attemptedRoutes = 0;
+		let drawnRoutes = 0;
 
 		await Promise.all(
 			routes.map(async (route, index) => {
@@ -146,9 +152,17 @@
 					// other routes still draw, and this one is simply absent from the
 					// lines, the legend, and the ring dots.
 					console.error('StopRoutesLayer: could not load shape', route.id, error);
+					attemptedRoutes++;
 					return;
 				}
-				if (token !== loadToken || !shape.points) return;
+				if (token !== loadToken) return;
+
+				attemptedRoutes++;
+
+				if (!shape.points) {
+					console.error('StopRoutesLayer: route shape contains no points', route.id);
+					return;
+				}
 
 				// untrack: this read happens after an await, so Svelte would not treat
 				// it as an effect dependency anyway — but reading it explicitly through
@@ -170,9 +184,15 @@
 					// the polyline this call just attached to the map is an orphan of
 					// the selection that's already gone — take it back off.
 					mapProvider.removePolyline(polyline);
+					attemptedRoutes--;
 					return;
 				}
-				if (!polyline) return;
+				if (!polyline) {
+					console.error('StopRoutesLayer: could not create polyline', route.id);
+					return;
+				}
+
+				drawnRoutes++;
 
 				// Retained so the promotion effect can re-pane this route later
 				// without redrawing it. isPromoted was already decided above (from
@@ -221,6 +241,10 @@
 				routeStopIds = new Map(nextStopIds);
 			})
 		);
+
+		if (attemptedRoutes > 0 && drawnRoutes < attemptedRoutes) {
+			notificationId = notifyPartialRouteShape();
+		}
 	}
 
 	function stopVehiclePolling() {
@@ -232,6 +256,7 @@
 	}
 
 	function teardown() {
+		notifications.dismiss(notificationId);
 		stopVehiclePolling();
 		// mapProvider can be null: a cold deep-link whose arrivals land before
 		// initMap() resolves mounts this layer with no provider yet, and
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. A superseded (or unmounted) load can still raise the toast. The `catch` bumps `attemptedRoutes` *before* the `token !== loadToken` check, and the final notify block has no token guard at all — so if a shape fetch rejects after the rider taps a second stop (or closes the sheet), `attemptedRoutes > 0 && drawnRoutes === 0` and a warning fires for a selection that is no longer on the map. This is the exact case #579 warned about ("counting them would fire a spurious warning every time the rider taps a second stop quickly") and that the PR description claims to have handled; the `attemptedRoutes--` at L187 covers only one of the three supersede exits. Worse on unmount: `onDestroy` already ran `teardown()`, so the toast has no owner left to dismiss it — the same hazard `RouteMap` guards with `if (!isMounted) return;` before `notifyRouteShapeFailed` (see its comment: "a toast raised now would have no owner to dismiss it"). Also note `MapExperience` unscoped-dismisses on stop selection, so this stale toast lands *after* the new stop cleared the slot.

https://github.com/OneBusAway/wayfinder/blob/6708067ad55b06af0962ba677d096c5843380da1/src/components/map/StopRoutesLayer.svelte#L243-L248

https://github.com/OneBusAway/wayfinder/blob/6708067ad55b06af0962ba677d096c5843380da1/src/components/map/StopRoutesLayer.svelte#L153-L158

2. A throwing `createPolyline` still leaves the failure silent, and now also suppresses the new toast. The `await mapProvider.createPolyline(...)` is unguarded, so a throw rejects the whole `Promise.all` — the notify block at L245 never runs, and the remaining routes are aborted mid-flight. This is reachable: Google's provider does an unguarded `await window.google.maps.importLibrary('geometry')`, and this file already has a test named "does not produce an unhandled rejection when createPolyline throws". `ca38834` fixed the identical shape in `TripPlanModal` with the remedy that applies here — "Catch per leg and log. The leg then counts as undrawn, so an existing partial-shape warning correctly covers it."

https://github.com/OneBusAway/wayfinder/blob/6708067ad55b06af0962ba677d096c5843380da1/src/components/map/StopRoutesLayer.svelte#L173-L181

aaronbrethorst (changes requested):
Thanks for taking this on — the counting approach is the right shape for #579, and scoping the dismiss into `teardown()` shows you were thinking about toast ownership, which is exactly the thing that's bitten us here before.

Two things need fixing before this lands. I verified both against the head commit.

**1. A superseded or unmounted load can still raise the toast.**

In the shape-fetch `catch`, `attemptedRoutes++` runs *before* the `if (token !== loadToken) return;` guard, and the notify block at the end of `drawRoutes` has no token guard at all:

```js
if (attemptedRoutes > 0 && drawnRoutes < attemptedRoutes) {
    notificationId = notifyPartialRouteShape();
}
```

So: rider taps stop A, shape fetches start, rider taps stop B. Stop A's fetches reject → `attemptedRoutes` is 1, `drawnRoutes` is 0 → a warning fires for a selection that's already gone. The successful path is fine, because it checks the token before incrementing; it's only the `catch` that leaks.

The unmount case is worse: `teardown()` dismisses whatever `notificationId` holds *at that moment*, then the still-pending `Promise.all` assigns a fresh id afterward — an orphan toast with nobody left to dismiss it. That's the same failure `RouteMap.svelte` guards against with `if (!isMounted) return;` and the comment "a toast raised now would have no owner to dismiss it."

The fix is a token (and mounted) check immediately before the notify block. Note that the `attemptedRoutes--` you already added on the supersede-after-`createPolyline` path is the same instinct — it just needs to cover all three exits instead of one.

**2. A throwing `createPolyline` suppresses the toast entirely.**

`const polyline = await mapProvider.createPolyline(...)` isn't wrapped, so a rejection takes down the whole `Promise.all` and the notify block never runs — the rider gets nothing at all in precisely the case where the most went wrong. It's reachable: `GoogleMapProvider.createPolyline` has an unguarded `await window.google.maps.importLibrary('geometry')`.

`ca38834` fixed the identical shape in `TripPlanModal`, and the approach there applies directly: catch per route and log, so the route counts as undrawn and the partial-shape warning covers it.

**One non-blocking note:** every prior commit in this toast lineage (`0e52278`, `7adf0f2`, `ca38834`) came with tests, and `StopRoutesLayer.test.js` is already set up well for this — neither the counting logic nor the scoped dismiss is currently exercised. A test that supersedes mid-fetch would have caught issue 1 directly.

Both fixes are small. Happy to re-review as soon as you push.

antriksh-9:
New tests to be added: 

- A superseded pending shape fetch does not trigger `notifyPartialRouteShape()`.
- An active partial-route notification is dismissed when the component unmounts.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blockers are genuinely fixed, and I verified each against your head commit rather than taking the diff at face value.

On the first one, you took a better route than the one I suggested. Instead of guarding all three supersede exits with `attemptedRoutes--`, you put a single gate on the only consumer — `if (token !== loadToken || !isMounted) return;` immediately before the notify block. That's equivalent and cleaner: `attemptedRoutes` and `drawnRoutes` are function-locals whose sole read is the line right below that gate, so the still-unguarded `attemptedRoutes++` in the shape-fetch `catch` is no longer observable. I also confirmed the redraw path is safe end to end — the effect does `++loadToken` then `teardown()` before the next `drawRoutes`, and `teardown()` dismisses the prior toast, so no toast can outlive the selection that raised it. `dismiss(null)` is a safe no-op, so the unconditional call in `teardown()` can't clobber another component's toast on the null-provider or empty-route paths.

The second is fixed exactly as `ca38834` did it in `TripPlanModal` — per-route try/catch, log, return, and since `attemptedRoutes++` already ran the route correctly counts as undrawn so the partial toast covers it. All four failure exits now log, where before a throwing `createPolyline` took down the whole `Promise.all` and killed the draw for every other route.

Two small things for whenever you're next in this file, neither blocking:

- `'StopRoutesLayer: could not create polyline'` is now emitted for two distinct failures — a thrown provider error and a `null` return. Grepping logs can't tell them apart.
- `isMounted` is actually redundant with the token check, since `onDestroy` increments `loadToken` before `teardown()`. It's harmless belt-and-braces, but this file is densely commented and the new guard arrives bare; a line of rationale like the one `RouteMap.svelte` carries would save a future reader the trace.

I saw your note that the supersede-mid-fetch and unmount-dismiss tests are still coming. Those are the two riskiest paths in the diff and I'd genuinely like them, but I called that non-blocking and the fixes are correct and green, so I'm landing this now — please send them as a follow-up PR.

Issue #529, opened 2026-06-27
colorUtils: silent fallbacks + missing contrast regression test + validate-env 8-digit hex gap


aaronbrethorst:
Follow-up from #513 (code-quality hardening; non-blocking).

### 1. Silent fallbacks in `darkenColor` / `lightenColor`
`src/lib/colorUtils.js` — both `darkenColor` (returns `#000000`) and `lightenColor` (returns `#ffffff`) swallow falsy/invalid input with no log. For `darkenColor` the black sentinel is indistinguishable from a legitimate extreme result (`darkenColor('#fff', 1.0)` is correctly `#000000`). The sibling `generatePalette` — fed the same `process.env.COLOR_BRAND_ACCENT` in `tailwind.config.js` — instead `console.warn`s and falls back to a visible brand color. Consider giving both helpers an optional `fallbackHex` param and a warn-on-bad-input path, mirroring `generatePalette`. (Update the tests at `src/tests/lib/colorUtils.test.js` that currently lock in the silent behavior.)

### 2. No contrast-ratio regression test
The `darkenColor` suite asserts the arithmetic (`#3d571c`, "is darker") but nothing pins the actual a11y goal: white text on the computed `brand-accent-dark` must clear 4.5:1. If the `0.15` darken amount or a custom `COLOR_BRAND_ACCENT` changes, the hover could silently drop below AA with all tests still green. Add a test asserting contrast >= 4.5:1 for the default-derived token.

### 3. validate-env / hexToRgb 8-digit hex mismatch
`scripts/validate-env.js` accepts `#RRGGBBAA`, but `hexToRgb` (`colorUtils.js`) only matches 3- or 6-digit hex. So `COLOR_BRAND_ACCENT="#486621ff"` passes `npm run validate-env` yet causes `darkenColor` to silently return `#000000`. Either teach `hexToRgb` to handle 8-digit hex or tighten the env-schema color regex so a value can't validate yet fail `hexToRgb`.

Surfaced during review of #513; none of these block that PR.

PR #584, opened 2026-08-01
Improve color utility error handling and validation

antriksh-9:
## Summary

This PR currently addresses the first two follow-up item from #529:

- [x] Add fallback handling and warning/error logging to `darkenColor` and `lightenColor`.
- [x] Add contrast-ratio regression test.
- [ ] Resolve the `validate-env` / `hexToRgb` 8-digit hex mismatch.

Unit tests have been updated to cover the new fallback behavior.

Working on the last fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved color adjustments for invalid inputs by supporting fallback colors.
  * Added predictable default behavior when no valid fallback is available, returning black or white as appropriate.
  * Added warnings and errors to clarify invalid color handling.

### DIFF
--- a/src/lib/colorUtils.js
+++ b/src/lib/colorUtils.js
@@ -124,11 +124,16 @@ export function generatePalette(baseHex, fallbackHex = '#486621') {
  * @param {number} amount - Amount to darken (0-1, where 1 is pure black)
  * @returns {string} Darkened hex color
  */
-export function darkenColor(hexColor, amount) {
-	if (!hexColor) return '#000000';
-
+export function darkenColor(hexColor, amount, fallbackHex = '#486621') {
 	const rgb = hexToRgb(hexColor);
-	if (!rgb) return '#000000';
+	if (!rgb) {
+		if (fallbackHex === null) {
+			console.error(`Invalid hex color "${hexColor}" and no fallback available`);
+			return '#000000';
+		}
+		console.warn(`Invalid hex color "${hexColor}", falling back to "${fallbackHex}"`);
+		return darkenColor(fallbackHex, amount, null);
+	}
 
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This mirrors the `generatePalette` fallback pattern cleanly, and the `null` sentinel on the recursive call is the right way to bound the recursion at depth one.

I checked the part I expected to be risky — that `darkenColor(null, …)` no longer returns `#000000` is a real behavior change, so I traced the callers rather than assuming. `tailwind.config.js:20` is the only non-test caller of either function, and it already passes `process.env.COLOR_BRAND_ACCENT || '#486621'`, so truthy-invalid input can't reach the new warn path in production. `lightenColor` has no non-test callers at all. I also hand-checked the new expected values (`#3d571c`, `#637d42`, `#808080`) and they're arithmetically right.

One fit-and-finish note for a follow-up, not worth holding this on:

The new contrast test reads `process.env.COLOR_BRAND_ACCENT` bare, while `tailwind.config.js` reads `process.env.COLOR_BRAND_ACCENT || '#486621'`. The Test workflow never does `cp .env.example .env` — only the Build job does — so in CI that variable is undefined and the test reaches its assertion *through* the new invalid-input fallback, with the `console.warn` swallowed by the spy installed in the same `describe`. It lands on the same `#3d571c` Tailwind produces, so it isn't wrong today, but the green depends on `darkenColor`'s internal default staying in sync with Tailwind's `||` default. Mirroring the Tailwind expression would make it a true regression test.

I'm landing items 1 and 2 now rather than holding them for the `validate-env` / `hexToRgb` 8-digit hex mismatch — please open a follow-up PR for that one.

PR #585, opened 2026-08-02
feat: favorites for stops and routes

### LINKED ISSUE #581: Favorites / Bookmarks for stops and routes
## Summary

Add a favorites feature so riders can save commonly used **stops** and **routes** for one-tap access, without searching every time.

This supersedes [#315](https://github.com/OneBusAway/wayfinder/issues/315). Several contributors attempted this earlier; none of those PRs landed. This issue is a fresh, fully-specced rewrite of the requirements, drawn largely from Reviewer's review feedback on [#436](https://github.com/OneBusAway/wayfinder/pull/436) and [#414](https://github.com/OneBusAway/wayfinder/pull/414).

## Scope

**In scope**
- Favorite / unfavorite **stops** and **routes**
- Persist in `localStorage` (no backend)
- Surface an inline favorites section in the Stops & Stations search tab when the query is empty
- Star toggle on stop and route views

**Out of scope (tracked separately)**
- Favoriting trip itineraries / pinning origin–destination pairs — see follow-up issue (blocked on [#580](https://github.com/OneBusAway/wayfinder/pull/580); full itineraries are also a bad fit because OTP times are absolute and never refreshed)
- Standalone `/favorites` page — deferred; revisit if the list grows past a comfortable scroll area
- Cross-tab `storage` sync — known limitation for v1
- Background refresh of denormalized labels when a stop/route is renamed server-side

Stops-only is intentionally **not** acceptable for the implementing PR: Reviewer blocked #436 for claiming to close #315 while only wiring stop favoriting UI.

### Storage

- Key: `wayfinder_favorites`
- Pattern: mirror [`recentTripsStore.js`](https://github.com/OneBusAway/wayfinder/blob/main/src/stores/recentTripsStore.js) (JSON array, `browser` guard, `try/catch` load, validate-and-filter on read, write inside mutations). **Not** `tripOptionsStore` (that’s for scalar preferences).
- Cap: `MAX_FAVORITES = 50` (drop oldest)
- Deduplicate on `type + id`
- `id` is always the full agency-prefixed OBA ID — never route `shortName` (Reviewer's #414 critical item)

### Record shape (denormalized so the list needs zero API calls)

```js
// stop
{ schemaVersion: 1, type: 'stop', id: '1_75403', name: 'Pine St & 3rd Ave',
  code: '75403', direction: 'N', lat: 47.6105, lon: -122.3363, savedAt: … }

// route
{ schemaVersion: 1, type: 'route', id: '1_100479', shortName: '10',
  description: '…', routeType: 3, savedAt: … }
```

`isValidFavorite` must reject stop entries whose `lat`/`lon` are not finite numbers (Reviewer's #436 critical: null coords later crash `flyTo` / `addMarker`). Use `!= null` / `Number.isFinite`, never truthiness, so `lat === 0` survives. Prefer `??` over `||` for optional strings.

### UI surfaces

| Surface | Placement |
|---------|-----------|
| Map stop sheet (`StopBottomSheet`) | Action row next to Refresh / Stop Info |
| Standalone stop header (`StopPageHeader`) | Beside the stop name; pass `stopLat` / `stopLon` / `stopCode` from both `/stops/[stopID]` pages |
| Route sheet (`RouteModal`) | Header between title and close |
| Search pane (Stops tab) | Inline `FavoritesList` above results when query is empty |

Do **not** put stars on `SearchResultItem` rows (nested interactive controls — Reviewer's #436 a11y note).

### Store / component conventions

- Export a `derived` `favoriteKeys` `Set` of `` `${type}:${id}` ``; components use `$derived($favoriteKeys.has(key))` — no `$effect` + manual `subscribe`
- No dead `isFavorite()` API if unused outside tests
- Native `<button>` only — no custom Enter/Space `keydown` handlers
- List item and delete control as flex **siblings**, not nested
- i18n via `$t()` / keys in `en.json` only (lazy-loaded locales fall back)
- Dark mode (`dark:`) and brand colors (`text-brand-accent`, etc.)

## Acceptance criteria

- [ ] Rider can favorite and unfavorite a stop from the map bottom sheet and from standalone stop pages
- [ ] Rider can favorite and unfavorite a route from the route sheet
- [ ] Favorites appear in the Stops tab when the search query is empty; clicking one opens the stop sheet or loads the route
- [ ] Standalone stop pages pass coordinates into the header so favorites never save `lat`/`lon` as null
- [ ] Click handlers guard on map readiness (`mapLoaded`)
- [ ] Corrupted / malformed localStorage entries are dropped on load; quota / disabled storage fails soft
- [ ] `npm run lint` passes from a clean checkout
- [ ] Tests cover store edge cases (corrupt JSON, missing/zero coords, quota) and list remove / Clear All without nested-button click leaks
- [ ] Coverage stays ≥ 70%; PR stays focused (~stops+routes favorites only, no unrelated refactors)

## Implementation ownership

I’ll take this as part of GSoC 2026 work. Happy to adjust scope or UX if maintainers prefer a different surface (e.g. a dedicated Favorites tab instead of an inline section).

tarunsinghofficial:
## Summary
Adds local favorites (bookmarks) for stops and routes so riders can re-open everyday places without searching again.

## What changed
- **Store** — `favoritesStore` persists a capped list (max 50) in `localStorage`, with dedupe, validation (stops need finite lat/lon), and add / remove / toggle / clearAll
- **Star toggle** — `FavoriteToggle` on the stop bottom sheet, standalone stop pages, and the route sheet header
- **Floating control** — map FAB opens a favorites panel (`FavoritesFloatingControl` + `FavoritesList`). Mobile: below the search pane. Desktop: top-right on the map
- **Feedback** — short success toasts on save/remove (“Saved to favorites.” / “Removed from favorites.”)
- **i18n** — English keys under `favorites.*` and `notifications.favorite_*`

## How it works
- Favorites are denormalized snapshots (name, code, coords / shortName, etc.) so the list renders with no API call
- Selecting a favorite stop flies the map and opens the stop sheet; selecting a route reuses the existing `routeSelectedFromModal` path
- Item open and remove are sibling buttons (no nested interactive controls

Closes #581 

## Working Screenshots
image
image
image

image

## Summary by CodeRabbit

* **New Features**
  * Added persistent favorites for stops and routes.
  * Added favorite buttons to stop and route views.
  * Added a map favorites panel with counts, selection, removal, and clear-all actions.
  * Added success notifications when favorites are saved or removed.

* **Bug Fixes**
  * Improved notification styling, including success messages.
  * Adjusted the route legend position to avoid overlapping map controls.

* **Tests**
  * Added comprehensive coverage for favorites, notifications, and related controls.

### DIFF
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -34,6 +34,7 @@
 	import TripOptionsModal from '$components/trip-planner/TripOptionsModal.svelte';
 	import { showTripOptionsModal } from '$stores/tripOptionsStore';
 	import { mapStopPath } from '$lib/mapStopUrl.js';
+	import { removeAgencyPrefix } from '$lib/utils';
 	import { clearVehicleMarkersMap } from '$lib/vehicleUtils';
 	import { activeRoutesFromArrivals, assignRouteColors } from '$lib/activeRoutes.js';
 
@@ -144,15 +145,14 @@
 	 */
 	function handleFavoriteStopClick(favorite) {
 		if (!mapProvider) return;
+		// Seed the marker so the selection effect has something to highlight, then
+		// hand off. Framing is the effect's job — flying here too would fight it.
 		mapProvider.addMarker({
 			stop: favorite,
 			position: { lat: favorite.lat, lng: favorite.lon },
 			onClick: () => handleStopMarkerSelect(favorite)
 		});
-		mapProvider.flyTo(favorite.lat, favorite.lon, 20);
-		setTimeout(() => {
-			handleStopMarkerSelect(favorite);
-		}, 100);
+		handleStopMarkerSelect(favorite);
 	}
 
 	/**
@@ -161,13 +161,16 @@
 	 * @param {Object} favorite
 	 */
 	function handleFavoriteRouteClick(favorite) {
+		// The listener drives the map straight away (clearAllPolylines et al), so
+		// don't fire before MapContainer has handed us a provider.
+		if (!mapProvider) return;
 		window.dispatchEvent(
 			new CustomEvent('routeSelectedFromModal', {
 				detail: {
 					route: {
 						id: favorite.id,
 						shortName: favorite.shortName,
-						nullSafeShortName: favorite.shortName,
+						nullSafeShortName: favorite.shortName ?? removeAgencyPrefix(favorite.id),
 						description: favorite.description,
 						type: favorite.routeType
 					}
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -6,6 +6,7 @@
 	import MapContainer from '$components/MapContainer.svelte';
 	import RouteModal from '$components/routes/RouteModal.svelte';
 	import ViewAllRoutesModal from '$components/routes/ViewAllRoutesModal.svelte';
+	import FavoritesFloatingControl from '$components/favorites/FavoritesFloatingControl.svelte';
 	import { isLoading } from 'svelte-i18n';
 	import AlertsModal from '$components/navigation/AlertsModal.svelte';
 	import { onMount, onDestroy } from 'svelte';
@@ -135,6 +136,44 @@
 		// structured-clones its state argument (DataCloneError on a proxy). Snapshot
 		// yields a plain, clone-safe copy.
 		pushState(mapStopPath(stopData.id), { stopData: $state.snapshot(stopData) });
+	}
+
+	/**
+	 * Open a favorited stop from the map floating control.
+	 * @param {Object} favorite
+	 */
+	function handleFavoriteStopClick(favorite) {
+		if (!mapProvider) return;
+		mapProvider.addMarker({
+			stop: favorite,
+			position: { lat: favorite.lat, lng: favorite.lon },
+			onClick: () => handleStopMarkerSelect(favorite)
+		});
+		mapProvider.flyTo(favorite.lat, favorite.lon, 20);
+		setTimeout(() => {
+			handleStopMarkerSelect(favorite);
+		}, 100);
+	}
+
+	/**
+	 * Open a favorited route via the same event ViewAllRoutesModal uses, so
+	 * SearchPane's existing handleRouteClick path loads shapes and vehicles.
+	 * @param {Object} favorite
+	 */
+	function handleFavoriteRouteClick(favorite) {
+		window.dispatchEvent(
+			new CustomEvent('routeSelectedFromModal', {
+				detail: {
+					route: {
+						id: favorite.id,
+						shortName: favorite.shortName,
+						nullSafeShortName: favorite.shortName,
+						description: favorite.description,
+						type: favorite.routeType
+					}
+				}
+			})
+		);
 	}
 
 	$effect(() => {
@@ -504,6 +543,10 @@
 {:else}
 	<h1 class="sr-only">{PUBLIC_OBA_REGION_NAME}</h1>
 	<div class="pointer-events-none absolute bottom-0 left-0 right-0 top-0 z-40">
+		<FavoritesFloatingControl
+			onStopClick={handleFavoriteStopClick}
+			onRouteClick={handleFavoriteRouteClick}
+		/>
 		<!-- Top spacing is padding (not margin) so h-full keeps the column's bottom
 		     edge — where the sheet anchors — exactly at the viewport bottom. Below md,
 		     horizontal margins live on the search wrapper and on each pane (not the
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -571,9 +571,11 @@
 			</div>
 
 			<!-- Mobile: sit in flow below the search pane. Desktop: pin to the map's
-			     top-right (absolute against the full-screen overlay ancestor). -->
+			     top-right (absolute against the full-screen overlay ancestor).
+			     Wrapper stays pointer-events transparent (and shrink-wrapped) so it
+			     cannot steal map pans; the control itself opts back in. -->
 			<div
-				class="pointer-events-auto relative z-30 mx-2 mt-2 flex justify-end md:absolute md:right-4 md:top-4 md:mx-0 md:mt-0"
+				class="relative z-30 mx-2 mt-2 w-fit self-end md:absolute md:right-4 md:top-4 md:mx-0 md:mt-0"
 			>
 				<FavoritesFloatingControl
 					onStopClick={handleFavoriteStopClick}
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -543,10 +543,6 @@
 {:else}
 	<h1 class="sr-only">{PUBLIC_OBA_REGION_NAME}</h1>
 	<div class="pointer-events-none absolute bottom-0 left-0 right-0 top-0 z-40">
-		<FavoritesFloatingControl
-			onStopClick={handleFavoriteStopClick}
-			onRouteClick={handleFavoriteRouteClick}
-		/>
 		<!-- Top spacing is padding (not margin) so h-full keeps the column's bottom
 		     edge — where the sheet anchors — exactly at the viewport bottom. Below md,
 		     horizontal margins live on the search wrapper and on each pane (not the
@@ -572,6 +568,17 @@
 						<SurveyLauncher />
 					{/snippet}
 				</SearchPane>
+			</div>
+
+			<!-- Mobile: sit in flow below the search pane. Desktop: pin to the map's
+			     top-right (absolute against the full-screen overlay ancestor). -->
+			<div
+				class="pointer-events-auto relative z-30 mx-2 mt-2 flex justify-end md:absolute md:right-4 md:top-4 md:mx-0 md:mt-0"
+			>
+				<FavoritesFloatingControl
+					onStopClick={handleFavoriteStopClick}
+					onRouteClick={handleFavoriteRouteClick}
+				/>
 			</div>
 
 			<div class="relative mt-2 flex-1 md:mt-4">
--- a/src/components/favorites/FavoriteToggle.svelte
+++ b/src/components/favorites/FavoriteToggle.svelte
@@ -13,7 +13,7 @@
 	@prop {string} [shortName] - Required for type=route
 	@prop {string|null} [description]
 	@prop {number|null} [routeType]
-	@prop {string} [class] - Extra classes on the button
+	@prop {string} [class] - Extra classes on the button (include h-/w- to override default size)
 -->
 <script>
 	import { FontAwesomeIcon } from '@fortawesome/svelte-fontawesome';
@@ -43,6 +43,9 @@
 
 	let label = $derived(isFav ? $t('favorites.remove') : $t('favorites.add'));
 	let icon = $derived(isFav ? faStarSolid : faStarRegular);
+	// Omit default size when the caller passes size classes — Tailwind cannot
+	// override conflicting utilities by HTML class order.
+	let sizeClass = $derived(className ? '' : 'h-10 w-12');
 
 	function handleClick() {
 		const result = favorites.toggle({
@@ -72,7 +75,7 @@
 	aria-pressed={isFav}
 	aria-label={label}
 	title={label}
-	class="flex h-10 w-12 flex-none items-center justify-center rounded-xl border border-gray-300 text-black hover:bg-gray-100 dark:border-gray-600 dark:text-white dark:hover:bg-gray-700 {className}"
+	class="flex {sizeClass} flex-none items-center justify-center rounded-xl border border-gray-300 text-black hover:bg-gray-100 dark:border-gray-600 dark:text-white dark:hover:bg-gray-700 {className}"
 >
 	<!-- {#key} remounts FontAwesome when the glyph swaps; the SVG component
 	     does not always update in place when only the icon prop changes. -->
--- a/src/components/favorites/FavoriteToggle.svelte
+++ b/src/components/favorites/FavoriteToggle.svelte
@@ -21,6 +21,7 @@
 	import { faStar as faStarRegular } from '@fortawesome/free-regular-svg-icons';
 	import { t } from 'svelte-i18n';
 	import { favorites } from '$stores/favoritesStore';
+	import { notifyFavoriteSaved, notifyFavoriteRemoved } from '$lib/favoriteNotifications';
 
 	let {
 		type,
@@ -44,7 +45,7 @@
 	let icon = $derived(isFav ? faStarSolid : faStarRegular);
 
 	function handleClick() {
-		favorites.toggle({
+		const result = favorites.toggle({
 			type,
 			id,
 			name,
@@ -56,6 +57,12 @@
 			description,
 			routeType
 		});
+
+		if (result === 'added') {
+			notifyFavoriteSaved();
+		} else if (result === 'removed') {
+			notifyFavoriteRemoved();
+		}
 	}
 </script>
 
--- a/src/components/favorites/FavoriteToggle.svelte
+++ b/src/components/favorites/FavoriteToggle.svelte
@@ -1,0 +1,71 @@
+<!--
+	@component
+	Star button that toggles a stop or route in the favorites store.
+	Membership is derived from `favoriteKeys` — no local isFavorite state.
+
+	@prop {'stop'|'route'} type
+	@prop {string} id - Full agency-prefixed OBA id
+	@prop {string} [name] - Stop name (required for type=stop)
+	@prop {string|null} [code]
+	@prop {string|null} [direction]
+	@prop {number} [lat] - Required for type=stop
+	@prop {number} [lon] - Required for type=stop
+	@prop {string} [shortName] - Required for type=route
+	@prop {string|null} [description]
+	@prop {number|null} [routeType]
+	@prop {string} [class] - Extra classes on the button
+-->
+<script>
+	import { FontAwesomeIcon } from '@fortawesome/svelte-fontawesome';
+	import { faStar as faStarSolid } from '@fortawesome/free-solid-svg-icons';
+	import { faStar as faStarRegular } from '@fortawesome/free-regular-svg-icons';
+	import { t } from 'svelte-i18n';
+	import { favorites, favoriteKeys } from '$stores/favoritesStore';
+
+	let {
+		type,
+		id,
+		name = null,
+		code = null,
+		direction = null,
+		lat = null,
+		lon = null,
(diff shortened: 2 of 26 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The favorites control wrapper carries `pointer-events-auto` on the wrapper `<div>` rather than on the control itself. On desktop the wrapper is `md:absolute` and shrink-wraps to the button, so this is harmless — but below `md` it is a static, block-level `flex justify-end` box spanning the full column width at the height of the 44px star button. Since it lives inside the `pointer-events-none absolute inset-0 z-40` overlay whose entire purpose is to let map drags/taps pass through, this creates an invisible full-width ~44px dead strip over the map just below the search pane on phones: pans and taps that start there never reach Leaflet/Google. Note the sibling markup in this same file deliberately keeps the wrapper transparent and passes `cssClasses="pointer-events-auto"` down to `SearchPane` instead. Moving `pointer-events-auto` onto `FavoritesFloatingControl`'s root (or adding `w-fit self-end` to the wrapper) restores pass-through.

https://github.com/OneBusAway/wayfinder/blob/8cdfeda726930c9f5f5038d24e64d97a40c7ddce/src/components/MapExperience.svelte#L573-L579

aaronbrethorst (changes requested):
This is a strong piece of work. The store is the part I'd normally expect to go wrong and it doesn't: the `browser` guard makes it SSR-safe, `isValidFavorite` uses `Number.isFinite` so a stop at lat/lon 0 survives instead of being silently dropped, the dedupe-then-cap ordering in `add` is right, and `persist` can't take the app down when localStorage is full or blocked. Denormalized snapshots so the list renders with no API call is the correct call. And every new component ships with a real test file, which is not the norm for a change this size.

One thing blocks the merge, plus two accessibility items I'd like folded into the same round.

## The favorites wrapper puts a dead strip across the map on mobile

https://github.com/OneBusAway/wayfinder/blob/8cdfeda726930c9f5f5038d24e64d97a40c7ddce/src/components/MapExperience.svelte#L573-L581

That wrapper lives inside the `pointer-events-none absolute inset-0 z-40` overlay, which is `pointer-events-none` precisely so map drags pass through it. At `md` and up the wrapper is `md:absolute` and shrink-wraps to the 44px button, so it's fine. Below `md` it's a static block-level `flex justify-end` box that spans the full column width at the button's height — and it carries `pointer-events-auto`. The result is an invisible full-width ~44px band sitting directly under the search pane that swallows pans and taps meant for the map.

The fix is the pattern this file already uses twelve lines up: `SearchPane` gets `cssClasses="pointer-events-auto"` on the pane itself while its wrapper `<div class="mx-2 md:mx-0">` stays transparent. Do the same here — move `pointer-events-auto` off the wrapper and onto the control's own button (or give the wrapper `w-fit self-end` so it stops spanning the row).

## Accessibility, worth fixing in the same pass

**`aria-modal="true"` on a panel that isn't modal.** `FavoritesFloatingControl.svelte` declares `role="dialog" aria-modal="true"`, but there's no focus trap, no close control inside the panel, and the page behind stays fully interactive. `aria-modal` tells assistive tech to hide everything else, so a screen reader user can tab out of a panel their AT has told them is the whole world. The Escape handler and click-outside dismissal you've got are the right behaviors for a popover — just drop `aria-modal` and keep `role="dialog"`.

**The star ends up inside the `<h1>`.** `StopPageHeader.svelte:35-49` nests `FavoriteToggle` in the heading, so the button's `aria-label` folds into the heading's accessible name and heading navigation announces "Pine St & 3rd Ave Add to favorites". Making the toggle a sibling of the `<h1>` inside the same flex container keeps the layout and fixes the announcement.

## Not blocking

The remove button on each row is `opacity-0 group-hover:opacity-100` with no `focus-visible:` counterpart, so it's invisible to keyboard users and effectively unreachable on touch — the first tap lands on the card and closes the panel. I'm not holding the PR on this because it's copied verbatim from `RecentTripsList.svelte:64`, so it's an existing house pattern rather than something you introduced. Adding `focus-visible:opacity-100` here anyway would be a nice improvement, and fixing both lists is a good follow-up.

Also worth a sanity check: #581 reads like it wants an inline favorites list in the Stops tab when the search field is empty, and this PR puts the list behind the map FAB instead. The FAB is a reasonable design and I'm not asking you to change it — I'd just like to know that was a deliberate choice rather than a missed requirement, so we can close #581 cleanly.

One logistical note: I just merged #580, which touches `SearchPane.svelte` and `MapExperience.svelte`, so you'll likely need to merge `develop` in before this is mergeable again.

Ping me when the pointer-events fix is up and I'll get straight back to it.

tarunsinghofficial:
Hey @aaronbrethorst This one is ready to be merged

aaronbrethorst:
### Code review

Found 1 issue:

1. Removing an item from the favorites list closes the whole panel. `handleWindowClick` treats any click whose `event.target` is no longer inside `rootEl` as an outside click, but the ✕ button on a row (and the Clear All button, which lives inside `{#if items.length > 0}`) destroys its own DOM node as part of the removal. Svelte 5.2.8 schedules that DOM update with `queueMicrotask` (`svelte/src/internal/client/runtime.js:577`, and there is no sync flush around DOM event handlers in this version), and a browser runs a microtask checkpoint between event listeners — so by the time the click bubbles from the delegated handler to the `<svelte:window onclick>` listener, the row is already detached, `rootEl.contains(event.target)` is `false`, and `close()` runs with the default `restoreFocus: true`. Net effect for a real user: deleting one favorite dismisses the panel and yanks focus back to the FAB, so you can't remove two in a row. I verified the listener/microtask ordering in a real browser with a minimal repro (delegated handler removes the node in a microtask; the window listener then sees `contains(target) === false`). The tests can't catch this — `user.click()` dispatches from JS, so no microtask checkpoint runs between the two listeners, and `favorites.remove` is mocked so the row never leaves the DOM. Guarding on `rootEl.contains(event.target) || !event.target.isConnected` (or checking containment during the capture phase / at `pointerdown`) would fix it.

https://github.com/OneBusAway/wayfinder/blob/88576b27421e2d8055ae8432f57ef63f4a404c7f/src/components/favorites/FavoritesFloatingControl.svelte#L55-L62

aaronbrethorst (changes requested):
All three of my earlier asks are properly addressed, and I checked each one rather than taking it on faith:

- `pointer-events-auto` moved off the wrapper and onto `FavoritesFloatingControl`'s own root, and the wrapper picked up `w-fit self-end` as well. The invisible dead strip over the map on phones is gone.
- `aria-modal` dropped, `role="dialog"` kept, with a test pinning it.
- The toggle is now a sibling of the `<h1>` inside a shared flex container, and there's a test asserting the heading's accessible name is exactly the stop name and that it doesn't contain the toggle. That's the right test to write for it.

The store itself holds up well under scrutiny — validate, normalize, dedupe, cap — and the per-unit test files are real tests, not coverage padding.

One blocker, and it's a small fix.

## Removing a favorite closes the entire panel

`handleWindowClick` treats any click whose `event.target` is no longer inside `rootEl` as an outside click:

https://github.com/OneBusAway/wayfinder/blob/88576b27421e2d8055ae8432f57ef63f4a404c7f/src/components/favorites/FavoritesFloatingControl.svelte#L55-L62

But a row's remove button destroys its own DOM node as part of the removal. We're on Svelte 5.2.8, which schedules that DOM update with `queueMicrotask` (`svelte/src/internal/client/runtime.js:577`) and has no synchronous flush around DOM event handlers. Browsers run a microtask checkpoint between event listeners, so the sequence is:

1. Click the row's remove button; its handler calls `favorites.remove(...)`.
2. Microtask checkpoint fires between listeners — Svelte flushes, the keyed `{#each}` row and its button are detached.
3. The click finishes bubbling to `<svelte:window onclick>`, which sees a detached `event.target`.
4. `rootEl.contains(event.target)` is `false`, so `close()` runs with the default `restoreFocus: true`.

For a rider, that means deleting one favorite dismisses the panel and throws focus back to the star button — so you can't remove two in a row without reopening. Clear All has the same problem, since it lives inside `{#if items.length > 0}` and destroys itself once the list empties.

The tests can't catch this: `user.click()` dispatches from JS so no microtask checkpoint runs between the two listeners, and `favorites.remove` is mocked in those tests so the row never actually leaves the DOM.

Cheapest fix is to treat a detached target as "not outside":

```js
if (!rootEl.contains(event.target) && event.target.isConnected) {
	close();
}
```

Checking containment at `pointerdown`, or in the capture phase, would also work if you prefer that shape.

## Not blocking, but worth a look

`handleFavoriteStopClick` in `MapExperience.svelte` calls `flyTo(lat, lon, 20)` and then, 100ms later, `handleStopMarkerSelect` pushes state and the selection effect calls `flyTo(lat, lon, 16, ...)`. Google's `flyTo` is a synchronous `setZoom`+`setCenter`, so that reads as a zoom-20 flash snapping back to 16; on Leaflet it re-targets an in-flight animation. It also works against the contract documented two functions up, that the selection effect owns framing the map. The bare 100ms timer has no cleanup either. Letting the selection effect do the framing on its own would be tidier.

`handleFavoriteRouteClick` has no `mapProvider` guard, where `handleFavoriteStopClick` does — and the control is tappable before the map initializes, with `SearchPane.handleRouteClick` calling `mapProvider.clearAllPolylines()` straight away.

Routes with no `shortName` fail `isValidFavorite`, so `toggle` returns null and the star silently does nothing — no fill, no toast. Given `nullSafeShortName` exists in the codebase, null short names presumably do occur.

Fix the panel-close bug and this is good to go. The feature is solid and the test coverage is honest.

aaronbrethorst:
### Code review

Both earlier asks check out at this head. `pointer-events-auto` is off the `MapExperience` wrapper and on `FavoritesFloatingControl`'s own root, and `handleWindowClick` now bails on a disconnected target, so removing a row (or Clear All) no longer dismisses the panel. The fix itself is right.

Found 1 issue:

1. The regression test added for that fix passes whether or not the fix is there — `detached.dispatchEvent(...)` never reaches the `<svelte:window onclick>` listener at all (a disconnected node's propagation path is just itself, so window is not on it), and the following `window.dispatchEvent(...)` arrives with `event.target === window`, which is caught by the earlier `!(target instanceof Node)` guard on line 61. Delete `if (!target.isConnected) return;` and this test stays green. I confirmed the propagation behavior in jsdom directly. A faithful version is writable — the propagation path is computed at dispatch time, so a node that removes itself in its own listener still reaches window with `isConnected === false`:

   ```js
   const row = document.createElement('button');
   document.body.appendChild(row);
   row.addEventListener('click', () => row.remove());
   row.dispatchEvent(new MouseEvent('click', { bubbles: true }));
   ```

   The comment on line 133-134 ("Simulate that exact shape") also doesn't describe what the test does.

https://github.com/OneBusAway/wayfinder/blob/0475b0a13ed4f14ff0aaf4214f8d6550cead8712/src/components/favorites/__tests__/FavoritesFloatingControl.test.js#L126-L143

aaronbrethorst (changes requested):
Both of the blockers I raised earlier are genuinely fixed, and I checked each one against the head commit rather than taking the commit message for it:

- **`pointer-events-auto`** — the wrapper in `MapExperience.svelte` is now `w-fit self-end` and stays transparent, and `FavoritesFloatingControl` opts back in on its own root (`<div bind:this={rootEl} class="pointer-events-auto relative">`). The dead strip over the map on phones is gone, and the `md:absolute` still resolves against the full-screen overlay, so desktop placement is unchanged.
- **Panel closing on item removal** — the `!target.isConnected` early return is exactly the right guard, and it covers Clear All as well as the per-row ✕.

The unrelated cleanups that came with it are good calls too: dropping the `flyTo` + `setTimeout(…, 100)` from `handleFavoriteStopClick` in favor of letting the selection effect frame the map is the correct division of labor, and letting routes without a `shortName` be favorited (falling back to `removeAgencyPrefix(id)`) fixes a silent no-op star.

I ran the full suite, lint, and build on `0475b0a` locally — all green.

One thing I do want fixed before this lands.

## The new regression test doesn't exercise the fix

`src/components/favorites/__tests__/FavoritesFloatingControl.test.js:127-143` — "stays open when the clicked node was detached by its own handler" passes whether or not the fix is present. I deleted the `if (!target.isConnected) return;` line from the component and re-ran the file: **7 passed, 0 failed.**

Two reasons it can't fail:

1. `detached.dispatchEvent(...)` on an already-removed node never reaches `window`. A disconnected node's propagation path is just itself, so the `<svelte:window onclick>` listener isn't called at all.
2. The follow-up `window.dispatchEvent(...)` arrives with `event.target === window`, which exits one line earlier at the `!(target instanceof Node)` guard — never reaching the `isConnected` check.

My note on the last review said tests couldn't catch this. That was too pessimistic, and I want to correct it: the part that's untestable is the *Svelte microtask flush*, not the propagation. An event's path is computed at dispatch time, so a node that removes itself **inside its own listener** still reaches `window`, with `isConnected === false` and `contains() === false` — the exact state the guard exists for. I verified that in this repo's own jsdom:

```js
btn.addEventListener('click', () => btn.remove());
window.addEventListener('click', (e) => { /* e.target.isConnected === false */ });
btn.dispatchEvent(new MouseEvent('click', { bubbles: true }));
```

So the test wants roughly this shape instead — a node inside the panel that detaches itself in its own handler:

```js
it('stays open when the clicked node was detached by its own handler', async () => {
	render(FavoritesFloatingControl);
	await user.click(screen.getByRole('button', { name: 'Open favorites' }));
	expect(screen.getByRole('dialog')).toBeInTheDocument();

	const row = document.createElement('button');
	screen.getByRole('dialog').appendChild(row);
	row.addEventListener('click', () => row.remove());

	row.dispatchEvent(new MouseEvent('click', { bubbles: true }));

	expect(screen.getByRole('dialog')).toBeInTheDocument();
});
```

That one goes red without the guard and green with it. Worth confirming that yourself by deleting the line — a guard whose test can't fail is worse than no test, because the next person to touch `handleWindowClick` gets a green suite telling them it's safe.

## Smaller, take it or leave it

`FavoritesList.svelte:113` — the remove ✕ is `opacity-0` with `group-hover:opacity-100` / `focus-visible:opacity-100`. Touch devices have no hover, so on phones — where this panel mostly lives — the ✕ is invisible, though still tappable if you happen to hit it. Clear All keeps the feature reachable, so it's not broken, just undiscoverable. Making it visible below `md` (or always) would fix it.

## Heads up on ordering with #589

This PR's new `src/components/favorites/*` components import from `@fortawesome/*`, and #589 removes those packages from `package.json` outright. Whichever of the two lands second won't just conflict textually — it'll fail to build. Since both are yours: easiest is to land this one first, then rebase #589 and convert the three favorites files to Lucide as part of it. If you'd rather do it the other way, say so and I'll merge #589 first, and this one picks up the conversion instead.

Happy to re-review as soon as the test is tightened — everything else here is ready.

tarunsinghofficial:
Ready for review again @aaronbrethorst :)

aaronbrethorst:
### Code review

The blocker from the last round is genuinely fixed. I re-ran the file at this head with `if (!target.isConnected) return;` deleted from `handleWindowClick`: **1 failed | 6 passed**, and the failure is exactly `stays open when a node inside the panel detaches itself in its own handler` at the final `expect(screen.getByRole('dialog')).toBeInTheDocument()`. With the guard restored, 7 passed. The test now exercises the guard. Full suite at this head: 1843 passed, Prettier and ESLint clean.

The touch fix also checks out — I compiled the new class string through this repo's Tailwind 3.4.15 and confirmed the emitted rules: base `.opacity-100` comes before `@media(hover:hover){ .[@media(hover:hover)]:opacity-0 }`, so the ✕ is always visible where `hover: none`, and `.focus-visible\:opacity-100:focus-visible` (0,2,0) still outranks the media-scoped `.[@media(hover:hover)]:opacity-0` (0,1,0), so keyboard reveal is unaffected on desktop.

Found 1 issue:

1. Dismissing the panel by clicking outside steals focus away from whatever the user just clicked. `handleWindowClick` calls `close()` with no argument, so `restoreFocus` defaults to `true` and `tick().then(() => toggleBtn?.focus())` runs a microtask after the outside element has already taken focus. I verified this in this repo's jsdom: with the panel closed, clicking an `<input>` leaves it focused; with the panel open, the same click ends with `document.activeElement` on the favorites toggle (same result for an outside `<button>`). On the map page the search field sits right next to this control, so "tap the search box while the panel is open" both closes the panel and yanks the caret to the star. The other two call sites already make this distinction deliberately (`handleStopClick` / `handleRouteClick` pass `restoreFocus: false`), and Escape correctly keeps it `true`. Note the fix isn't just flipping it to `false` — clicking a non-focusable area like the map does need the restore, or focus drops to `<body>` when the panel unmounts; the usual pattern is to skip the restore only when the outside click landed on something focusable.

https://github.com/OneBusAway/wayfinder/blob/9f6a7214816cbf550bb36653e591577050428109/src/components/favorites/FavoritesFloatingControl.svelte#L69-L73

aaronbrethorst (approved):
The test is fixed, and I checked it the way I asked you to rather than reading it and nodding: the new `stays open when a node inside the panel detaches itself in its own handler` appends a button *inside* the dialog, has it remove itself in its own listener, and dispatches a bubbling click. That's the real shape — the event path is fixed at dispatch time, so it reaches `<svelte:window>` with `isConnected === false` and `contains()` false, which is exactly the state the guard exists for. Delete `if (!target.isConnected) return;` and that test goes red. That's a guard whose test can actually fail, which is what was missing last round.

The comment you left above it explaining *why* the propagation works that way is a genuinely good addition — that's the part a future reader would otherwise get wrong.

You also picked up the touch-device ✕ note, and the `@media (hover: hover)` approach is the right one: base `opacity-100` with the hover-only hiding scoped behind the media query means it's permanently visible where there's no hover, and `focus-visible:opacity-100` still wins on desktop at higher specificity. Keyboard reveal isn't regressed.

Full suite, lint, and build are green at `9f6a721`.

Merging this. One thing I found while reading it that I'm not holding the PR for — I've filed it as #604:

`handleWindowClick` calls `close()` bare, so `restoreFocus` defaults to `true` and `tick().then(() => toggleBtn?.focus())` fires after the clicked element has already taken focus. Open the panel, click the adjacent search field, and the caret gets pulled out of the field onto the star. `handleStopClick`/`handleRouteClick` already pass `restoreFocus: false` deliberately; this call site just wasn't distinguished. It's not a one-word fix — clicking the non-focusable map area genuinely does need the restore, or focus drops to `<body>` — so it's worth its own small change rather than a rushed one here.

Thanks for sticking with this through four rounds. The store was solid from the first review and the parts that needed work were all at the edges; that's a good place for a feature this size to end up.

Note on ordering: this landing means #589 now needs a rebase, and the three `src/components/favorites/*` files need converting to Lucide as part of it, since #589 drops `@fortawesome/*` from `package.json`. I'll say the same over there.

PR #586, opened 2026-08-02
fix: hexToRgb() to support 8-digit hex

### LINKED ISSUE #529: colorUtils: silent fallbacks + missing contrast regression test + validate-env 8-digit hex gap

antriksh-9:
### Summary
This PR is a follow-up to #584.
It addresses the third follow-up item from #529 by resolving the `validate-env / hexToRgb` 8-digit hex mismatch. 

### Changes
- Updated `hexToRgb()` method to support 8-digit hex colors (#RRGGBBAA).
- Added condition to ignore the alpha channel in 8-digit hex colors while parsing before regex 
- Added tests to verify the 8-digit hex parsing and to verify that `darkenColor` correctly accepts it. 

Closes: #529 

## Summary by CodeRabbit

* **New Features**
  * Added support for 8-digit hexadecimal color values.
  * Alpha channels are ignored when converting or darkening colors.
  * Support includes uppercase and lowercase hexadecimal formats.

* **Bug Fixes**
  * Color palettes now consistently output normalized six-digit hexadecimal values.

* **Tests**
  * Added coverage for 8-digit color parsing and darkening behavior.

### DIFF
--- a/src/lib/colorUtils.js
+++ b/src/lib/colorUtils.js
@@ -7,7 +7,7 @@ import { COLORS } from './colors.js';
 
 /**
  * Converts a hex color string to RGB object
- * Supports both 3-digit (#fff) and 6-digit (#ffffff) hex formats
+ * Supports 3-digit (#fff), 6-digit (#ffffff) and 8-digit (#ffffffff) hex formats
  * @param {string} hex - Hex color string (with or without #)
  * @returns {{r: number, g: number, b: number} | null} RGB object or null if invalid
  */
@@ -21,6 +21,11 @@ export function hexToRgb(hex) {
 			.map((c) => c + c)
 			.join('');
 	}
+	// ignore alpha channel in 8-digit hex
+	if (hex.length === 8) {
+		hex = hex.slice(0, 6);
+	}
+
 	const result = /^([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
 	return result
 		? {
--- a/src/lib/colorUtils.js
+++ b/src/lib/colorUtils.js
@@ -108,7 +108,7 @@ export function generatePalette(baseHex, fallbackHex = '#486621') {
 	// 600 is slightly darker, 900 is very dark
 	const darkWeights = { 600: 0.15, 700: 0.3, 800: 0.45, 900: 0.6 };
 
-	const palette = { 500: baseHex };
+	const palette = { 500: rgbToHex(base.r, base.g, base.b) };
 
 	for (const [shade, weight] of Object.entries(lightWeights)) {
 		const mixed = mixColors(base, white, weight);
--- a/src/tests/lib/colorUtils.test.js
+++ b/src/tests/lib/colorUtils.test.js
@@ -220,7 +220,7 @@ describe('colorUtils', () => {
 
 		test('works with shorthand hex', () => {
 			const palette = generatePalette('#f00');
-			expect(palette['500']).toBe('#f00');
+			expect(palette['500']).toBe('#ff0000');
 			expect(Object.keys(palette).length).toBe(10);
 		});
 
--- a/src/tests/lib/colorUtils.test.js
+++ b/src/tests/lib/colorUtils.test.js
@@ -43,6 +43,7 @@ describe('colorUtils', () => {
 			expect(hexToRgb('#AABBCC')).toEqual({ r: 170, g: 187, b: 204 });
 			expect(hexToRgb('#aAbBcC')).toEqual({ r: 170, g: 187, b: 204 });
 			expect(hexToRgb('#ABC')).toEqual({ r: 170, g: 187, b: 204 });
+			expect(hexToRgb('#486621FF')).toEqual({ r: 72, g: 102, b: 33 });
 		});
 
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
Thanks for picking this up — this closes the last open item from #529, and the fix is in exactly the right place. Slicing after the 3-digit expansion means a short hex can never be mistaken for an 8-digit one, the `{r,g,b} | null` return shape is unchanged so no caller needs touching, and aligning `hexToRgb` with the `#RGB | #RRGGBB | #RRGGBBAA` set that `scripts/validate-env.js` already accepts is the right call. I checked the `darkenColor('#486621ff', 0.15)` → `#3d571c` expectation arithmetically and it holds.

One thing to fix before this lands.

## The palette still leaks the alpha into shade 500

`generatePalette` builds every shade except 500 by re-serializing through `rgbToHex`, but 500 passes the caller's raw string straight through:

https://github.com/OneBusAway/wayfinder/blob/afcc6a1a35c14ab8bbfa8fc3b9f2c0a5f7ba4e70/src/lib/colorUtils.js#L105-L116

Before this PR that was harmless: an 8-digit `COLOR_BRAND_ACCENT` made `hexToRgb` return null, so the whole palette fell back and 500 was never an 8-digit value. This PR makes that path reachable — and `scripts/validate-env.js:110` explicitly blesses `#RRGGBBAA`, so it's a configuration we tell deployers is valid. `tailwind.config.js:32` feeds `process.env.COLOR_BRAND_ACCENT` directly into `generatePalette`, so with something like `COLOR_BRAND_ACCENT="#48662180"` you get `primary-500: #48662180` — translucent — sitting between an opaque `primary-400` and an opaque `primary-600`. Every `bg-primary-500` surface in the app renders see-through while its neighbors don't.

The fix is one line:

```js
const palette = { 500: rgbToHex(base.r, base.g, base.b) };
```

That keeps 500 identical for every input that works today and normalizes the 8-digit case the same way the other nine shades already are.

## Optional, not blocking

The alpha nibbles are dropped without being validated, so `hexToRgb('#486621zz')` now returns `{r:72,g:102,b:33}` while `hexToRgb('#gg0000')` correctly returns null. If you want the validation to be uniform, extending the regex and dropping the slice does it in one move:

```js
const result = /^([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})?$/i.exec(hex);
```

Low practical impact given `validate-env` screens env values with its own stricter regex, so take it or leave it.

Happy to re-review as soon as the 500 line is fixed — everything else here is good to go.

aaronbrethorst (approved):
The one-line fix is exactly right, and the test at `colorUtils.test.js` moving from `'#f00'` to `'#ff0000'` is the correct companion change — shade 500 is now normalized the same way the other nine already were, so a `COLOR_BRAND_ACCENT="#48662180"` can no longer produce a translucent `primary-500` sitting between opaque neighbors.

I re-checked the callers: all seven `hexToRgb` call sites live inside `colorUtils.js` and expect the unchanged `{r,g,b} | null` shape, so nothing downstream shifts. I also confirmed this legitimately closes #529 — #584 handled the silent-fallback and contrast-test items, and this is the third and last one.

You left the alpha nibbles unvalidated rather than folding them into the regex, which is the call I said was yours to make. Fine by me.

Thanks for the clean follow-through.

PR #591, opened 2026-08-15
fix(accordion): clear stale selection on item removal

### LINKED ISSUE #590: SingleSelectAccordion leaves stale activeItem/activeData when the active item unmounts
`SingleSelectAccordion` keeps `activeItem` and `activeData` in module-local stores that are only ever written from `activate()`, which runs on click:

https://github.com/OneBusAway/wayfinder/blob/develop/src/components/containers/SingleSelectAccordion.svelte#L26-L38

Nothing clears them when the currently-active `AccordionItem` unmounts. Each item registers under a `crypto.randomUUID()` generated in its own instance, so once that instance is destroyed the id in `activeItem` no longer matches anything mounted — but the `$effect` that calls `handleAccordionSelectionChanged` only re-runs when the store values *change*, and an unmount doesn't change them. So the callback never fires.

In `StopPane`, that callback drives the map:

```js
function handleAccordionSelectionChanged(event) {
	const data = event.activeData;
	const show = !!data;
	if (tripSelected) tripSelected({ detail: data });
	if (handleUpdateRouteMap) handleUpdateRouteMap({ detail: { show } });
	...
}
```

### Repro

1. Open a stop with real-time arrivals.
2. Expand an arrival — its trip polyline and vehicle marker draw on the map.
3. Wait for that vehicle to arrive.
4. On the next 30s poll the row is filtered out and fades away.

The list now has nothing expanded, but the departed trip's polyline and vehicle marker are still drawn, and stay drawn until you expand something else.

### Why now

The mechanism predates #438 — before it, an unkeyed `{#each}` reused component instances positionally, so the expanded row survived but ended up rendering a *different* trip's data while `activeData` still held the old one. Different symptom, same root cause.

What #438 changed is the frequency. The `{#each}` is now keyed and departed arrivals are filtered client-side every poll, so the expanded row genuinely unmounts — and it does so for precisely the arrival the rider was watching, on a predictable 30-second cadence. It went from a rare glitch to a routine one.

### Suggested fix

Clear the stores when the active item unmounts, in `registerItem`:

```js
registerItem: (id) => {
	$effect(() => () => {
		if (get(activeItem) === id) {
			activeItem.set(null);
			activeData.set(null);
		}
	});
	...
}
```

or have the consumer reconcile — e.g. `StopPane` clearing the selection when the active key is no longer present in the arrivals list. The first is preferable since it fixes every consumer of the accordion at once.

Worth adding a test that unmounts the active item and asserts `handleAccordionSelectionChanged` fires with `activeData: null`.

Found while reviewing #438; not a regression introduced by it, so it merged without this.

soumajitgh:
## Summary

- Clear accordion selection state when its active item unmounts.
- Add a regression test covering removal of an expanded item.

## Root cause

`activeItem` and `activeData` were only updated through clicks. When a keyed
arrival row was removed after a real-time refresh, its selection state remained
and the map continued to show the departed trip.

Closes #590

## Validation

- `npm test -- --run src/components/containers/__tests__/SingleSelectAccordion.test.js`
- `npm run lint:eslint -- src/components/containers/SingleSelectAccordion.svelte src/components/containers/__tests__/SingleSelectAccordion.test.js src/components/containers/__tests__/SingleSelectAccordionFixture.svelte`

## Summary by CodeRabbit

* **Bug Fixes**
  * Cleared the active accordion selection when the selected item is removed, preventing stale selection data and callbacks.

* **Tests**
  * Added coverage for removing the currently selected accordion item and verifying that selection state resets correctly.


soumajitgh:
@coderabbitai review

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice fix. The teardown lands in exactly the right place: because `registerItem` is called at the top level of `AccordionItem`'s `<script>`, the `$effect` is created under the branch effect that instantiated the item, so its lifetime tracks the item and the teardown fires precisely on removal. The `get(activeItem) === id` guard is the important detail — it keeps a departing item from clobbering a selection that has already moved elsewhere.

I verified this locally against the head commit rather than relying on CI (this PR's workflow runs are still sitting in `action_required`, since it's your first PR here):

- `npx vitest run` — 1625 tests across 90 files, all green
- `npm run lint` — clean
- `npm run build` — clean
- Reverted the teardown block and re-ran the new test: it fails on the `{activeItem: null, activeData: null}` assertion. That's a real red-to-green regression test, not a tautology.

The `Element.prototype.animate` stub covers just the surface Svelte's slide transition touches and is restored in `afterEach`, so it doesn't leak into other files. The fixture component is a clean way to drive mount/unmount without reaching into internals.

One thing worth a follow-up, not a blocker for this PR: `StopPane.handleAccordionSelectionChanged` calls `analytics.reportArrivalClicked` unconditionally, so now that the callback also fires when a selected arrival is filtered out, a departure will get logged as an arrival click. That handler already misreports collapse-by-click the same way, so it's a pre-existing problem this change slightly widens — I'll take it separately.

Thanks for the clear root-cause writeup in the description.

PR #592, opened 2026-08-20
fix(a11y): add trip planner combobox pattern

### LINKED ISSUE #527: a11y: add full combobox/listbox ARIA pattern to trip-planner autocomplete
Follow-up to #517 / PR #525.

PR #525 fixed the invalid HTML (buttons are now wrapped in `<li>` under the `<ul>`). The remaining optional enhancement from #517 is to implement the full WAI-ARIA combobox pattern in `src/components/trip-planner/TripPlanSearchField.svelte`:

- `role="combobox"`, `aria-expanded`, `aria-controls`, and `aria-activedescendant` on the input
- `role="listbox"` on the `<ul>` and `role="option"` on each result
- Arrow-key navigation between options (active-descendant focus management)

This gives screen readers proper announcement of the options list and result count.

Ref: WCAG 1.3.1, WAI-ARIA Authoring Practices combobox pattern.

soumajitgh:
## Summary
- add combobox and listbox semantics to trip-planner autocomplete
- support active-descendant Arrow-key navigation, Enter selection, and Escape dismissal
- cover ARIA relationships and keyboard behavior with component tests

## Testing
- npm test -- --run src/components/trip-planner/__tests__/TripPlanSearchField.test.js

Closes #527

## Summary by CodeRabbit

- **New Features**
  - Added keyboard navigation for trip search suggestions using arrow keys.
  - Press Enter to select a highlighted suggestion.
  - Press Escape to dismiss autocomplete results or close the trip planner when no suggestions are open.
  - Improved accessibility with combobox and listbox semantics, active-option tracking, and clearer screen reader relationships.

- **Bug Fixes**
  - Prevented outdated autocomplete responses from replacing current results.
  - Preserved independent loading states for each location field.
  - Prevented handled keyboard events from triggering shortcuts.

### DIFF
--- a/src/components/trip-planner/TripPlanSearchField.svelte
+++ b/src/components/trip-planner/TripPlanSearchField.svelte
@@ -11,6 +11,7 @@
 	 * @property {(value: string) => void} onInput
 	 * @property {() => void} onClear
 	 * @property {any} onSelect
+	 * @property {() => void} [onDismiss]
 	 */
 
 	/** @type {Props} */
@@ -21,10 +22,16 @@
 		isLoading = false,
 		onInput,
 		onClear,
-		onSelect
+		onSelect,
+		onDismiss = () => {}
 	} = $props();
 
+	let activeIndex = $state(-1);
+	let listboxId = $derived(`${inputId}-listbox`);
+	let hasResults = $derived(!isLoading && Array.isArray(results) && results.length > 0);
+
 	function handleInput(event) {
+		activeIndex = -1;
 		onInput(event.target.value);
 	}
 
@@ -33,7 +40,37 @@
 	}
 
 	function handleSelect(result) {
+		activeIndex = -1;
 		onSelect(result);
+	}
+
+	function optionId(index) {
+		return `${listboxId}-option-${index}`;
+	}
+
+	function handleKeydown(event) {
+		if (!hasResults) return;
+
+		switch (event.key) {
+			case 'ArrowDown':
+				event.preventDefault();
+				activeIndex = activeIndex < results.length - 1 ? activeIndex + 1 : 0;
+				break;
+			case 'ArrowUp':
+				event.preventDefault();
+				activeIndex = activeIndex > 0 ? activeIndex - 1 : results.length - 1;
+				break;
+			case 'Enter':
+				if (activeIndex < 0) return;
+				event.preventDefault();
+				handleSelect(results[activeIndex]);
+				break;
+			case 'Escape':
+				event.preventDefault();
+				activeIndex = -1;
+				onDismiss();
+				break;
+		}
 	}
 </script>
 
@@ -43,6 +80,13 @@
 		type="text"
 		bind:value={place}
 		oninput={handleInput}
+		onkeydown={handleKeydown}
+		role="combobox"
+		aria-autocomplete="list"
+		aria-expanded={hasResults}
+		aria-haspopup="listbox"
+		aria-controls={hasResults ? listboxId : undefined}
+		aria-activedescendant={hasResults && activeIndex >= 0 ? optionId(activeIndex) : undefined}
 		placeholder="{$t('trip-planner.search_for_a_place')}..."
 		class="block w-full rounded-md border-gray-300 pr-10 text-sm text-black shadow-sm focus:border-blue-500 focus:ring-blue-500"
 	/>
@@ -58,18 +102,31 @@
 	{/if}
 	{#if isLoading}
 		<p
+			role="status"
 			class="absolute z-10 mt-1 w-full rounded-md border border-gray-300 bg-white px-4 py-2 text-gray-500 shadow-lg"
 		>
 			{$t('trip-planner.loading')}...
 		</p>
 	{:else if results && results.length > 0}
 		<ul
+			id={listboxId}
+			role="listbox"
 			class="absolute z-10 mt-1 max-h-60 w-full overflow-auto rounded-md border border-gray-300 bg-white shadow-lg"
 		>
-			{#each results as result}
-				<li>
+			{#each results as result, index}
+				<li role="presentation">
 					<button
-						class="flex w-full cursor-pointer items-center px-4 py-2 text-left hover:bg-gray-100 dark:text-black"
+						id={optionId(index)}
+						type="button"
+						role="option"
+						tabindex="-1"
+						aria-selected={activeIndex === index}
+						aria-posinset={index + 1}
+						aria-setsize={results.length}
+						class="flex w-full cursor-pointer items-center px-4 py-2 text-left hover:bg-gray-100 dark:text-black {activeIndex ===
+						index
+							? 'bg-gray-100'
+							: ''}"
 						onclick={() => handleSelect(result)}
 					>
 						<FontAwesomeIcon icon={faMapMarkerAlt} class="mr-2 text-gray-400  " />
--- a/src/components/trip-planner/TripPlanSearchField.svelte
+++ b/src/components/trip-planner/TripPlanSearchField.svelte
@@ -49,6 +49,13 @@
 	}
 
 	function handleKeydown(event) {
+		if (event.key === 'Escape') {
+			event.preventDefault();
+			activeIndex = -1;
+			onDismiss();
+			return;
+		}
+
 		if (!hasResults) return;
 
 		switch (event.key) {
@@ -64,11 +71,6 @@
 				if (activeIndex < 0) return;
 				event.preventDefault();
 				handleSelect(results[activeIndex]);
-				break;
-			case 'Escape':
-				event.preventDefault();
-				activeIndex = -1;
-				onDismiss();
 				break;
 		}
 	}
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -73,8 +73,11 @@
 	const fetchLocationResults = debounce(async (query, isFrom, requestId) => {
 		if (!isCurrentAutocompleteRequest(isFrom, requestId)) return;
 
-		isLoadingFrom = isFrom;
-		isLoadingTo = !isFrom;
+		if (isFrom) {
+			isLoadingFrom = true;
+		} else {
+			isLoadingTo = true;
+		}
 
 		try {
 			const results = await fetchAutocompleteResults(query);
@@ -88,8 +91,11 @@
 			}
 		} finally {
 			if (isCurrentAutocompleteRequest(isFrom, requestId)) {
-				isLoadingFrom = false;
-				isLoadingTo = false;
+				if (isFrom) {
+					isLoadingFrom = false;
+				} else {
+					isLoadingTo = false;
+				}
 			}
 		}
 	}, 500);
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -161,6 +161,14 @@
 		clearTripUrl();
 	}
 
+	function dismissSearchResults(isFrom) {
+		if (isFrom) {
+			fromResults = [];
+		} else {
+			toResults = [];
+		}
+	}
+
 	function swapLocations() {
 		const result = swapTripLocations({
 			fromPlace,
@@ -443,6 +451,7 @@
 						onInput={(query) => handleSearchInput(query, true)}
 						onClear={() => clearInput(true)}
 						onSelect={(location) => selectLocation(location, true)}
+						onDismiss={() => dismissSearchResults(true)}
 					/>
 				</div>
 			</div>
@@ -471,6 +480,7 @@
 						onInput={(query) => handleSearchInput(query, false)}
 						onClear={() => clearInput(false)}
 						onSelect={(location) => selectLocation(location, false)}
+						onDismiss={() => dismissSearchResults(false)}
 					/>
 				</div>
 			</div>
--- a/src/components/trip-planner/TripPlan.svelte
+++ b/src/components/trip-planner/TripPlan.svelte
@@ -41,6 +41,8 @@
 	let loading = $state(false);
 	let fromRequestId = 0;
 	let toRequestId = 0;
+	let fromAutocompleteRequestId = 0;
+	let toAutocompleteRequestId = 0;
 	// Recent trips stay behind a compact control so the plan form doesn't grow
 	// a tall history list under From/To (see #577).
 	let showRecentTrips = $state(false);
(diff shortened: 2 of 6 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. Escape inside the search field also closes the whole mobile trip-plan sheet, so the new dismiss behavior can never be observed there. `TripPlanModal`'s close button uses `use:keybinding={{ code: 'Escape' }}`, which registers a **window**-level keydown listener that does not check `event.defaultPrevented`. `handleKeydown` calls `event.preventDefault()` but not `event.stopPropagation()`, so both handlers run. Verified by rendering `TripPlanModal` with `showForm={true}` and firing Escape on `#from-location-input`: `closePane` is called once. Per the APG combobox pattern, Escape should close the popup without closing the containing dialog — the handler needs to stop propagation (or otherwise signal "handled") when it actually dismisses an open popup.

https://github.com/OneBusAway/wayfinder/blob/460442ba1ac506d9e7771550621a189974752541/src/components/trip-planner/TripPlanSearchField.svelte#L50-L58

2. The new `does not reopen results when a pending response resolves after Escape` regression test passes regardless of the behavior it is meant to protect. After `resolveSuggestions()` it only awaits one microtask plus `tick()`, which is not enough to drain the `fetch` -> `response.json()` -> `fetchAutocompleteResults` -> `fetchLocationResults` await chain, so the `queryByRole('listbox')` assertion runs before `fromResults` would ever be assigned. Verified by stripping every `isCurrentAutocompleteRequest` guard out of `TripPlan.svelte`: the test still passes. Adding a proper flush (e.g. `await vi.advanceTimersByTimeAsync(0)` plus a few `await Promise.resolve()` iterations) makes it fail against the unguarded source and pass against this PR's source — so the underlying fix is correct, but the test as written does not lock it in.

https://github.com/OneBusAway/wayfinder/blob/460442ba1ac506d9e7771550621a189974752541/src/components/trip-planner/__tests__/TripPlan.test.js#L150-L156

src/components/trip-planner/TripPlanSearchField.svelte
```diff
@@ -33,8 +40,40 @@
...
+		activeIndex = -1;
 		onSelect(result);
 	}
+
+	function optionId(index) {
+		return `${listboxId}-option-${index}`;
+	}
+
+	function handleKeydown(event) {
+		if (event.key === 'Escape') {
```
aaronbrethorst, line 52: This needs `event.stopPropagation()` as well as `preventDefault()`.

`TripPlanModal.svelte:234` has `use:keybinding={{ code: 'Escape' }}` on its close button, and `keybinding` registers on **`window`** and doesn't check `e.defaultPrevented`:

```js
handler = (e) => {
    if (... currentParams.code != e.code) return;
    e.preventDefault();
    currentParams.callback ? currentParams.callback() : node.click();
};
window.addEventListener('keydown', handler);
```

So Escape here dismisses the listbox *and* clicks the modal close button — on the mobile plan sheet the whole sheet goes away. APG wants Escape to close the popup without closing the containing dialog.

Alternative worth considering: make `keybinding` return early on `e.defaultPrevented`, which fixes this for every consumer rather than just this one.

aaronbrethorst (changes requested):
The ARIA work here is genuinely good, and it's the part I expected to find problems in. `role="combobox"` on the input with `aria-controls`/`aria-activedescendant` pointing at a `role="listbox"` of `role="option"` children, `aria-posinset`/`aria-setsize` on each option, arrow keys moving active-descendant without moving DOM focus — that's the APG pattern implemented properly, not approximated. The `aria-expanded` binding to `hasResults` is right, and swapping the old `getByRole('list')` assertions for `getByRole('listbox')` in the tests was the correct follow-through.

The request-invalidation fix in `TripPlan.svelte` is also a real fix — a late autocomplete response genuinely could reopen a dismissed list before this.

Two things to sort out before it lands.

## Escape closes the entire trip-plan sheet, so the new dismissal is unobservable where it matters

`handleKeydown` calls `event.preventDefault()` on Escape but not `event.stopPropagation()`. Meanwhile `TripPlanModal.svelte:234` puts `use:keybinding={{ code: 'Escape' }}` on its close button, and `keybinding` (`src/lib/keybinding.js`) registers its handler on **`window`** with no `defaultPrevented` check:

```js
handler = (e) => {
    if (... currentParams.code != e.code) return;
    e.preventDefault();
    currentParams.callback ? currentParams.callback() : node.click();
};
window.addEventListener('keydown', handler);
```

So Escape in the search field runs your handler *and* clicks the modal's close button. On the mobile plan sheet — where this field actually lives, via `MapExperience.svelte` rendering `TripPlanModal` with `showForm` — pressing Escape to dismiss the suggestion list dismisses the whole sheet instead.

APG is explicit that Escape should close the popup without closing the containing dialog. `event.stopPropagation()` alongside the existing `preventDefault()` is the smallest fix; teaching `keybinding` to bail on `e.defaultPrevented` would fix it for every future consumer, which I'd slightly prefer but won't insist on.

This isn't a regression — Escape closed the sheet before this PR too — but it does mean the headline feature doesn't work in the place people will try it.

## The async regression test passes without the fix

`does not reopen results when a pending response resolves after Escape` only awaits `await Promise.resolve(); await tick();` after `resolveSuggestions()`. That isn't enough to drain `fetch → response.json() → fetchAutocompleteResults → fetchLocationResults`, so the final `queryByRole('listbox')` assertion runs before `fromResults` would have been assigned either way. Strip every `isCurrentAutocompleteRequest` guard out of `TripPlan.svelte` and it still goes green.

Worth confirming that yourself by deleting the guards — a test that can't fail is worse than no test, because the next person to touch that code gets a green suite telling them it's safe. The intermediate `queryByRole('status')` assertion *is* meaningful; it's specifically the post-resolve one that needs a real flush.

## Not blocking

- No `scrollIntoView` on the active option. The listbox is `max-h-60 overflow-auto` and `aria-activedescendant` doesn't scroll the container — but Google Places returns at most 5 suggestions (~200px in a 240px box), so it isn't reachable today. Worth remembering if that cap ever rises.
- APG suggests Alt+Down (or Down) reopens a popup closed with Escape; right now the only way back is editing the text. Optional in the spec.
- During `isLoading` the `role="status"` popup is visible while `aria-expanded` is `false`, since `hasResults` excludes the loading state. Cosmetic.

Fix the Escape propagation and tighten that one test and I'll merge this — the accessibility work underneath is solid.

soumajitgh:
Addressed the requested changes in c27d3ca:

- Global keybinding handlers now skip events already handled by a child control, so Escape dismisses autocomplete without closing the mobile trip-plan sheet.
- Tightened the delayed-autocomplete regression test to drain the full async chain. I mutation-checked it by removing the request guards locally; the test then fails as expected.

Verified with Prettier, ESLint, and the focused Vitest suite (44 tests).

aaronbrethorst (changes requested):
Both of my blocking points are genuinely fixed, and you took the better of the two options on the first one. Bailing in `keybinding` on `e.defaultPrevented` rather than having the field call `stopPropagation()` is the right call — it fixes this for every current and future consumer instead of one component, and I checked the other four `use:keybinding` sites plus the window-level Escape handlers in `FavoritesFloatingControl` and `ServiceAlerts`; none of them get stranded by the change.

The async test fix is real too. I verified that stripping the post-`await` guard in `TripPlan.svelte` now makes `does not reopen results when a pending response resolves after Escape` fail — before this commit it passed vacuously either way. That's the behavior actually locked in now.

One new problem, introduced by the fix, and it's on the same line I commented on.

## Escape is now swallowed unconditionally in the From/To inputs

`handleKeydown` in `TripPlanSearchField.svelte` calls `event.preventDefault()` in the Escape branch *before* the `if (!hasResults) return;` guard — so every Escape from those inputs is marked handled, whether or not a popup is open. Combined with the new `defaultPrevented` bail in `keybinding.js`, the trip-plan sheet can no longer be closed with Escape while focus is in From or To at all.

| scenario | `closePane` before | `closePane` now |
| --- | --- | --- |
| Escape, no popup open | 1 | **0** |
| Escape, listbox open | 1 (the bug I filed) | 0 (correct) |

That's a regression against `develop`, where the field had no keydown handler and Escape always reached the close button. It also breaks the second half of the standard APG sequence — Escape to dismiss the list, Escape again to close the dialog. Note `hasResults` is `!isLoading && ...`, so this bites during loading too.

My comment scoped the "handled" signal to when Escape *actually dismisses an open popup*; the code signals it for every Escape. Gate the branch on a popup genuinely being open:

```js
const popupOpen = hasResults || isLoading;
if (event.key === 'Escape') {
    if (!popupOpen) return;          // let it bubble to the sheet's keybinding
    event.preventDefault();
    activeIndex = -1;
    onDismiss();
    return;
}
```

The two existing Escape tests only assert `onDismiss` fires, so nothing caught this — please add one asserting the sheet still closes when no popup is open.

## Smaller notes, take or leave

- The flush idiom (`await vi.advanceTimersByTimeAsync(0)` plus exactly four microtask turns) works today but depends on the current await depth of `fetch -> json() -> fetchAutocompleteResults -> fetchLocationResults`. Add one `await` to that chain and all three tests quietly go back to passing vacuously. It's the idiom I suggested, so this is my problem as much as yours — but a `waitFor`-style assertion would be more durable.
- `keybinding`'s JSDoc doesn't mention the new "ignores already-handled events" contract, which is now load-bearing for every consumer.

The three non-blocking items from last time (scrollIntoView, Alt+Down reopen, `role="status"` while `aria-expanded="false"`) are all still open. I flagged them as optional and I still mean that — no need to address them here.

The per-field loading commit is fine; I traced the invalidation paths and there's no reachable stuck-loading state. Fix the Escape gating and I'll merge this.

soumajitgh:
@aaronbrethorst Addressed the remaining Escape gating issue in f7b75d5. Escape now dismisses only an open/loading autocomplete popup; with no popup it reaches the sheet close binding. Added a modal regression test, and the full lint/test suite passes.

aaronbrethorst:
### Code review

Found 1 issue:

1. The guarded `finally` can leave a field's loading flag stuck on forever (regression vs `develop`). `fetchLocationResults` now only clears `isLoadingFrom`/`isLoadingTo` when the resolving request is still the current one — but `debounce` in `src/lib/utils.js` keeps a **single shared timer** for both fields, so a To keystroke cancels a pending From invocation. Repro: type in From and pause so the debounce fires (request A in flight, `isLoadingFrom = true`) → type one more character in From (A is now stale, follow-up scheduled) → type in To within 500ms (the From follow-up is dropped) → A resolves, the `finally` guard is false, and `isLoadingFrom` is never cleared. The From field then shows "Loading…" permanently, `aria-expanded` is pinned to `false`, and `hasResults` can never become true, so that combobox cannot reopen until the rider types in From again. I ran this scenario against both revisions: at this head SHA one loading indicator is still rendered after every request has resolved and timers are drained; on `develop` (unconditional `isLoadingFrom = false; isLoadingTo = false`) it is zero. The `catch` path has the same hole, and To is symmetric. Guarding only the *results* assignment on the request token, while always clearing this field's own loading flag, would fix it.

https://github.com/OneBusAway/wayfinder/blob/f7b75d5b548ef620f70fc53c7ea39b1988395c87/src/components/trip-planner/TripPlan.svelte#L90-L101

aaronbrethorst (changes requested):
The ARIA work here is right, and both of my earlier blocking points are genuinely closed. Taking the `e.defaultPrevented` bail in `keybinding.js` rather than `stopPropagation()` was the better of the two options I offered, and you updated the JSDoc to document the new contract. The `hasResults || isLoading` gating on the Escape branch is exactly the shape I asked for, and the new `TripPlanModal` test really does fail if that gating regresses. The combobox wiring itself checks out: `aria-controls` matches the listbox id, `aria-activedescendant` tracks the active option, and `posinset`/`setsize` are correct.

One thing needs fixing before this lands, and it's in the per-field loading change rather than the ARIA work.

## The From field can get stuck in "Loading..." forever

`debounce` in `src/lib/utils.js` closes over a **single** `timeout`, and `TripPlan.svelte` calls `debounce(...)` once for both fields. So a keystroke in To `clearTimeout`s a pending From invocation — the two fields share one timer.

That was harmless while the `finally` cleared both flags unconditionally, which is what `develop` does:

```js
} finally {
    isLoadingFrom = false;
    isLoadingTo = false;
}
```

At this head, the clear is guarded on `isCurrentAutocompleteRequest(isFrom, requestId)`. That opens a reachable gap:

1. Type in From. The debounce fires, `isLoadingFrom = true`, request A is in flight with `requestId` 1.
2. Type another character in From. `fromAutocompleteRequestId` becomes 2, and a follow-up invocation is scheduled.
3. Type in To within 500ms. That `clearTimeout`s the pending From follow-up, which is now dropped entirely.
4. Request A resolves. Its `requestId` is 1, but `fromAutocompleteRequestId` is 2, so the guard is false and `finally` skips the clear.

Nothing ever sets `isLoadingFrom` back to false. The field renders "Loading..." indefinitely, `aria-expanded` stays pinned to `false`, and `hasResults` can't become true, so that combobox can't reopen. It only recovers if the rider types in From again and that invocation survives to completion. Tabbing From to To inside a few hundred milliseconds of network latency is ordinary behavior, and more likely on a slow mobile connection.

This is a regression against `develop`, not a pre-existing issue.

## The fix

Smallest version: keep the token guard on the results assignment, where it's doing real work, but always clear this field's own flag:

```js
} finally {
    if (isFrom) {
        isLoadingFrom = false;
    } else {
        isLoadingTo = false;
    }
}
```

A stale response can then end the spinner slightly early if a newer request is already in flight, which is a brief flicker rather than a dead field.

Cleaner version, if you'd rather fix the cause: call `debounce()` once per field so each has its own timer. That also removes the cross-field cancellation entirely - right now typing in To silently kills a pending From lookup, which is on `develop` too but only becomes visible once the loading flags are per-field.

Either way, a test in the shape of "start a From request, supersede it, then type in To before the follow-up fires" would lock it down. The existing `keeps each field loading until its own request completes` test doesn't reach this path because it never drops an invocation.

Everything else is ready. Fix the loading clear and I'll take it.

soumajitgh:
Addressed the loading-state issue in c83edd2. From and To now use independent debounce timers, so typing in one field cannot cancel the other field’s queued follow-up request. Added a regression test covering the exact in-flight From → superseding From → To sequence and verifying all requests run and no loading indicator remains. Full suite passes (1,801 tests), along with Prettier and ESLint.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is fixed the right way. You went with separate debouncers, `fetchFromLocationResults` and `fetchToLocationResults`, so a keystroke in To can no longer cancel a pending From lookup. That was the cause, not only the symptom, so the guarded `finally` can stay as it is. I walked through each path that advances a field's token, and each one either clears that field's flag directly or schedules a follow-up that the other field can't cancel.

The new test is the one I wanted. It runs the From request, the superseding From keystroke, and then To inside the debounce window. Against the previous head it fails twice: `fetch` is called twice instead of three times, and the From status is still on screen at the end. The fake timers are also reset in `afterEach`, so it won't leak into the rest of the file.

The combobox wiring, the Escape gating, and the `defaultPrevented` contract in `keybinding.js` are all untouched since the last round. Thanks for sticking with this through several rounds. Merging.

PR #593, opened 2026-08-20
fix(analytics): handle missing event envelope

### LINKED ISSUE #333: Analytics: follow-up improvements from PR #332 review

soumajitgh:
## Summary
- default Plausible event envelopes to an empty object
- return the existing clear validation error for no-argument calls
- document the intentional disabled-before-validation behavior

## Testing
- npm test -- --run src/tests/lib/Insights/adapters/PlausibleAdapter.test.js src/tests/api/events.test.js
- npx prettier --check src/lib/Insights/adapters/PlausibleAdapter.js src/tests/lib/Insights/adapters/PlausibleAdapter.test.js

Closes #333

### DIFF
--- a/src/lib/Insights/adapters/PlausibleAdapter.js
+++ b/src/lib/Insights/adapters/PlausibleAdapter.js
@@ -25,7 +25,9 @@ export class PlausibleAdapter {
 		return `${this.env.PUBLIC_ANALYTICS_API_HOST}/api/event`;
 	}
 
-	async forwardEvent(envelope, requestContext) {
+	async forwardEvent(envelope = {}, requestContext) {
+		// Disabled analytics intentionally ignores malformed events so local development
+		// and tests without a complete analytics configuration remain quiet.
 		if (!this.isEnabled()) {
 			return { status: 'analytics disabled' };
 		}
--- a/src/tests/lib/Insights/adapters/PlausibleAdapter.test.js
+++ b/src/tests/lib/Insights/adapters/PlausibleAdapter.test.js
@@ -102,6 +102,12 @@ describe('PlausibleAdapter.forwardEvent', () => {
 		).rejects.toThrow('forwardEvent requires name and url');
 	});
 
+	it('throws a clear validation error when called without an envelope', async () => {
+		await expect(new PlausibleAdapter(fullEnv).forwardEvent()).rejects.toThrow(
+			'forwardEvent requires name and url'
+		);
+	});
+
 	it('POSTs to {apiHost}/api/event', async () => {
 		await new PlausibleAdapter(fullEnv).forwardEvent(envelope, ctx);
 		expect(global.fetch).toHaveBeenCalledWith(


aaronbrethorst (approved):
This does exactly what items 1 and 3 of #333 asked for, and it does them cleanly. I checked the behavior change rather than taking the description for it: with `envelope = {}`, `name` and `url` are still `undefined`, so the existing guard throws — the net effect is that a no-argument call now produces the domain error `forwardEvent requires name and url` instead of a raw destructuring `TypeError`. Both propagate identically through `/api/events`, so no status code moves. The new test genuinely fails on `develop`, where the message would be `Cannot destructure property 'name' of 'undefined'`.

The comment above the `isEnabled()` early return is the right call too — that ordering looked like a bug in the #332 review, and now it reads as deliberate.

I've opened #603 for three things that came up while reading this. None of them are yours and none block the merge:

- `= {}` is a default parameter, so it only fires on `undefined`. The malformed input `/api/events` can actually produce is `null` (a literal `null` body — `request.json()` returns it, and `JSON.parse` never returns `undefined`), which still throws. `envelope ?? {}` would close that.
- `UmamiAdapter.forwardEvent` has the same signature and body and didn't get the same treatment, so the two providers now diverge on malformed input.
- Validation errors surface as 500 rather than 400, which is pre-existing.

Merging. Thanks for picking up the follow-up items — those issues have a habit of sitting.

PR #594, opened 2026-08-20
WIP: Upgrade Tailwind V3 to V4

### LINKED ISSUE #353: Upgrade from Tailwind CSS v3 to v4
The project currently uses Tailwind CSS v3.4.15. Tailwind CSS v4 has been released with significant changes including a new CSS-first configuration approach, a new high-performance engine, and automatic content detection.

### Current setup
- `tailwindcss@3.4.15`
- `@tailwindcss/forms@0.5.9`
- `@tailwindcss/line-clamp@0.4.4`
- JS-based config in `tailwind.config.js`

### Migration considerations
- Tailwind v4 replaces `tailwind.config.js` with CSS-based `@theme` configuration
- The `@tailwindcss/line-clamp` plugin is no longer needed (line-clamp is built-in since v3.3, and the plugin is removed in v4)
- `@tailwindcss/forms` needs to be checked for v4 compatibility
- Flowbite (used for UI components) needs to support Tailwind v4
- Brand color CSS custom properties and `generatePalette()` in `src/lib/colorUtils.js` may need adjustments for the new theming system
- PostCSS config changes may be required

tarunsinghofficial:
## Summary

Migrates Wayfinder from Tailwind CSS v3.4 to v4, using the `@tailwindcss/vite` plugin and keeping the existing JS config via `@config` so env-var branding (`generatePalette()`, brand/surface colors) continues to work unchanged.

Closes #353

## What changed

**Build pipeline**
- Install `tailwindcss@4` + `@tailwindcss/vite`
- Wire Tailwind into `vite.config.js`; remove `postcss.config.js`, `autoprefixer`, and `@tailwindcss/line-clamp` (built into v4)
- Bump `@tailwindcss/forms` and `prettier-plugin-tailwindcss`

**CSS / config**
- Replace `@tailwind` directives with `@import "tailwindcss"` + `@config "../tailwind.config.js"`
- Add `@custom-variant dark` using `:is(.dark, .dark *)` so dark utilities beat light ones on the same node (fixes search pane in dark mode)
- Adapt `tailwind.config.js` for ESM plugin imports

**Codemod + fixes**
- Run `@tailwindcss/upgrade` for v4 utility renames (`bg-gradient-to-*` → `bg-linear-to-*`, `break-words` → `wrap-break-word`, etc.)
- Add `@reference` in `StopMarker.svelte` for `@apply` in component styles
- Add `maplibre-gl` as a direct dependency (peer of `@maplibre/maplibre-gl-leaflet`, dropped by `--legacy-peer-deps` installs)

## Flowbite

Spiked `flowbite-svelte@1.33` — needs Svelte 5.40+, Datepicker build/API churn, and component API updates across 6 files. **`flowbite-svelte@0.47` works with Tailwind v4** (build + UI verified), so it stays on 0.47 for this PR. A Flowbite 1.x upgrade can be a follow-up issue.

Components in use: Modal, Button, Tabs/TabItem, Datepicker, Radio, Checkbox, Input (6 files).

## Visual QA

Compared 8 dark-mode before/after screenshots (search pane, All Routes, route detail, map pins, etc.) — no visible regressions after the dark-mode specificity fix.

## Test plan

- [x] `npm run lint`
- [x] `npx vitest run` (103 files, 1794 tests)
- [x] `npm run build`

## Summary by CodeRabbit

* **Refactor**
  * Updated styling for compatibility with Tailwind CSS 4.
  * Improved text wrapping for route names and badges.
  * Refined light and dark theme colors across route, stop, search, and itinerary views.
  * Updated menu, header, toast, accordion, and skip-link positioning.

* **Bug Fixes**
  * Improved dark-mode backgrounds in search panels.
  * Standardized gradient and layout styling across loading, error, and itinerary screens.

* **Documentation**
  * Updated arrival and departure redesign specifications to reflect revised text wrapping.


aaronbrethorst:
### Code review

Found 3 issues:

1. `bg-opacity-*` / `ring-opacity-*` were removed in Tailwind v4 and three of them survive the codemod on lines this PR touched, so they now emit no CSS at all. Verified by compiling `tailwindcss@4.3.3` with an equivalent `@config` legacy config: neither class produces a rule. The loading overlay below loses its 80% transparency and becomes a solid `bg-neutral-800` sheet over the map; `bg-brand-accent bg-opacity-80` in `src/components/routes/RouteModal.svelte#L80` becomes fully opaque; `focus:ring-opacity-50` in `src/components/surveys/SurveyLauncher.svelte#L43` is a no-op. These need the v4 opacity-modifier form (`bg-neutral-800/80`, `focus:ring-green-400/50`). Note that `src/components/__tests__/LoadingSpinner.test.js#L84` still asserts `bg-opacity-80` as a class string, so the suite stays green while the style is dead.

https://github.com/OneBusAway/wayfinder/blob/a383acfb02ba3a94753f1b849d1b1027a94e8812/src/components/LoadingSpinner.svelte#L8-L10

2. The v4 preflight changes the default border color from `gray-200` to `currentColor` (`*, ::after, ... { border: 0 solid }`), and no compatibility base layer was added to `src/app.css` to preserve the old default. Three elements on lines this PR modified use a bare `border` with no color class, all in header chrome: the nav-link chip below, the overflow-menu button at `src/components/navigation/Header.svelte#L265`, and the language-switcher button at `src/components/navigation/LanguageSwitcher/LanguageSwitcher.svelte#L134`. The first two inherit the header's `text-brand-foreground` (white by default), so their borders go from gray to effectively invisible against `bg-surface/80`; the switcher inherits `text-surface-foreground` and goes to a hard black border. This is the area commits `7194fd8` and `2712902` were recently tuning for contrast. Either add the `border-color: var(--color-gray-200, currentColor)` compat layer that `@tailwindcss/upgrade` normally injects, or give these three an explicit border color. The visual QA in the description was dark-mode only, which is why this wouldn't have shown up.

https://github.com/OneBusAway/wayfinder/blob/a383acfb02ba3a94753f1b849d1b1027a94e8812/src/components/navigation/Header.svelte#L250-L252

3. `docs/tailwind-bef/` adds 9 PNGs totalling ~18 MB to the repo, with raw macOS screenshot filenames (including the narrow no-break space characters). Nothing references them and there is no matching "after" set. These look like the QA scratch artifacts mentioned in the description rather than intended documentation; once merged they are in the history permanently.

https://github.com/OneBusAway/wayfinder/tree/a383acfb02ba3a94753f1b849d1b1027a94e8812/docs/tailwind-bef

src/components/LoadingSpinner.svelte
```diff
@@ -6,11 +6,11 @@
 	aria-label={$t('loading')}
 	role="status"
 	aria-live="polite"
-	class="absolute inset-0 z-50 flex items-center justify-center bg-neutral-800 bg-opacity-80 md:rounded-lg"
+	class="bg-opacity-80 absolute inset-0 z-50 flex items-center justify-center bg-neutral-800 md:rounded-lg"
```
aaronbrethorst, line 9: `bg-opacity-80` is on v4's removed-deprecated list, so this emits **no CSS** — silently, no build error.

The loading overlay goes from 80% to fully opaque, which means the map is completely hidden behind it during every load rather than dimmed.

```diff
-class="bg-opacity-80 absolute inset-0 z-50 ... bg-neutral-800 md:rounded-lg"
+class="absolute inset-0 z-50 ... bg-neutral-800/80 md:rounded-lg"
```

Same family to sweep for across the tree: `text-opacity-*`, `border-opacity-*`, `divide-opacity-*`, `placeholder-opacity-*`, `ring-opacity-*` (there's one in `SurveyLauncher.svelte:43`).

src/components/routes/RouteModal.svelte
```diff
@@ -77,7 +77,7 @@
 	{#if stops && selectedRoute}
 		<div class="space-y-4">
 			<div>
-				<div class="min-h-36 rounded-lg bg-brand-accent bg-opacity-80 p-4">
+				<div class="bg-brand-accent bg-opacity-80 min-h-36 rounded-lg p-4">
```
aaronbrethorst, line 80: Same as `LoadingSpinner` — `bg-opacity-80` emits nothing in v4, so this header renders at full opacity.

`bg-brand-accent/80` is the replacement.

src/app.css
```diff
@@ -1,22 +1,24 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import 'tailwindcss';
+@config '../tailwind.config.js';
```
aaronbrethorst, line 2: This is the right place for the v4 default-border-color compat layer, and it's missing.

In v3 a bare `border` defaulted to `gray-200`; in v4 it defaults to `currentColor`, so every bare `border` now draws in the element's text color. There are several on lines this PR modifies (the `LanguageSwitcher` / `Header` dropdown uses `rounded-md border` with no color three times).

The upgrade guide's own snippet:

```css
@layer base {
  *, ::after, ::before, ::backdrop, ::file-selector-button {
    border-color: var(--color-gray-200, currentColor);
  }
}
```

Auditing each bare `border` and giving it an explicit color is the better end state, but for a migration PR the compat layer keeps the blast radius at zero.

aaronbrethorst (changes requested):
The structural decisions here are all the ones I'd have made. Moving to `@tailwindcss/vite` and dropping `postcss.config.js`/`autoprefixer`, keeping `tailwind.config.js` alive through the `@config` bridge so the `COLOR_*` → `generatePalette()` pipeline keeps working, and the `@custom-variant dark (&:is(.dark, .dark *))` fix for v4's dark-mode specificity — that's a well-reasoned path through a migration that could easily have turned into a rewrite. Deferring the Flowbite upgrade and paying for it with a scoped Vite patch is a defensible trade too, and it's honestly commented.

The title says WIP and I think that's accurate. Three things need fixing, and two of them silently change what renders.

## 1. Dead opacity utilities survive the codemod

`bg-opacity-*` and `ring-opacity-*` are on v4's removed-deprecated list — they emit no CSS at all, so these fail silently rather than erroring:

- `LoadingSpinner.svelte:9` — `bg-opacity-80` on `bg-neutral-800`. The loading overlay becomes fully opaque instead of 80%, so the map disappears behind it during every load.
- `RouteModal.svelte:80` — `bg-opacity-80` on `bg-brand-accent`. Same, on the route modal header.
- `SurveyLauncher.svelte:43` — `focus:ring-opacity-50`.

The replacement is the slash modifier: `bg-neutral-800/80`, `bg-brand-accent/80`, `focus:ring-green-400/50`. Worth grepping the whole tree for the rest of that family (`text-opacity-*`, `border-opacity-*`, `divide-opacity-*`, `placeholder-opacity-*`) since the codemod clearly didn't catch them all.

## 2. v4's default border color change isn't compensated

In v3, `border` defaulted to `gray-200`. In v4 it defaults to `currentColor`, so every bare `border` with no explicit color now draws in the element's text color. `src/app.css` has no base layer restoring the old behavior, and there are bare `border` classes on lines this PR modified — several in the `LanguageSwitcher` / `Header` dropdown region:

```
rounded-md border            (flex-shrink-0 wrapper)
rounded-md border            (8x8 button)
rounded-md border px-2       (font-semibold trigger)
```

Those pick up the surrounding text color, which on the dark surface means a border that either changes color or effectively vanishes. This is also almost certainly why it slipped through — the QA pass was dark-mode only.

The upgrade guide's own compat snippet is the cheapest fix, and it goes in `app.css`:

```css
@layer base {
  *, ::after, ::before, ::backdrop, ::file-selector-button {
    border-color: var(--color-gray-200, currentColor);
  }
}
```

Auditing every bare `border` and giving it an explicit color is the better long-term answer, but for a migration PR I'd take the compat layer and do the audit separately.

## 3. 19 MB of screenshots got committed

`docs/tailwind-bef/` has nine PNGs totaling about 19 MB — raw macOS screenshots with timestamp filenames, no "after" set, and nothing referencing them. Those want to be in the PR description, not in git history where they're permanent. Please drop them from the branch.

## Smaller notes

- The PR body says "Adapt `tailwind.config.js` for ESM plugin imports," but that file is byte-identical to `develop` — `require('@tailwindcss/forms')` and all. That's harmless (jiti handles `require` in an ESM config under `@config`), but the description should match the diff.
- `focus:outline-none` is still used in ~14 places. In v4 that's `outline-hidden`; plain `outline-none` drops v3's forced-colors fallback. Most sites pair it with a visible `focus:ring-*`, so the loss is confined to forced-colors mode — low priority, but it's part of the same codemod gap.
- The codemod rewrote class strings inside `docs/superpowers/plans/*` and `specs/*`. Those are dated historical records; rewriting them makes them no longer describe the code as it was. I'd revert those files.
- `overrides: { "@mapbox/jsonlint-lines-primitives": "2.0.2" }` appears with no explanation — worth a line in the body or a comment.
- `postcss` is still in devDependencies after `postcss.config.js` was deleted.
- The `flowbiteSvelteNoCircular` plugin regex-patches `node_modules` source, so it'll silently no-op on any flowbite-svelte update rather than failing loudly. Worth a version assertion so a future bump doesn't quietly break the build.

Worth saying explicitly since the PR reads as reaching v4: this lands on the *compat* path, not the CSS-first `@theme` migration — `tailwind.config.js` still drives everything through `@config`. I think that's the right call given the brand pipeline needs JS at build time, and I verified the `COLOR_*` flow still works end-to-end under v4. Just want it to be a stated choice rather than an assumed way-station.

Fix the opacity utilities and the border default, drop the screenshots, and I'll do a full pass on the rest. This is good work on a genuinely unpleasant migration.

PR #596, opened 2026-08-23
docs: refresh README and CONTRIBUTING for setup and pre-PR checks

tarunsinghofficial:
Updated README and CONTRIBUTING files for better readability and usage.

## Summary by CodeRabbit

* **Documentation**
  * Added environment validation steps to development setup instructions.
  * Clarified commands for starting the app and running tests, linting, and pre-push checks.
  * Updated pull request guidance with recommended validation steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,19 @@ This is the next-generation OneBusAway web application, built on top of [SvelteK
 ```bash
 npm install
 cp .env.example .env
-# edit .env with your editor of choice
+# edit .env with your values, then check them:
+npm run validate-env
 npm run dev
+```
+
+Required and optional variables are listed in `.env.example`. Run `npm run validate-env` to check your `.env` against `env-schema.json` before starting the app.
+
+## Testing and checks
+
+```bash
+npm run test          # run the test suite once
+npm run lint          # prettier + eslint
+npm run prepush       # format, lint, and test (handy before opening a PR)
 ```
 
 ## `.env` File Keys
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,19 @@ Thanks for your interest in contributing! Wayfinder is an open source project an
 
 ## Getting Started
 
-Head over to the [README](README.md) for setup instructions — it's just `npm install`, copy `.env.example` to `.env`, and `npm run dev`. If you hit any issues getting things running, open a discussion or ask in the [OneBusAway Slack workspace](https://opentransitsoftwarefoundation.org/join-our-slack).
+Head over to the [README](README.md) for setup. In short:
+
+```bash
+npm install
+cp .env.example .env
+# edit .env, then:
+npm run validate-env
+npm run dev
+```
+
+`npm run validate-env` checks your `.env` against `env-schema.json`. Fix anything it reports before debugging runtime errors.
+
+If you still get stuck, open a discussion or ask in the [OneBusAway Slack workspace](https://opentransitsoftwarefoundation.org/join-our-slack).
 
 ## Finding Something to Work On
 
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `npm run test` does not "run the test suite once" — `package.json` defines `"test": "vitest"` (no `run` subcommand), so in a normal terminal it starts Vitest in watch mode and never exits. The scripts that actually run once are `vitest run`-based (`test:coverage`, `test:components`, `prepush`). Since this README block is what a new contributor will copy, the comment should say watch mode, or point at `npx vitest run` / `npm run test:coverage` for a single pass (bug due to `package.json` `"test": "vitest"` vs `"prepush": "... && vitest run"`).

https://github.com/OneBusAway/wayfinder/blob/27714182be44cbf10b62ad92bf675c54b9b80f9b/README.md#L21-L23

README.md
```diff
@@ -9,10 +9,21 @@ This is the next-generation OneBusAway web application, built on top of [SvelteK
...
+npm run validate-env
 npm run dev
 ```
 
+Required and optional variables are listed in `.env.example`. Run `npm run validate-env` to check your `.env` against `env-schema.json` before starting the app.
+
+## Testing and checks
+
+```bash
+npm run test          # run the test suite once
```
aaronbrethorst, line 22: `npm run test` is `vitest` with no `run` subcommand, so in a TTY this starts **watch mode** and never exits — it doesn't "run the test suite once."

The one-shot scripts are the `vitest run`-based ones:

```bash
npm run test:coverage   # vitest run --coverage
npm run test:components # vitest run src/components
npm run prepush         # ... && vitest run
```

Either swap in one of those, label this one as watch mode, or use `npx vitest run` directly.

aaronbrethorst (changes requested):
Thanks for this — the README and CONTRIBUTING both needed exactly this kind of pass, and most of it is accurate. I checked every command, env var, and path you reference against `develop`: `validate-env`, `lint`, `prepush`, and `format` all exist and do what you say they do, `scripts/validate-env.js` really does read `.env` and check it against `env-schema.json`, and both `.env.example` and `env-schema.json` are where you point people. Replacing the "run `npm run format`" bullet with `npm run lint` / `npm run prepush` is the right advice.

One line needs to change before this lands, and it's the one a new contributor is most likely to copy.

## `npm run test` does not run the suite once

`package.json` on `develop` defines `"test": "vitest"` with no `run` subcommand, and `vite.config.js` doesn't set `watch: false`. In a terminal that's watch mode — it runs the suite and then sits there waiting. Someone following your new "Testing and checks" block gets a session that never returns to the prompt, which is a rough first impression for a doc whose whole job is getting people started.

The one-shot commands are the `vitest run`-based ones: `npm run test:coverage`, `npm run test:components`, or `npm run prepush`. Either point at one of those or label this one honestly as watch mode.

Worth knowing, and definitely not your fault: `CLAUDE.md` makes the same wrong claim, and CONTRIBUTING's "Make sure tests pass. Run `npm run test` locally" line (which you didn't touch) inherits it too. If you want to fix all three in this PR that'd be a real service — but I'll take just the line you added, and I'm happy to clean up the rest myself.

Two smaller notes, take or leave:

- Dropping `npm run format` from CONTRIBUTING means the auto-fix command is no longer discoverable there. `npm run lint` only *checks* formatting and fails on it; a contributor who hits that failure now has to go find `format` on their own. Mentioning both would be kinder.
- "Required and optional variables are listed in `.env.example`" is true as far as it goes, but `.env.example` doesn't mark which are which — `env-schema.json` is what actually encodes required-ness, and `validate-env` is what surfaces it.

Fix the `npm run test` line and I'll merge this.

PR #597, opened 2026-08-24
Fix short-line route schedules

### LINKED ISSUE #251: When viewing the Route Schedule, short-lines aren't handled properly

soumajitgh:
## Summary

- Label short-line departures using per-trip headsigns from `schedule-for-route` references.
- Cache only trip headsigns by route and service date, with a 24-hour TTL, a 100-entry limit, and shared in-flight lookups.
- Resolve undated requests once in `PUBLIC_OBA_TIMEZONE` and send the same explicit date to both OBA endpoints and the cache.

## Design

image

## Testing

- All 1,814 tests passed; 93.94% line coverage.
- `npm run lint` and `npm run build` passed.
- Verified live OBA stop/route response shapes with HTTP 200 responses.
- Added regressions for changing `entry.date` timestamps, region midnight, DST transitions, and upstream requests crossing midnight. The corrected tests fail against the previous implementation.

Fixes #251

## Summary by CodeRabbit

* **New Features**
  * Short-line trips are now identified in stop schedules with highlighted times and destination labels.
  * A notice appears when short-line service is present in the schedule.
  * Trip destinations are shown more accurately when they differ from the route direction.

* **Bug Fixes**
  * Stop schedules now consistently use the correct service date, including timezone and daylight-saving transitions.
  * Schedule information remains available when individual destination lookups fail.

### DIFF
--- a/src/routes/stops/[stopID]/schedule/+page.svelte
+++ b/src/routes/stops/[stopID]/schedule/+page.svelte
@@ -3,7 +3,7 @@
 	import RouteScheduleTable from '$components/schedule-for-stop/RouteScheduleTable.svelte';
 	import StopPageHeader from '$components/stops/StopPageHeader.svelte';
 	import StandalonePage from '$components/StandalonePage.svelte';
-	import { msToTimeString } from '$lib/dateTimeFormat.js';
+	import { groupStopTimesByHour } from '$lib/scheduleForStop.js';
 	import Accordion from '$components/containers/Accordion.svelte';
 	import AccordionItem from '$components/containers/AccordionItem.svelte';
 	import { Datepicker } from 'flowbite-svelte';
@@ -14,7 +14,6 @@
 	let selectedDate = $state(new Date());
 	let prevSelectedDate = $state(null);
 	let emptySchedules = $state(false);
-	let scheduleError = $state(false);
 	let schedules = $state([]);
 	let stopName = $state('');
 	let stopId = $state('');
@@ -31,21 +30,12 @@
 	async function fetchScheduleForStop(stopId, date) {
 		try {
 			emptySchedules = false;
-			scheduleError = false;
 			const response = await fetch(`/api/oba/schedule-for-stop/${stopId}?date=${date}`);
 			if (!response.ok) throw new Error('Failed to fetch schedule for stop');
 			const scheduleForStop = await response.json();
 			handleScheduleForStopResponse(scheduleForStop.data);
 		} catch (error) {
 			console.error('Error fetching schedules:', error);
-			schedules = [];
-			scheduleError = true;
-		}
-	}
-
-	function retryScheduleFetch() {
-		if (stopId && selectedDate) {
-			fetchScheduleForStop(stopId, selectedDate.toISOString().split('T')[0]);
 		}
 	}
 
@@ -101,25 +91,6 @@
 	function getRouteName(routeId, tripHeadsign) {
 		const route = routeReference.get(routeId);
 		return `${route.shortName ?? route.longName} - ${tripHeadsign}`;
-	}
-
-	function groupStopTimesByHour(stopTimes, directionHeadsign) {
-		const grouped = {};
-		for (let stopTime of stopTimes) {
-			const date = new Date(stopTime.arrivalTime);
-			const hour = date.getHours();
-			if (!grouped[hour]) grouped[hour] = [];
-
-			// The direction headsign describes the route's usual destination. A stop-specific
-			// headsign is supplied for trips that take a different path or terminate early.
-			const destination = stopTime.stopHeadsign?.trim() || directionHeadsign;
-			grouped[hour].push({
-				arrivalTime: msToTimeString(stopTime.arrivalTime),
-				destination,
-				isShortLine: destination !== directionHeadsign
-			});
-		}
-		return grouped;
 	}
 
 	function toggleAllRoutes() {
@@ -198,19 +169,7 @@
 			<div
 				class="flex-1 rounded-lg border border-gray-200 bg-white p-2 dark:border-gray-700 dark:bg-black"
 			>
-				{#if scheduleError}
-					<div
-						role="alert"
-						class="mx-auto flex max-w-lg flex-col items-center gap-3 px-4 py-10 text-center"
-					>
-						<p class="text-gray-700 dark:text-gray-300">
-							{$isLoading ? '' : $t('schedule_for_stop.unable_to_load_schedule')}
-						</p>
-						<button class="button" onclick={retryScheduleFetch}>
-							{$isLoading ? '' : $t('schedule_for_stop.retry')}
-						</button>
-					</div>
-				{:else if emptySchedules}
+				{#if emptySchedules}
 					<p class="text-center text-gray-700 dark:text-gray-400">
 						{$isLoading ? '' : $t('schedule_for_stop.no_schedules_available')}
 					</p>
--- a/src/routes/stops/[stopID]/schedule/+page.svelte
+++ b/src/routes/stops/[stopID]/schedule/+page.svelte
@@ -14,6 +14,7 @@
 	let selectedDate = $state(new Date());
 	let prevSelectedDate = $state(null);
 	let emptySchedules = $state(false);
+	let scheduleError = $state(false);
 	let schedules = $state([]);
 	let stopName = $state('');
 	let stopId = $state('');
@@ -30,12 +31,21 @@
 	async function fetchScheduleForStop(stopId, date) {
 		try {
 			emptySchedules = false;
+			scheduleError = false;
 			const response = await fetch(`/api/oba/schedule-for-stop/${stopId}?date=${date}`);
 			if (!response.ok) throw new Error('Failed to fetch schedule for stop');
 			const scheduleForStop = await response.json();
 			handleScheduleForStopResponse(scheduleForStop.data);
 		} catch (error) {
 			console.error('Error fetching schedules:', error);
+			schedules = [];
+			scheduleError = true;
+		}
+	}
+
+	function retryScheduleFetch() {
+		if (stopId && selectedDate) {
+			fetchScheduleForStop(stopId, selectedDate.toISOString().split('T')[0]);
 		}
 	}
 
@@ -74,7 +84,10 @@
 			let stopRouteDirectionSchedules = routeSchedule.stopRouteDirectionSchedules;
 
 			stopRouteDirectionSchedules.forEach((directionSchedule) => {
-				const stopTimesGroupedByHour = groupStopTimesByHour(directionSchedule.scheduleStopTimes);
+				const stopTimesGroupedByHour = groupStopTimesByHour(
+					directionSchedule.scheduleStopTimes,
+					directionSchedule.tripHeadsign
+				);
 				const routeName = getRouteName(routeId, directionSchedule.tripHeadsign);
 
 				schedulesMap.set(routeName, {
@@ -90,14 +103,20 @@
 		return `${route.shortName ?? route.longName} - ${tripHeadsign}`;
 	}
 
-	function groupStopTimesByHour(stopTimes) {
+	function groupStopTimesByHour(stopTimes, directionHeadsign) {
 		const grouped = {};
 		for (let stopTime of stopTimes) {
 			const date = new Date(stopTime.arrivalTime);
 			const hour = date.getHours();
 			if (!grouped[hour]) grouped[hour] = [];
+
+			// The direction headsign describes the route's usual destination. A stop-specific
+			// headsign is supplied for trips that take a different path or terminate early.
+			const destination = stopTime.stopHeadsign?.trim() || directionHeadsign;
 			grouped[hour].push({
-				arrivalTime: msToTimeString(stopTime.arrivalTime)
+				arrivalTime: msToTimeString(stopTime.arrivalTime),
+				destination,
+				isShortLine: destination !== directionHeadsign
 			});
 		}
 		return grouped;
@@ -179,7 +198,19 @@
 			<div
 				class="flex-1 rounded-lg border border-gray-200 bg-white p-2 dark:border-gray-700 dark:bg-black"
 			>
-				{#if emptySchedules}
+				{#if scheduleError}
+					<div
+						role="alert"
+						class="mx-auto flex max-w-lg flex-col items-center gap-3 px-4 py-10 text-center"
+					>
+						<p class="text-gray-700 dark:text-gray-300">
+							{$isLoading ? '' : $t('schedule_for_stop.unable_to_load_schedule')}
+						</p>
+						<button class="button" onclick={retryScheduleFetch}>
+							{$isLoading ? '' : $t('schedule_for_stop.retry')}
+						</button>
+					</div>
+				{:else if emptySchedules}
 					<p class="text-center text-gray-700 dark:text-gray-400">
 						{$isLoading ? '' : $t('schedule_for_stop.no_schedules_available')}
 					</p>
--- a/src/components/schedule-for-stop/RouteScheduleTable.svelte
+++ b/src/components/schedule-for-stop/RouteScheduleTable.svelte
@@ -16,15 +16,19 @@
 	);
 
 	function renderScheduleTable(schedule) {
+		const stopTimes = Object.entries(schedule.stopTimes);
+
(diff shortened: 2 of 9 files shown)


soumajitgh:
@coderabbitai full review

aaronbrethorst:
### Code review

Found 1 issue:

1. `isShortLine` can never become true against real OBA data, so #251 isn't actually fixed. `scheduleStopTime.stopHeadsign` is empty on every trip for the exact route/stop the issue cites — a live call to `https://realtime.sdmts.com/api/api/where/schedule-for-stop/MTS_12434.json` returns 37 stop times for `MTS_120`, all with `stopHeadsign: ""` (Puget Sound returns `""` as well). `'' ?.trim() || directionHeadsign` therefore always falls back to the direction headsign, `destination !== directionHeadsign` is always false, `hasShortLines` is always false, and the whole amber short-line UI is dead code in production. The root of #251 is a level up: OBA's `stopRouteDirectionSchedules[].tripHeadsign` is a *merged* direction label (Puget Sound returns `"Northgate Station Roosevelt Station"` for one direction; San Diego collapses the Fashion Valley and Kearny Mesa trips under a single `"Kearny Mesa"`), so the per-trip destination has to be recovered from per-trip data, not from `stopHeadsign`. The code comment on these lines asserts the opposite of what the API actually returns. None of the new tests catch this: `RouteScheduleTable.test.js` hand-injects `isShortLine`/`destination` into the component props, and nothing exercises `groupStopTimesByHour` — the same gap that was flagged on the earlier attempt in #441. (bug due to `src/routes/stops/[stopID]/schedule/+page.svelte`)

https://github.com/OneBusAway/wayfinder/blob/def586b98fc9508542743e7c56622d3e6ee443a2/src/routes/stops/[stopID]/schedule/+page.svelte#L112-L121

src/routes/stops/[stopID]/schedule/+page.svelte
```diff
@@ -90,14 +103,20 @@
...
+	function groupStopTimesByHour(stopTimes, directionHeadsign) {
 		const grouped = {};
 		for (let stopTime of stopTimes) {
 			const date = new Date(stopTime.arrivalTime);
 			const hour = date.getHours();
 			if (!grouped[hour]) grouped[hour] = [];
+
+			// The direction headsign describes the route's usual destination. A stop-specific
+			// headsign is supplied for trips that take a different path or terminate early.
+			const destination = stopTime.stopHeadsign?.trim() || directionHeadsign;
```
Comment (aaronbrethorst) [inline src/routes/stops/[stopID]/schedule/+page.svelte]: `stopHeadsign` is empty on every stop time this endpoint returns, so this always falls through to `directionHeadsign`.

Verified against the stop named in #251:

```
GET /api/where/schedule-for-stop/MTS_12434.json

MTS_3   | tripHeadsign='UCSD Med Ctr/Hillcrest' | 77 stop times | stopHeadsign: {'': 77}
MTS_120 | tripHeadsign='Kearny Mesa'            | 37 stop times | stopHeadsign: {'': 37}
```

114/114 empty. `references.trips` is also `0` on this endpoint, so the per-trip headsign isn't reachable from here either.

src/routes/stops/[stopID]/schedule/+page.svelte
```diff
@@ -90,14 +103,20 @@
...
 			grouped[hour].push({
-				arrivalTime: msToTimeString(stopTime.arrivalTime)
+				arrivalTime: msToTimeString(stopTime.arrivalTime),
+				destination,
+				isShortLine: destination !== directionHeadsign
```
Comment (aaronbrethorst) [inline src/routes/stops/[stopID]/schedule/+page.svelte]: Given the line above, this comparison is always `false` — which makes `hasShortLines` always `false` and the whole amber banner/chip UI dead code in production.

The comment two lines up ("A stop-specific headsign is supplied for trips that take a different path or terminate early") describes behavior the API doesn't have. What OBA actually does is *merge* the variants into the direction label — Puget Sound returns `tripHeadsign: "Northgate Station Roosevelt Station"` for one direction, which is the real shape of #251.

aaronbrethorst (changes requested):
The supporting pieces of this are good. The `response?.data` guard on the schedule endpoint is a real fix, the retryable error state is a genuine improvement over a silently blank panel, and converting the per-hour cells from `<td>` to `<th scope="row">` is the correct table semantics. The case-insensitive `extractMinutes` regex is a nice catch.

But the headline feature doesn't work against real OBA data, and I verified that rather than reasoning about it.

## `isShortLine` can never be true

The detection is:

```js
const destination = stopTime.stopHeadsign?.trim() || directionHeadsign;
isShortLine: destination !== directionHeadsign
```

That requires `scheduleStopTimes[].stopHeadsign` to be populated. It isn't. I pulled the exact stop from #251 — MTS_12434 on the San Diego server:

```
MTS_3   | tripHeadsign='UCSD Med Ctr/Hillcrest' | 77 stop times | stopHeadsign: {'': 77}
MTS_120 | tripHeadsign='Kearny Mesa'            | 37 stop times | stopHeadsign: {'': 37}
references.trips: 0
```

Every one of the 114 stop times has `stopHeadsign: ""`. So `destination` always falls back to `directionHeadsign`, the comparison is always `false`, `hasShortLines` is always `false`, and the entire amber UI — banner, chips, the `short_line*` strings — is unreachable in production. Merging this would close #251 without fixing anything a rider can see.

The mechanism behind #251 is the other way around: OBA *collapses* the variants into the direction label. On Puget Sound you can watch it happen — one direction comes back as `tripHeadsign: "Northgate Station Roosevelt Station"`, two destinations concatenated. And `references.trips` is empty on this endpoint, so the per-trip headsign isn't available here at all. A real fix needs a different data source — probably `trip-details` per trip, or `stops-for-route` — which is a bigger design question worth settling before writing the UI.

The code comment asserting "A stop-specific headsign is supplied for trips that take a different path or terminate early" says the opposite of what the API returns, so that's worth correcting whichever direction this goes.

## The tests can't catch that

`RouteScheduleTable.test.js` hand-injects `isShortLine: true` and `destination` as props, so `groupStopTimesByHour` — where the actual detection lives — has zero coverage. That's the same point Copilot raised on your earlier attempt at this in #441, and it's why the suite is green on a feature that never activates.

A couple of the other new assertions can't fail either:

- `expect(screen.queryByText(/05 am/i)).not.toBeInTheDocument()` passes without the `/i` fix too — unfixed output is `"05am"` with no space, so the pattern misses either way. (The `getByTitle('Full Time: 8:05')` assertions in that test *are* real.)
- `expect(screen.queryByText('Short line to Kearny Mesa')).not.toBeInTheDocument()` can never match, since non-short-line chips render no destination text at all.
- In `schedule-for-stop.test.js`, `expect(response).toBeNull()` is asserting on the mock — `handleOBAResponse` is stubbed to echo its argument. It does still prove no `TypeError` escapes, which is the point of the guard, so that one earns partial credit.

## On scope

The AM/PM redesign is a separate change riding along. It removes the section rowgroups and replaces the `no_am_schedules_available` / `no_pm_schedules_available` empty states for *every* route, short-line or not, and it leaves those two strings dead in `en.json` and the other 24 locale files. Same for the error/retry state — useful, unrelated.

I closed #441 asking for smaller PRs and I'll ask again here: the endpoint guard and the error/retry state would sail through on their own, and I'd merge that today. The short-line feature needs the data question answered first.

Happy to talk through where the real headsign data should come from if that's useful — it's the interesting part of this problem and I don't want the work you've already done to go to waste.

soumajitgh:
Addressed in commit 24304b4.

- The schedule endpoint now enriches multi-trip route schedules with each trip headsign from schedule-for-route: one parallel lookup per route, not one per trip.
- Short-line detection now uses that per-trip value; it no longer relies on schedule stop times stopHeadsign.
- Added API and transformation coverage using the reported production shape with an empty stopHeadsign and a short Fashion Valley trip under the Kearny Mesa direction.
- Removed the unrelated AM/PM redesign, lowercase meridiem change, retry UI, and endpoint guard from this PR final diff.

The full Vitest suite passes: 105 files, 1,799 tests.

aaronbrethorst (changes requested):
The cleanup here is real and I want to acknowledge it before the bad news. Scope is tight now — the AM/PM redesign, the retry state, and the endpoint guard are all out, and the PR is just short lines. The false comment about `stopHeadsign` is gone and replaced with an accurate one. `groupStopTimesByHour` moved to `src/lib/scheduleForStop.js` and finally has direct unit coverage, which was the gap I cared most about. And the two bogus test assertions are fixed properly.

More importantly: **you picked the right data source.** `schedule-for-route` really does carry the per-trip headsigns that `schedule-for-stop` collapses away. The design is correct.

The field path isn't, and it's the same failure mode as last time in a new location.

## `data.entry.trips` doesn't exist

`addTripHeadsigns` reads `routeResponse?.data?.entry?.trips`. I called the endpoint against Puget Sound to be sure:

```
GET /api/where/schedule-for-route/1_100259.json

data.entry keys       : [ 'routeId', 'scheduleDate', 'serviceIds', 'stopTripGroupings' ]
data.entry.trips      : undefined
data.references keys  : [ 'agencies', 'routes', 'situations', 'stopTimes', 'stops', 'trips' ]
data.references.trips : 130
```

The trips are in `data.references.trips`. So the `tripHeadsigns` Map is always empty, no `stopTime.tripHeadsign` is ever set, `destination` always falls back to `directionHeadsign`, and `isShortLine` is always `false`. The amber banner and the chips are still unreachable in production — exactly the state I rejected last time.

I know where this came from and it isn't carelessness: the SDK's own generated type at `node_modules/onebusaway-sdk/resources/schedule-for-route.d.ts` declares `entry.trips` (and `entry.stops`). The generated type is wrong about the real API. Worth knowing generally — check OBA response shapes against a live call rather than the `.d.ts`.

## The good news: you're one line from a working feature

I ran the corrected logic against live data. Trip-ID resolution was 100%, and it finds genuine short lines:

| stop | direction label | per-trip headsigns | flagged |
| --- | --- | --- | --- |
| 1_41255 (rt 4) | `Judkins Park Downtown Seattle` | 40x same, 3x `Downtown Seattle` | 3 |
| 1_18085 (rt 44) | `UW Medical Center Wallingford` | 69x same, 9x `University District` | 9 |
| 1_56151 (rt 36) | `Downtown Seattle N Beacon Hill` | 101x same, 6x `Intl Dist Chinatown Station N Beacon Hill` | 6 |

So `references.trips` makes this work. Two things have to come with that change:

**The new API test will block the fix.** `src/tests/api/schedule-for-stop.test.js` mocks `scheduleForRoute.retrieve` as `{ data: { entry: { trips: [...] } } }` — a response the API never produces. That's worse than no coverage: correct the field path and this test starts failing. Please rewrite the mock to the real shape (`data.references.trips`) so it fails on the current code and passes on the fix.

**The payload cost needs an answer.** One `schedule-for-route` per route at the stop, measured live:

```
schedule-for-route 1_100259 : 1.2 MB
schedule-for-route 1_100219 : 1.3 MB
schedule-for-route 1_102615 : 2.5 MB
schedule-for-stop  1_23230  : 45 KB
```

Stop 1_23230 serves 4 routes, so that's roughly 5 MB fetched to enrich a 45 KB response, on every page load and every date change, uncached. `Promise.all` parallelizes the latency but not the bytes, and the `> 1 distinct tripId` guard prunes almost nothing — every direction I sampled had 38 to 107 distinct trips. I don't need it solved perfectly, but I do need a plan: caching by `(routeId, scheduleDate)` would probably do it, since the data is static for a service day.

## Two smaller things

- `addTripHeadsigns` will throw and 500 the whole endpoint if any `stopRouteSchedules` entry is missing `stopRouteDirectionSchedules` or `scheduleStopTimes`. On `develop` these were only touched client-side, where a malformed entry degraded one route instead of the request.
- `RouteScheduleTable.test.js` drops the `expect(amHeader.tagName).toBe('TH')` / `expect(pmHeader.tagName).toBe('TH')` assertions that exist on `develop`. The component still renders `<th scope="rowgroup">`, so that's pure coverage loss — looks like leftover from reverting the AM/PM work. Please put them back.

Also: you removed the `response?.data` guard and the retry state rather than splitting them out. Those were the two pieces I said I'd merge today, so they'll need their own PR now — worth doing, they were both good.

Change the field path, fix the mock, tell me how you want to handle the payload size, and this lands.

soumajitgh:
Addressed in 988c702: corrected the lookup to `data.references.trips`, updated the API fixture and regression coverage, guarded malformed schedule entries, and restored the AM/PM `<th>` assertions.

Added a per-process cache keyed by route/date (24-hour TTL, 100-entry limit), storing only trip headsigns and sharing concurrent lookups. Failed lookups are retried; cold requests still fetch the full route schedule.

Validation: all 1,809 tests and coverage checks pass; lint and production build pass.

aaronbrethorst:
### Code review

Found 2 issues:

1. `response.data.entry.scheduleDate` does not exist on a schedule-for-stop response, so the cache is bypassed on every request that omits `?date=`. The field is `entry.date` — verified with a live call to `https://api.pugetsound.onebusaway.org/api/where/schedule-for-stop/1_75403.json`, whose `data.entry` keys are `['date', 'stopId', 'stopRouteSchedules']` (`entry.scheduleDate` is `undefined`). `scheduleDate` is the schedule-for-**route** entry field, not the stop one. Because it is always `undefined`, `getTripHeadsigns` takes the `if (date == null) return fetchTripHeadsigns(...)` branch and refetches the full route schedule (1.47 MB for `1_100259` on Puget Sound, measured) for every route on every undated request, which is the payload problem the cache was added to solve. The comment "Use the upstream service date for requests without an explicit date" describes behavior the endpoint does not have. Worth knowing before swapping in `entry.date`: it is only a service date when the caller passes `date=` (`date=2026-09-10` returns `1789023600000`, midnight); on an undated request it comes back as wall-clock-now to the millisecond (`1788817054393`, four ms off the response's own `currentTime`), so it is unusable as a cache key there and would thrash the 100-entry LRU. The undated branch probably needs a normalized service day derived server-side rather than an upstream field. (bug due to `src/routes/api/oba/schedule-for-stop/[stopId]/+server.js`)

https://github.com/OneBusAway/wayfinder/blob/988c702f4090e0c52a9e93c4738aa5f07c995425/src/routes/api/oba/schedule-for-stop/%5BstopId%5D/+server.js#L21-L25

2. The new API test fixture puts `scheduleDate` on the schedule-for-stop entry, a shape the API never returns, so the two cache tests that depend on it are green against a code path that cannot work in production. `stopResponse()` builds `data.entry.scheduleDate`, and `'separates routes and uses the upstream service day for undated requests'` and `'does not cache an undated response with no service date'` both assert on it — against the real envelope (`entry.date`) the first would fetch four times instead of three. This is the same failure mode as the `entry.trips` mock from the last round: correcting the field path will not make these tests fail first, and fixing them will not make them pass. (`routeResponse()`'s `entry.scheduleDate` on line 67 is correct — schedule-for-route really does have it.) (bug due to `src/tests/api/schedule-for-stop.test.js`)

https://github.com/OneBusAway/wayfinder/blob/988c702f4090e0c52a9e93c4738aa5f07c995425/src/tests/api/schedule-for-stop.test.js#L25-L33

soumajitgh:

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is ready. I read the final state myself, not just the bot pass.

`tripHeadsigns.js` reads `data.references.trips` and keeps only the headsigns. It caches the promise, not the resolved value, so concurrent requests for nearby stops share one upstream fetch, and a failed lookup is evicted so it can retry. Resolving undated requests in the handler with `getTodayDateForInput(env.PUBLIC_OBA_TIMEZONE)` is the right answer to the `entry.date` problem from the last round. It gives both upstream calls the same service-day key, and the cache no longer thrashes on wall-clock timestamps. The payload concern I raised is answered: a warm cache means one route fetch per route per service day, not one per page load.

The robustness points are covered too. Malformed `stopRouteSchedules` entries are skipped instead of throwing, a failed headsign lookup logs and falls back to the direction headsign instead of returning a 500, and the `TH`/`rowgroup` assertions are back in `RouteScheduleTable.test.js`.

I updated the branch against `develop` because it was a long way behind, and CI is green on the updated head. Merging. Thanks for working through the SDK type mismatch. That one caught me too.

PR #598, opened 2026-08-24
feat: add ArcGIS map provider

### LINKED ISSUE #316: Add support for ArcGIS Maps
https://developers.arcgis.com/javascript/latest/

We currently support Google Maps and OSM (via Leaflet). We should investigate adding support for ArcGIS with custom base maps(?)

soumajitgh:
Closes #316

## Summary

Adds ArcGIS Maps SDK as a third Wayfinder map provider while preserving the Google Maps and OpenStreetMap providers.

### Architecture

- Uses a provider factory with OSM fallback for an unknown configured provider.
- Dynamically loads @arcgis/core and creates isolated graphics layers for route stops, vehicles, user location, route casing, routes, and promoted routes.
- Uses one shared HTML overlay for Svelte stop and pin markers, shared reactive positioning, and a single shared ArcGIS hit-test click handler.
- Normalizes the provider contract around synchronous polylines, getZoom, WGS84 bounds, and idempotent teardown.

### Prior implementation material

The old arcgis branch informed the graphics-layer separation, Svelte marker overlay, reactive movement handling, encoded-polyline reuse, and arrow approach. PR #335 informed optional API keys, optional custom vector-tile basemaps, dynamic SDK loading, WGS84 extent conversion, and popup cleanup. Neither was merged because the arcgis branch is substantially behind and diverged from current develop.

### Provider parity

| Capability | Google | OSM | ArcGIS |
| --- | --- | --- | --- |
| Stops, pins, vehicles, popups | Yes | Yes | Yes |
| WGS84 bounds | Yes | Yes | Yes |
| Synchronous polylines | Yes | Yes | Yes |
| Theme-aware default basemap | Existing | Existing | Yes |
| Custom basemap preserved on theme change | N/A | N/A | Yes |

### Configuration

```env
PUBLIC_OBA_MAP_PROVIDER="arcgis"
PUBLIC_ARCGIS_API_KEY=""
PUBLIC_ARCGIS_CUSTOM_BASEMAP_URL=""
```

The API key is optional for public/free basemaps. A custom vector-tile basemap remains active during theme changes.

### Tests and validation

- Added provider-selection, configuration, lifecycle, popup, overlay, bounds, and polyline tests.
- npm run lint
- npm test -- --run: 105 files, 1804 tests passed
- npm run build
- npm run validate-env
- npm run test:coverage

Manual local verification included ArcGIS initial rendering and the refined context-menu and location-control presentation.

### Aaron requested-changes checklist

Addresses [Aaron’s requested-changes review](https://github.com/OneBusAway/wayfinder/pull/335#pullrequestreview-3787774285):

- [x] Prettier passes.
- [x] Reuses polyline-encoded.
- [x] No redundant graphic attributes assignment.
- [x] No circular HTML-marker reference.
- [x] hitTest failures are caught.
- [x] Marker collections are initialized consistently.
- [x] view.when rejection cleans partial state.
- [x] Popup watcher handles do not accumulate and mounted Svelte components unmount.
- [x] Off-screen HTML markers are hidden.
- [x] Unavailable or invalid bounds return null, avoiding Null Island requests.

### Known differences

ArcGIS uses its native vector basemaps and CIM repeated markers for directional arrows; arrow appearance can vary slightly from the other provider renderers.

## Summary by CodeRabbit

* **New Features**
  * Added ArcGIS as a map provider alongside OpenStreetMap and Google Maps.
  * Added optional ArcGIS API key and custom basemap URL configuration.
  * Added ArcGIS map styling for routes, markers, popups, themes, and dark mode.

* **Bug Fixes**
  * Improved handling of incomplete stop data and map teardown.
  * Improved cleanup for Google Maps and OpenStreetMap providers.
  * Restored map layout after closing trip-planning views.

* **Documentation**
  * Updated map provider configuration guidance.

* **Tests**
  * Added coverage for ArcGIS mapping and provider selection.


soumajitgh:
Addressed the CodeRabbit feedback in bbdd65b.

- Cancel queued map stop loads during teardown and guard async initialization.
- Use explicit .svelte.js provider imports and matching test mocks.
- Reset ArcGIS listener handles, make casing opacity SDK-compatible, and reapply dimming only after a replacement basemap is ready.
- Clean up OSM popup components and references during teardown.
- Avoid O(n²) marker removal and cover both optional API-key paths.

Validated with npm run lint, npm test -- --run (105 files, 1808 tests), and npm run build.

aaronbrethorst:
### Code review

Found 1 issue:

1. `updateMarkersRouteLabelVisibility()` is defined on the ArcGIS provider but never called — nothing in the provider or the app invokes it (the shared `extent`/`zoom`/`stationary` watch in `_registerSharedHandles()` only calls `_requestPositionUpdate()`). Both other providers wire this to a zoom listener inside `initMap()` (`GoogleMapProvider.svelte.js` `this.map.addListener('zoom_changed', ...)`, `OpenStreetMapProvider.svelte.js` `zoomend`). Result: with `PUBLIC_OBA_MAP_PROVIDER="arcgis"`, `showRoutesLabel` is frozen at whatever it was when each `StopMarker` was mounted, so route labels never appear/disappear as the rider zooms, and markers added after a zoom disagree with markers already on the map.

https://github.com/OneBusAway/wayfinder/blob/bbdd65b6103e54582acaa810d1ac2f3f6a3a416c/src/lib/Provider/ArcGISMapProvider.svelte.js#L330-L337

src/lib/Provider/ArcGISMapProvider.svelte.js
```diff
@@ -0,0 +1,887 @@
...
+	getMarker(stopId) {
+		return this.markersMap.get(stopId);
+	}
+
+	clearAllStopMarkers() {
+		for (const marker of [...this.markersMap.values()]) this.removeMarker(marker);
+		this.markersMap.clear();
+	}
+
+	updateMarkersRouteLabelVisibility() {
```
aaronbrethorst, line 340: Nothing calls this method.

Both existing providers hook it to a zoom event in `initMap()`:

```js
// GoogleMapProvider.svelte.js:80
this.map.addListener('zoom_changed', () => this.updateMarkersRouteLabelVisibility());

// OpenStreetMapProvider.svelte.js:111
this.map.on('zoomend', () => this.updateMarkersRouteLabelVisibility());
```

Here, `_registerSharedHandles()`'s extent/zoom/stationary watch only calls `_requestPositionUpdate()`, and there's no external caller anywhere in `src/`.

Since `showRoutesLabel` is set once at mount (line 293), stop route labels never toggle on zoom under `arcgis`, and markers added after a zoom disagree with ones already placed. Adding the call to the existing extent/zoom watch should be all it needs.

aaronbrethorst (changes requested):
This is a lot of well-organized work, and the parts I most expected to go wrong didn't. Things I checked specifically that hold up:

- **The SDK is genuinely code-split.** Every `@arcgis/core/*` runtime import is a dynamic `import()` inside `initMap()`, so an OSM or Google deployment never pulls the ArcGIS bundle. That was my main worry with adding a dependency this size and you handled it correctly.
- **The edits to the other two providers are safe and well-scoped.** Making `GoogleMapProvider.createPolyline` synchronous by hoisting the geometry library load into `initMap`, letting `getBoundingBox()` return `null`, adding `.cancel()` to `debounce()`, and giving all three a `destroy()` that `MapView` calls on unmount — that's the right way to normalize a contract rather than special-casing the new provider.
- Replacing the inline provider `if/else` in `MapContainer.svelte` with a `createMapProvider` factory (with an OSM fallback) is the correct shape.
- The items from the earlier #335 review are addressed.

One thing to fix before it lands, plus a few worth a look.

## `updateMarkersRouteLabelVisibility()` is never called

Both existing providers wire it to a zoom event inside `initMap()`:

```js
// GoogleMapProvider.svelte.js:80-82
this.map.addListener('zoom_changed', () => {
    this.updateMarkersRouteLabelVisibility();
});

// OpenStreetMapProvider.svelte.js:111-113
this.map.on('zoomend', () => {
    this.updateMarkersRouteLabelVisibility();
});
```

The ArcGIS provider defines the method but nothing calls it — `_registerSharedHandles()`'s extent/zoom/stationary watch only reaches `_requestPositionUpdate()`, and a repo-wide grep finds no external caller either. Since `showRoutesLabel` is a plain prop set once at mount (line 293, from `getZoom() >= showStopsRoutesAtZoom`), the effect is that under `arcgis` stop route labels never toggle on zoom, and markers added after a zoom disagree with markers already on the map. It's a one-line fix in the watch you already have.

## Worth a look, not blocking

- **`destroy()` during an in-flight `initMap()` doesn't abort view creation.** `initMap()` sets `this._destroyed = false`, then awaits ~17 dynamic imports plus `projection.load()` and `view.when()` without re-checking it. The ArcGIS SDK download is multi-MB, so that window is wide. If `MapView.onDestroy` fires inside it, `destroy()` runs against a null view and sets `_destroyed = true`, and `initMap()` then goes on to build a full `MapView` — WebGL context, watches, handlers — on a detached container that nothing will ever tear down, because the `catch` path's `this.destroy()` early-returns on `_destroyed`. `MapView.svelte` added `if (isDestroyed) return` guards for exactly this race, so the awareness is there; the provider just needs the matching re-check after the awaits. I'm not blocking on it since Google and OSM have the same shape, but their init is synchronous enough that it's theoretical there and real here.
- **`fitToPolylines()` ignores `padding`.** `TripPlanModal.svelte:149` passes `{ padding }`; the ArcGIS implementation accepts only `duration`/`maxZoom`, so a planned itinerary can fit underneath the trip sheet instead of clear of it.
- **`CLAUDE.md` is now stale.** The Architecture section still says "Both map providers ... implement the same interface" and documents `PUBLIC_OBA_MAP_PROVIDER` as `"osm" or "google"`. That file is what future contributors and agents read first, so it's worth updating in this PR.
- **`Extent` is imported and assigned to `this.Extent` but never used.** Property assignment, so nothing flags it.
- **`updatePopupContent()` closes and re-opens the popup** where Google and OSM swap content in place. Only reached on trip selection, so the flicker is rare, but it's a parity difference.
- **`arcgis-map.css` ships in every build.** It's a static top-level import in the provider, which `mapProviderFactory.js` imports statically — 45 lines of `.esri-*` rules that are inert for non-ArcGIS deployments. It's the one piece of this integration that isn't code-split, and it'd be nice to match the JS.
- **`getCenter()` returns `{lat: 0, lng: 0}`** when the view has no center, which is the Null Island shape this PR otherwise works to eliminate. Unreachable today because `getBoundingBox()` returns `null` first and `MapView` bails, but it's a trap for later.

Wire up the label visibility and I'll take another pass — the foundation here is good.

aaronbrethorst (changes requested):
Good round. The blocking item is properly fixed — `updateMarkersRouteLabelVisibility()` is now driven from the existing extent watch, the method matches the Google and OSM shapes including the `routeLabelsVisible` short-circuit, and the constructor seeds the cache consistently against the initial zoom. The regression test drives the watch callback directly, which is the right level to test it at.

The other items came back better than I asked for, not just cosmetically:

- The `destroy()`-during-`initMap()` race now has re-checks after all three async gaps plus the catch, and I walked the remaining windows — a `destroy()` landing inside `view.when()` finds `this.view` already assigned and tears it down properly. No detached view survives.
- `updatePopupContent()` swapping content in place also quietly fixes a shape bug: the old path passed `{ arrivalTime }` through `openStopMarker(stop, stopTime)`, and the new one binds it directly.
- Moving the CSS import inside `initMap()` behind the `browser` guard keeps the override order intact while getting it out of the main chunk.

Two things left, both small.

## 1. `setPadding` leaks permanent view padding on two of the three fit paths

`setPadding(padding = 50)` writes `this.view.padding`, which is durable view state rather than a per-call option — Google passes padding straight into `fitBounds` and OSM spreads it into `flyToBounds`, both scoped to the one animation. `fitToPolylines` calls `this.setPadding(options.padding)` unconditionally, and two of its three callers pass nothing:

- `src/components/search/SearchPane.svelte:178` — `await mapProvider.fitToPolylines?.()`
- `src/components/map/RouteMap.svelte:135` — `await mapProvider.fitToPolylines?.()`

`options.padding` is `undefined` there, so the default parameter fires and sets 50px on all four sides. Neither caller ever calls `resetPadding()` — only `TripPlanModal` does. Per the ArcGIS docs `padding` defines a subsection of the view for centering and extent calculations, so `view.extent` afterward describes the inset box, which is what `getBoundingBox()` hands to `/api/oba/stops-for-location`. Net effect under `PUBLIC_OBA_MAP_PROVIDER="arcgis"`: the first route selection leaves the view permanently padded, and stops are under-fetched at the viewport edges for the rest of the session.

Save and restore around the fit, or reset in a `finally` for calls that didn't supply a padding. While you're there, `fitToPolylines` sets padding before the `if (!this.polylines.length) return false` bail, so a no-op fit still mutates view state.

## 2. `CLAUDE.md:121` still has the old string

You updated the Architecture section — line 84 correctly reads `"osm"`, `"google"`, or `"arcgis"` — but the Environment Variables list further down still says:

```
- `PUBLIC_OBA_MAP_PROVIDER` - "osm" or "google"
```

That's the exact line I flagged. Adding `PUBLIC_ARCGIS_API_KEY` and `PUBLIC_ARCGIS_CUSTOM_BASEMAP_URL` to that list is your call — it's explicitly "Key variables" — but the provider line should match line 84.

## One to consider, not blocking

`getCenter()` returning `null` is the right contract, but `MapView.svelte:143` and `:242` both do `const center = mapInstance.getCenter(); ... center.lat` with no null guard. Reachability looks genuinely low, since `view.center` is populated once `view.when()` resolves and `mapInstance` is only assigned after that. Still, a one-line `if (!center) return;` at each call site would make the new contract safe, and it's worth noting OSM still returns `{lat: 0, lng: 0}` in the same situation, so the two providers now disagree.

Fix the padding leak and the CLAUDE.md line and this is ready.

soumajitgh:
Addressed the remaining review feedback in ee4eb8a: route fits restore prior padding in `finally`, empty fits leave padding untouched, the CLAUDE.md provider list includes ArcGIS, and both MapView center reads have null guards. Added padding regression tests.

Validation: lint and production build passed; all 1,816 tests passed (four metrics tests rerun with local port access).

soumajitgh:
Addressed all three latest CodeRabbit findings in 8654a01: overlapping ArcGIS fits share one padding baseline, walking-route dashes are preserved with or without arrows, and OSM stop/pin components unmount on removal and teardown. Added regression coverage for each.

Validation: all 1,825 tests, lint, and production build passed.

aaronbrethorst:
### Code review

Found 1 issue:

1. `_createRouteSymbol()` nests the CIM line symbol under `data.data` instead of `data.symbol`, so every arrowed or dashed route polyline gets a `CIMSymbolReference` with no symbol. The pinned SDK's own typings (`@arcgis/core@4.33.14/interfaces.d.ts`) declare `CIMSymbolReference { type; primitiveOverrides?; symbol?; minScale?; maxScale? }` — there is no `data` member, and every Esri CIM sample uses `data: { type: "CIMSymbolReference", symbol: { type: "CIMLineSymbol", ... } }`. Since `createPolyline` defaults `withArrow` to `true`, this is the branch taken for essentially every route line (SearchPane, RouteMap, StopRoutesLayer) and every trip-plan leg, so under `PUBLIC_OBA_MAP_PROVIDER="arcgis"` those lines have no renderable symbol. The unit tests can't catch it because the SDK is mocked and the assertions read back the same wrong shape (`polyline.symbol.data.data.symbolLayers`). While you're in there: the `CIMVectorMarker` layer omits `size` and `frame`, both of which the same typings mark as required and which Esri's CIM marker samples always set.

https://github.com/OneBusAway/wayfinder/blob/8654a01a6192c4f09d7d4238e41f04d7448ddfa6/src/lib/Provider/ArcGISMapProvider.svelte.js#L695-L704

soumajitgh:
Fixed the CIM symbol issue in 9eccd05, including arrow size/frame, direction, and opacity. Also fixed Web Mercator bounds conversion and a late-request teardown race. Regression tests now exercise real SDK symbols and geometry.

Validated in Chrome; all 1,829 tests and GitHub lint, test, and build checks pass.

soumajitgh:

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Every blocking item from both rounds is genuinely fixed, and the ones I most doubted are now verifiable rather than self-confirming.

The CIM symbol work is the clearest example. It's `data.symbol` rather than the nested `data.data`, the `CIMVectorMarker` has the `size` and `frame` it needs, the arrow ring points forward with `angleToLine: true`, and layer order puts the arrow above the stroke. More to the point, the tests now import `getCIMSymbolColor`/`getCIMSymbolSize` from the real `cimSymbolUtils` unmocked, so they're asserting against Esri's actual symbol handling instead of against a mock that agrees with the implementation by construction. That's the difference between a test that can fail and one that can't.

`updateMarkersRouteLabelVisibility()` is wired into the shared extent/zoom/stationary watch with the same short-circuit as the other two providers. The `setPadding` leak is closed properly - `fitToPolylines` snapshots `{...view.padding}` and restores in `finally`, and the `setPadding` call moved below the empty-polylines bail so a no-op fit stops mutating view state. The overlapping-fit tests in both directions are the right coverage for that.

I checked the bundle question myself rather than trusting the description, because a static import in the factory would have pulled the SDK into every deployment's main chunk. `ArcGISMapProvider.svelte.js` has zero static `@arcgis` imports and 19 dynamic ones; the only two static `@arcgis` imports in the diff are in the test file. So `createMapProvider` importing the provider statically is fine - OSM and Google deployments pay for the provider module and nothing more.

The cross-provider changes are the part that affects existing deployments, so I read those too. Making `createPolyline` synchronous by hoisting `importLibrary('geometry')` into `initMap` is safe for existing callers - awaiting a non-promise is a no-op - and `unmount` is imported at `GoogleMapProvider.svelte.js:18`, so the new `destroy()` won't throw on the component teardown path.

Things I'm taking as follow-ups rather than holding this on:

- **Right-click menu.** `enableContextMenu()` uses `immediate-click` with `event.button === 2` but nothing calls `preventDefault` on `contextmenu`. Leaflet and Google both suppress the native menu implicitly; ArcGIS doesn't. Under `arcgis` the browser menu probably appears over our popup. Esri's own guidance is a `contextmenu` listener on `view.container`.
- **Dark theme.** Only the ArcGIS light theme CSS is imported, so the zoom widget and attribution stay light in dark mode. Parity gap against the other two providers.
- **Pinned dependency.** `@arcgis/core` is pinned to exactly `4.33.14` where everything else uses `^`, and it brings ~50 transitive packages into the lockfile. All behind the dynamic import and out of the main chunk, but it's in every install and every cold CI cache now.

Nice piece of work, and the provider contract is cleaner across all three implementations than it was before this.

PR #599, opened 2026-08-26
test: add unit tests for skipSurvey()

### LINKED ISSUE #444: Add unit tests for skipSurvey() function

omlahore:
Closes #444

`skipSurvey()` had no test coverage. #420 fixed recurring surveys being permanently skipped instead of skipped for one week, and nothing currently pins that behaviour.

Three cases, matching the issue body:

- **Recurring survey** (`allows_multiple_responses: true, always_visible: true`) sets only `survey_<id>_skipped_timestamp`, and `survey_<id>_skipped` stays absent.
- **One-time survey** sets only `survey_<id>_skipped`, and no timestamp.
- **Stale flag cleanup**: a recurring survey that already carries a `_skipped` flag from the old buggy behaviour has it removed.

Two notes on the approach:

Fake timers pin `Date.now()` so the timestamp assertion is an exact value rather than a not-null check, which would still pass if the code wrote something wrong.

I checked the third test actually earns its place: dropping the `localStorage.removeItem` line from `skipSurvey` fails it and leaves the other 23 green.

Test-only, no production changes. `prettier --check` and `eslint` both clean. 24 tests pass in `surveysUtils.test.js`, up from 21.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for skipping recurring and one-time surveys.
  * Verified timestamp-based skipping, permanent skip flags, and cleanup of stale skip settings.

### DIFF
--- a/src/tests/lib/surveysUtils.test.js
+++ b/src/tests/lib/surveysUtils.test.js
@@ -15,7 +15,8 @@ beforeEach(() => {
 	vi.stubGlobal('localStorage', {
 		getItem: vi.fn((key) => store[key] || null),
 		setItem: vi.fn((key, value) => {
-			store[key] = value;
+			// Storage coerces on write, so getItem always hands back a string.
+			store[key] = String(value);
 		}),
 		removeItem: vi.fn((key) => {
 			delete store[key];
@@ -337,7 +338,7 @@ describe('skipSurvey', () => {
 
 		skipSurvey(survey);
 
-		expect(localStorage.getItem('survey_1_skipped_timestamp')).toBe(NOW);
+		expect(localStorage.getItem('survey_1_skipped_timestamp')).toBe(String(NOW));
 		expect(localStorage.getItem('survey_1_skipped')).toBeNull();
 	});
 
@@ -346,17 +347,17 @@ describe('skipSurvey', () => {
 
 		skipSurvey(survey);
 
-		expect(localStorage.getItem('survey_2_skipped')).toBe(true);
+		expect(localStorage.getItem('survey_2_skipped')).toBe('true');
 		expect(localStorage.getItem('survey_2_skipped_timestamp')).toBeNull();
 	});
 
 	it('should clear a stale permanent flag when skipping a recurring survey', () => {
 		const survey = { id: 3, allows_multiple_responses: true, always_visible: true };
-		localStorage.setItem('survey_3_skipped', true);
+		localStorage.setItem('survey_3_skipped', 'true');
 
 		skipSurvey(survey);
 
 		expect(localStorage.getItem('survey_3_skipped')).toBeNull();
-		expect(localStorage.getItem('survey_3_skipped_timestamp')).toBe(NOW);
+		expect(localStorage.getItem('survey_3_skipped_timestamp')).toBe(String(NOW));
 	});
 });
--- a/src/tests/lib/surveysUtils.test.js
+++ b/src/tests/lib/surveysUtils.test.js
@@ -6,7 +6,8 @@ import {
 	getMapSurvey,
 	submitHeroQuestion,
 	updateSurveyResponse,
-	getPrioritySurvey
+	getPrioritySurvey,
+	skipSurvey
 } from '../../lib/Surveys/surveyUtils';
 
 beforeEach(() => {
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new assertions pin values that real `localStorage` can never return, so they lock in a bug in the test file's mock rather than the browser contract. `skipSurvey()` calls `setItem(key, now)` and `setItem(key, true)`, and a real `Storage` coerces both to strings — `"1787700000000"` and `"true"`. These tests only pass because this file's `beforeEach` stubs `localStorage` with a plain object that skips that coercion, so `toBe(NOW)` (number) and `toBe(true)` (boolean) succeed here and would fail in a browser. I verified the coupling by mutation: changing `surveyUtils.js` to the browser-faithful `String(now)` / `'true'` breaks these tests even though behavior is unchanged (`getValidSurveys` already does `parseInt(..., 10)` and `!!`), so the suite now blocks the correct fix. Note also line 355 seeds a boolean `true` while every other seed in this file uses the string `'true'` (lines 55-56, 259, 265, 281, 288). This is the same point raised on #447, where the fix was to make the mock coerce to strings and assert `'true'` / `String(NOW)`. The rest of the block is solid — deleting the `removeItem` line or restoring the pre-#420 double-write both fail these tests, so they do earn their place.

https://github.com/OneBusAway/wayfinder/blob/c34aed8f5e71259bc1488a9a7491908e2c7b0f1e/src/tests/lib/surveysUtils.test.js#L339-L350

src/tests/lib/surveysUtils.test.js
```diff
@@ -318,3 +319,44 @@ describe('Survey Visibility and Multiple Responses', () => {
...
+	afterEach(() => {
+		vi.useRealTimers();
+	});
+
+	it('should only set the timestamp when skipping a recurring survey', () => {
+		const survey = { id: 1, allows_multiple_responses: true, always_visible: true };
+
+		skipSurvey(survey);
+
+		expect(localStorage.getItem('survey_1_skipped_timestamp')).toBe(NOW);
```
aaronbrethorst: `localStorage.getItem()` returns a **string** in a browser — `Storage` coerces on write. This passes only because the `beforeEach` stub at the top of the file stores raw values (`store[key] = value`) without coercing.

Real behavior is `"1787700000000"`, not `1787700000000`.

The sharp edge: `surveyUtils.js` writes `setItem(key, now)` today. If anyone corrects that to `String(now)` — which is what actually reaches storage — this assertion goes red on *correct* code.

src/tests/lib/surveysUtils.test.js
```diff
@@ -318,3 +319,44 @@ describe('Survey Visibility and Multiple Responses', () => {
...
+		expect(localStorage.getItem('survey_1_skipped')).toBeNull();
+	});
+
+	it('should only set the permanent flag when skipping a one-time survey', () => {
+		const survey = { id: 2, allows_multiple_responses: false, always_visible: false };
+
+		skipSurvey(survey);
+
+		expect(localStorage.getItem('survey_2_skipped')).toBe(true);
```
aaronbrethorst: Same as above: a browser returns `'true'` here, not the boolean `true`.

Note that line 56 of this same file already seeds the string form (`localStorage.setItem('survey_2_skipped', 'true')`) — that's the convention to match.

src/tests/lib/surveysUtils.test.js
```diff
@@ -318,3 +319,44 @@ describe('Survey Visibility and Multiple Responses', () => {
...
+		expect(localStorage.getItem('survey_2_skipped_timestamp')).toBeNull();
+	});
+
+	it('should clear a stale permanent flag when skipping a recurring survey', () => {
+		const survey = { id: 3, allows_multiple_responses: true, always_visible: true };
+		localStorage.setItem('survey_3_skipped', true);
```
aaronbrethorst: Minor, same family: seed this as `'true'` to match line 56 and what a browser would actually have stored.

aaronbrethorst (changes requested):
Thanks for taking this on — `skipSurvey()` genuinely was an untested branch, and the three cases you picked are the right three. The recurring/one-time split is exactly where the #420 bug lived, and the third test (clearing a stale permanent flag) pins the specific thing that was broken. Using `vi.useFakeTimers()` + `setSystemTime` to assert an exact timestamp is the right instinct rather than a loose `expect.any(Number)`.

One thing needs fixing before this can land, and it's the same point that sent #447 back.

## The assertions encode the mock's behavior, not the browser's

`localStorage.getItem()` always returns a **string** — `Storage` coerces on write. This file's `beforeEach` stubs it with a plain object that stores values as-is, so `getItem()` hands back the raw number and boolean you put in, and `toBe(NOW)` / `toBe(true)` pass. In a browser those same calls return `"1787700000000"` and `"true"`, and both assertions fail.

That matters more than a fidelity nitpick, because of which direction it breaks. `surveyUtils.js` currently writes `localStorage.setItem(key, now)` and `setItem(key, true)` — raw non-strings. If someone later corrects that to the browser-faithful `String(now)` / `'true'`, **your tests go red on correct code.** A test suite that blocks the right fix is worse than no coverage on that branch.

The fix is small, and this file already shows the way — line 56 seeds `localStorage.setItem('survey_2_skipped', 'true')` with a string. Either:

1. Make the `beforeEach` stub coerce like the real thing (`store[key] = String(value)`) and assert `String(NOW)` and `'true'`, or
2. Assert on the mock call instead: `expect(localStorage.setItem).toHaveBeenCalledWith('survey_1_skipped_timestamp', NOW)`, which pins what the code *writes* and sidesteps the read-path fidelity question entirely.

I'd go with (1) — it makes every future test in this file honest, not just yours.

While you're in there, line 355 seeds a boolean `true` where the rest of the file uses `'true'`; same fix.

## One process note

CI hasn't actually run on this PR. The Lint and Test workflows are sitting at `action_required` because fork PRs need a maintainer to approve the run — that's on me, not you, and I've approved them now. Worth re-checking them once you push.

Fix the assertions and I'll merge this. The test structure underneath is good — it's just pointed at the wrong contract.

omlahore:
Done, went with (1). The stub coerces with `String(value)` now, the assertions compare `String(NOW)` and `'true'`, and line 355 seeds `'true'` to match the rest of the file.

I checked the direction you flagged rather than assuming it. With the coercing stub in place I patched `surveyUtils.js` to write `String(now)` and `'true'`, ran the file, and the tests still pass, then reverted the source. So they no longer go red on that fix.

Full suite is green: 103 files, 1797 tests. Rebased on develop, so the runs you approved should pick up the new head.

aaronbrethorst (approved):
The string-contract fix is exactly right. Coercing the stub with `String(value)` and asserting `String(NOW)` / `'true'` means these tests now pin what real `localStorage` can actually return, so they'll stay honest if `skipSurvey()` is ever corrected to write strings itself. I re-checked all three against the implementation and every expectation matches, and the stale-flag test genuinely fails if the `removeItem` line is deleted — it's a real guard, not a tautology. Nice, tight piece of testing.

PR #601, opened 2026-09-01
Fix vehicle marker contrast on route lines

### LINKED ISSUE #600: Vehicle marker invisible when route color matches vehicle color
## Description
When the route color is the same as the vehicle color, the vehicle marker becomes invisible or indistinguishable on the map, making it difficult for users to identify the vehicle position along the route.

## Expected Behavior
The vehicle marker should be clearly visible and distinguishable from the route line, regardless of whether the colors match.

## Current Behavior
When route color equals vehicle color, the vehicle marker blends into the route line and becomes invisible to the user.

## Suggested Solution
- Use a distinct color for the vehicle marker based on contrast/accessibility standards
- Add an outline or border to the vehicle marker to distinguish it from the route
- Implement a dynamic vehicle color selection that ensures visibility against the route color

## Screenshots

Image

soumajitgh:
## Summary

- add a contrast-selected black or white backing behind vehicle markers
- outline directional arrows so they remain visible over matching route lines
- cover dark, light, and same-colour marker scenarios with regression tests

## Testing

- npm run lint
- all 1,796 tests passed; the four metrics tests were rerun with local port binding enabled

## Screenshots

image

Closes #600

## Summary by CodeRabbit

* **New Features**
  * Routes now update their colors immediately when route colors or map themes change.
  * Existing vehicle markers refresh automatically when switching between light and dark map themes.
  * Vehicle markers use contrasting outlines, backing circles, and arrow styling for improved visibility across route colors and map themes.

* **Bug Fixes**
  * Improved directional arrow and marker legibility on dark maps and dark-colored routes.
  * Cached vehicle data now refreshes with updated route colors without requiring another network request.

### DIFF
--- a/src/lib/MapHelpers/generateVehicleIcon.js
+++ b/src/lib/MapHelpers/generateVehicleIcon.js
@@ -1,4 +1,5 @@
 import { env } from '$env/dynamic/public';
+import { contrastRatio } from '$lib/colorUtils.js';
 import { toDirection } from '$lib/mathUtils';
 import { generateRouteTypeSvgForDisplay, RouteType } from '$config/routeConfig';
 
@@ -28,6 +29,19 @@ function getDirectionFromOrientation(orientation) {
 // match it to their theme instead of the hardcoded amber.
 const HIGHLIGHT_GLOW_COLOR = env.PUBLIC_COLOR_VEHICLE_HIGHLIGHT || '#FACC15';
 
+/**
+ * Returns the neutral colour that has the strongest contrast against the
+ * vehicle/route colour. The vehicle marker deliberately keeps the route colour
+ * for recognition, but this backing prevents it from disappearing into a
+ * same-colour route polyline.
+ *
+ * @param {string} color
+ * @returns {'#ffffff' | '#000000'}
+ */
+function getVehicleMarkerContrastColor(color) {
+	return contrastRatio(color, '#ffffff') >= contrastRatio(color, '#000000') ? '#ffffff' : '#000000';
+}
+
 function createVehicleIconSvg(
 	orientation,
 	color = '#007BFF',
@@ -36,10 +50,16 @@ function createVehicleIconSvg(
 ) {
 	const direction = getDirectionFromOrientation(toDirection(orientation));
 	const angle = DIRECTIONS.find((d) => d.icon === direction).angle;
+	const contrastColor = getVehicleMarkerContrastColor(color);
 
+	// Draw the directional arrow twice: a broad neutral stroke first, then the
+	// route-coloured arrow. This makes the direction indicator legible even
+	// when it lies directly over a route of the same colour.
 	const arrowPath = `
-    <line x1="0" y1="0" x2="0" y2="-15" stroke-width="2" transform="rotate(${angle})"/>
-    <polygon points="0,-25 5,-15 -5, -15" stroke="white" stroke-width="1" transform="rotate(${angle})"/>
+    <line x1="0" y1="0" x2="0" y2="-15" stroke="${contrastColor}" stroke-width="6" stroke-linecap="round" transform="rotate(${angle})"/>
+    <polygon points="0,-25 5,-15 -5,-15" fill="${contrastColor}" stroke="${contrastColor}" stroke-width="4" stroke-linejoin="round" transform="rotate(${angle})"/>
+    <line x1="0" y1="0" x2="0" y2="-15" stroke="${color}" stroke-width="2" stroke-linecap="round" transform="rotate(${angle})"/>
+    <polygon points="0,-25 5,-15 -5,-15" fill="${color}" stroke="${color}" stroke-width="1" stroke-linejoin="round" transform="rotate(${angle})"/>
 `;
 
 	// A soft blurred halo behind the marker for the selected trip. Drawn first so
@@ -59,15 +79,18 @@ function createVehicleIconSvg(
 	return `
         <svg width="${iconWidth}" height="${iconHeight}" viewBox="-28 -28 56 56" xmlns="http://www.w3.org/2000/svg">
             ${highlightDefs}
-            <g stroke="${color}" fill="${color}">
-                <!-- Highlight glow for the selected vehicle (behind everything) -->
-                ${highlightGlow}
+            <!-- Highlight glow for the selected vehicle (behind everything) -->
+            ${highlightGlow}
 
+            <!-- Mask the route under the vehicle with a contrasting backing. -->
+            <circle cx="0" cy="0" r="16" fill="${contrastColor}"/>
+
+            <g stroke="${color}" fill="${color}">
                 <!-- Directional arrow -->
                 ${arrowPath}
 
                 <!-- Circle background -->
-                <circle cx="0" cy="0" r="13" stroke-width="2" fill="white"/>
+                <circle cx="0" cy="0" r="13" stroke-width="2" fill="${contrastColor}"/>
 
                 <!-- vehicle icon inside the circle -->
                 ${vehicleSvg}
@@ -75,4 +98,4 @@ function createVehicleIconSvg(
         </svg>`;
 }
 
-export { createVehicleIconSvg, iconWidth, iconHeight };
+export { createVehicleIconSvg, getVehicleMarkerContrastColor, iconWidth, iconHeight };
--- a/src/lib/MapHelpers/generateVehicleIcon.js
+++ b/src/lib/MapHelpers/generateVehicleIcon.js
@@ -30,27 +30,47 @@ function getDirectionFromOrientation(orientation) {
 const HIGHLIGHT_GLOW_COLOR = env.PUBLIC_COLOR_VEHICLE_HIGHLIGHT || '#FACC15';
 
 /**
- * Returns the neutral colour that has the strongest contrast against the
- * vehicle/route colour. The vehicle marker deliberately keeps the route colour
- * for recognition, but this backing prevents it from disappearing into a
- * same-colour route polyline.
+ * Returns the neutral colour with the strongest worst-case contrast against
+ * both the vehicle/route colour and the current basemap. The vehicle marker
+ * deliberately keeps the route colour for recognition, but this backing keeps
+ * it visible where it crosses its route without letting it sink into a dark
+ * basemap.
  *
  * @param {string} color
+ * @param {boolean} [dark=false]
  * @returns {'#ffffff' | '#000000'}
  */
-function getVehicleMarkerContrastColor(color) {
-	return contrastRatio(color, '#ffffff') >= contrastRatio(color, '#000000') ? '#ffffff' : '#000000';
+function getVehicleMarkerContrastColor(color, dark = false) {
+	// These are representative neutral basemap colours for the two map themes.
+	// Score each candidate by its weaker contrast: a backing that only contrasts
+	// with the route but blends into the map is not useful.
+	const basemapColor = dark ? '#1a1a1a' : '#ffffff';
+	const contrastScore = (candidate) =>
+		Math.min(contrastRatio(color, candidate), contrastRatio(basemapColor, candidate));
+
+	return contrastScore('#ffffff') >= contrastScore('#000000') ? '#ffffff' : '#000000';
 }
 
+/**
+ * Creates the SVG used for a live vehicle marker.
+ *
+ * @param {number} orientation
+ * @param {string} [color='#007BFF']
+ * @param {number} [routeType=RouteType.BUS]
+ * @param {boolean} [highlighted=false]
+ * @param {boolean} [dark=false]
+ * @returns {string}
+ */
 function createVehicleIconSvg(
 	orientation,
 	color = '#007BFF',
 	routeType = RouteType.BUS,
-	highlighted = false
+	highlighted = false,
+	dark = false
 ) {
 	const direction = getDirectionFromOrientation(toDirection(orientation));
 	const angle = DIRECTIONS.find((d) => d.icon === direction).angle;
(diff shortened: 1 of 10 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The backing/outline color is chosen only against the route color, never against the basemap, so it regresses dark mode. `RouteMap.svelte` passes the vehicle the same `mapContrastColor(route?.color, { dark })` value it uses for the polyline, and in dark mode that lightens dark route colors (e.g. `#0a4ea2` -> `#85a7d1`, `#007BFF` -> `#59a9ff`, and the default no-color case too). Every one of those lightened colors has relative luminance above the ~0.179 crossover, so `getVehicleMarkerContrastColor` returns `#000000` for essentially every route in dark mode. The result is a black r=16 body disc and a black arrow outline (`stroke-width="6"` / `stroke-width="4"`) drawn on the dark basemap — openfreemap's `dark` style in `OpenStreetMapProvider.setTheme`, `nightModeStyles()` in the Google provider — roughly 1.4:1 against the map, where the previous hardcoded `fill="white"` circle and `stroke="white"` arrowhead were ~14.8:1. The bright ring and glyph keep the marker from vanishing outright, but the silhouette and the direction arrow's outline both sink into the map, which is the same failure mode the PR set out to fix, just moved from the route line to the basemap. Worth constraining the pick (e.g. only invert when the route color is genuinely close to the marker body, or factor the current theme into the choice) and adding a dark-mode case to the new tests.

https://github.com/OneBusAway/wayfinder/blob/0eec0c61b1fd33ff01d84e022b2bdd16d17cd332/src/lib/MapHelpers/generateVehicleIcon.js#L40-L44

src/lib/MapHelpers/generateVehicleIcon.js
```diff
@@ -59,20 +79,23 @@ function createVehicleIconSvg(
...
         <svg width="${iconWidth}" height="${iconHeight}" viewBox="-28 -28 56 56" xmlns="http://www.w3.org/2000/svg">
             ${highlightDefs}
-            <g stroke="${color}" fill="${color}">
-                <!-- Highlight glow for the selected vehicle (behind everything) -->
-                ${highlightGlow}
+            <!-- Highlight glow for the selected vehicle (behind everything) -->
+            ${highlightGlow}
 
+            <!-- Mask the route under the vehicle with a contrasting backing. -->
+            <circle cx="0" cy="0" r="16" fill="${contrastColor}"/>
```
aaronbrethorst: This is the line that turns dark mode black-on-black.

`contrastColor` is derived from `color` alone, but by the time `color` gets here in dark mode it has already been lightened by `mapContrastColor(route?.color, { dark })` in `RouteMap.svelte:90`. A lightened color always sits above the ~0.179 luminance crossover, so `getVehicleMarkerContrastColor` returns `#000000` for essentially every route:

| route color | dark-mode color | backing |
|---|---|---|
| `#000000` | `#808080` | `#000000` |
| `#007BFF` | `#59A9FF` | `#000000` |

Result: an opaque black disc on the dark basemap (~1.4:1), where `develop` renders white (~14.8:1).

The backing needs to know about the basemap, not just the route color. `RouteMap.svelte` already has `dark` in scope.

src/lib/MapHelpers/generateVehicleIcon.js
```diff
@@ -59,20 +79,23 @@ function createVehicleIconSvg(
...
+
+            <g stroke="${color}" fill="${color}">
                 <!-- Directional arrow -->
                 ${arrowPath}
 
                 <!-- Circle background -->
-                <circle cx="0" cy="0" r="13" stroke-width="2" fill="white"/>
+                <circle cx="0" cy="0" r="13" stroke-width="2" fill="${contrastColor}"/>
```
aaronbrethorst, line 110: Same root cause as the `r=16` disc above — this was `fill="white"` and is now theme-blind.

Separately: this `r=13` circle and the `r=16` disc are now filled with the same color, so the "backing" and the "circle background" render as one flat disc. Worth a look at whether you still need both.

aaronbrethorst (changes requested):
This is a real problem worth fixing and the approach is sound — a vehicle marker that vanishes into a same-colored route line is exactly the kind of thing #600 describes. Drawing the arrow twice (broad neutral stroke, then the route-colored arrow on top) is the right technique, and `getVehicleMarkerContrastColor` picking by measured `contrastRatio` rather than a brightness guess is better than what I'd have expected. It also fixes a pre-existing legibility bug for free: a yellow `#FFFF00` glyph on the old hardcoded white circle was about 1.07:1.

There's one thing I need fixed before this lands, and the screenshot wouldn't show it because it's light mode.

## In dark mode this turns essentially every marker into a black disc on a black map

The backing color is chosen against the route color alone, with no knowledge of the basemap. That's fine in isolation — but the route color reaching this function has *already* been lightened for dark mode upstream:

`src/components/map/RouteMap.svelte:89-90` computes `mapContrastColor(route?.color, { dark })`, and in dark mode `adjustColorForDarkMode` mixes dark colors toward white (very-dark → 50%, dark → 35%, somewhat-dark → 20%).

So the input to `getVehicleMarkerContrastColor` in dark mode is always a *lightened* color, which lands above the ~0.179 luminance crossover, which returns `#000000`. I worked two through the real functions:

- `#000000` → dark mode `#808080` → luminance 0.216 → contrast vs white 3.95, vs black 5.32 → **black backing**
- `#007BFF` (the default) → dark mode `#59A9FF` → luminance 0.416 → contrast vs white 2.25, vs black 9.33 → **black backing**

On `develop` both of those render a white `r=13` circle that pops on the dark basemap (`tiles.openfreemap.org/styles/dark`, and `nightModeStyles()` on Google). After this change they're a black disc on a dark map — roughly 1.4:1 against the tile background, versus ~14.8:1 today. The arrow's new outline pass goes black too, so the direction indicator loses the white halo that used to make it readable against anything.

To be fair to the change: the marker doesn't become invisible, because the `r=13` ring stroke and the glyph are drawn in the (now bright) route color. So you get a bright ring on a black disc. But the silhouette and the arrow outline both stop doing their job, and that's a systematic change across every marker on the map rather than an edge case.

The fix is to feed the basemap into the decision — pick the backing that contrasts with *both* the route color and the current theme, rather than the route color alone. `RouteMap.svelte` already knows `dark`, so threading it through to `createVehicleIconSvg` is the natural shape. In practice "white in dark mode unless the route color is itself near-white" would cover it.

## Two smaller things while you're in there

- **The highlight glow gets clipped.** The new opaque `r=16` disc is drawn over the `r=20` blurred `highlightGlow`, cutting the visible halo band from ~7 units to ~4. The comment right above it reasons explicitly about the glow being "sized well beyond the white circle so it's clearly visible" — the circle grew and the `r=20` didn't.
- **Comment rot.** That same comment, and the `<!-- Circle background -->` region, still describe a white circle that's now white only about half the time.

Also worth saying plainly so it's a deliberate choice rather than a surprise: this changes the appearance of *every* marker, not just same-colored ones. Roughly half of common route colors — including the `#007BFF` default and the `#808080` no-realtime-data grey — get a black body in light mode too. I think that's defensible (glyph contrast genuinely improves), but the PR reads as a narrow fix and the screenshot only shows the black-route case, so I want to make sure that's intended.

Sort out the dark-mode case and I'll take this.

aaronbrethorst (changes requested):
The dark-mode fix is correct, and it's the fix I was asking for. Scoring each candidate against both the route color and a representative basemap is the right generalization, `#1a1a1a`/`#ffffff` are reasonable proxies for the actual tiles under both providers, and raising the highlight glow to `r=23` restores the 7px band exactly. The comment rot is cleaned up too. I worked the numbers and both of my examples now flip the right way: `#808080` and `#59A9FF` both get a white backing in dark mode instead of black.

Unfortunately the same fix inverts the bug into light mode, and I need that closed before this lands.

## `Math.min` makes the white backing unreachable in light mode

In light mode `basemapColor` is `#ffffff`, so `contrastRatio(basemapColor, '#ffffff')` is exactly `1`. That means:

```
contrastScore('#ffffff') = min(anything, 1) = 1        // always
contrastScore('#000000') = min(ratio(color, #000), 21) = (L + 0.05) / 0.05
```

The black candidate's score is `>= 1` for every color, and strictly greater for anything with nonzero luminance. So `getVehicleMarkerContrastColor(color, false)` returns `#000000` for every input except pure `#000000`.

That matters because the glyph and arrow are route-colored — `generateRouteTypeSvgForDisplay` emits a bare `<path>` inheriting `fill="${color}"` from the `<g>`, and the arrow's top pass is explicitly `fill="${color}"`. So a dark agency color now sits on a black disc:

| route color | backing on `develop` | backing here | glyph contrast |
| --- | --- | --- | --- |
| `#003366` | white | `#000000` | **1.67:1** (was 12.61:1) |
| `#4B0082` | white | `#000000` | **1.62:1** (was 12.95:1) |
| `#800000` | white | `#000000` | **1.92:1** (was 10.95:1) |
| `#2E2E2E` | white | `#000000` | **1.55:1** (was 13.58:1) |

`mapContrastColor` only darkens colors that are already bright in light mode, so dark navy and maroon reach the marker unmodified. This is a regression against `develop`, not just against your earlier commit — `develop`'s hardcoded `fill="white"` gave `#003366` 12.6:1.

The reason it slipped through: the existing light-mode assertion in `generateVehicleIcon.test.js` uses `#000000`, which is the one input that still returns white.

## The fix

The two things being combined are different questions — "does the backing let the glyph read against the route color" and "does the marker have a silhouette against the map". `Math.min` treats them as interchangeable, and on a white basemap the second one unconditionally vetoes white. Weighting the route-color term, or using the basemap term only as a tiebreaker, keeps the dark-mode fix without the light-mode collapse. Please add a light-mode case for a non-black dark color (`expect(getVehicleMarkerContrastColor('#003366')).toBe('#ffffff')`) — it fails today and would have caught this.

## Two open questions from last time

Still unanswered, and the second one matters more now:

- On `:113`, the `r=13` circle and the `r=16` disc are both `fill="${contrastColor}"`. Having looked again, I think the answer is that the *fill* is redundant but the *stroke* isn't — it inherits `stroke="${color}"` from the enclosing `<g>` and draws the route-colored ring. Worth confirming that's what you intended.
- I asked whether the black body in light mode was deliberate, and said I thought it was defensible because glyph contrast genuinely improves. For the colors in that table it now gets substantially worse, so I'd like to know what you're seeing — if there's a case I'm missing, say so.

The theme-refresh commit is a good addition and I checked it properly: `vehicleMarkers` is kept in sync by `removeVehicleMarker`/`clearVehicleMarkers`, `vehicleIconOptions` is an own property collected with the marker, and OSM correctly runs the refresh before its redundant-style early return. No issues there.

Fix the light-mode scoring and I'll take this.

soumajitgh:
Fixed in ac9abe0. Separated glyph contrast from the dark-mode halo, fixing dark and pale route colors. Black bodies are used only when they improve glyph contrast; the inner circle preserves the route-colored ring and masks the arrow shaft.

Theme switches now recolor route lines, arrows, and cached vehicles immediately without refetching or moving the map. Added regressions for both themes.

Lint, production build, and all 1,817 tests passed (metrics tests rerun with local port access).

src/lib/vehicleUtils.js
```diff
@@ -180,6 +198,7 @@ export async function fetchAndUpdateVehiclesForRoutes(
 					resolveHighlightedTripId(),
 					colorsByRouteId.get(route.id)?.line
 				);
+				lastData.set(route.id, data);
```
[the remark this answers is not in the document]
soumajitgh: Fixed in 4e0b247. Each tick now has a generation check before applying positions, updating the cache, removing markers, or reporting counts. Added regression tests for both completion orders and verified that refresh retains the newest positions. Lint and all 1,819 tests pass.

aaronbrethorst:
### Code review

soumajitgh:

aaronbrethorst (approved):
This is the fix. You didn't patch the light-mode inversion I found last round - you removed the thing that caused it, which is the better move.

Dropping the `Math.min` basemap scoring and taking `getVehicleMarkerContrastColor` back to a pure glyph-contrast pick closes the regression at the source: in light mode `#003366`, `#4B0082`, `#800000` and `#2E2E2E` all get a white backing again, so the 1.6:1 glyph contrast I measured is back to develop's ~12:1. Moving basemap separation into a separate `haloColor` - forced white in dark mode - is the right decomposition, because the two problems genuinely are separate and trying to score them together is what inverted the bug in the first place.

I traced the layering at head. The `r=16` disc takes `haloColor` and the `r=13` disc takes `contrastColor`, so in dark mode with a bright route color you get a white outer ring around a black body: the silhouette survives on dark tiles, which was my original complaint, and the glyph still sits on the backing that contrasts with it. The conditional `arrowHalo` only draws when `haloColor` and `contrastColor` actually differ, so it costs nothing in the cases that don't need it. The glow at `r=23` against the `r=16` disc restores the 7px band exactly.

The tests are what make me comfortable this stays fixed. Parameterizing over the exact four colors from my table and asserting a white backing in light mode means the specific regression I hit has a named guard. The gamut sweep at 4.5:1 and the dark/light silhouette test driven through the real `mapContrastColor` cover the general case rather than just the reported one.

One thing I want on the record as a deliberate choice rather than a surprise, since it's the point I raised in round one and it's still true: this changes the appearance of every marker, not just same-colored ones. `#007BFF` - the default - now renders as a black disc with a blue glyph in light mode where develop drew a white one. Glyph contrast genuinely improves there, 4.02:1 to 5.22:1, and a black disc reads fine on light tiles, so I'm taking it. But it's a visible change across the whole map for every deployment that hasn't set a route color, and if anyone asks why the markers look different, this is why.

PR #602, opened 2026-09-01
Rename Obaco/OneBusAway.co to Sidecar, deprecating the old env vars

aaronbrethorst:
## Summary

- Renames the Obaco / OneBusAway.co service to **Sidecar** throughout. Every pre-rename environment variable keeps working as a deprecated alias, so existing deployments don't break — each warns once at first use naming its replacement.
- Adds `src/lib/sidecarConfig.js` as the single place that resolves Sidecar settings, and points the four alerts/surveys API routes at it instead of reading `$env/dynamic/private` directly.
- Teaches `env-schema.json` / `scripts/validate-env.js` two general fields: `deprecatedNames` and `requiredWith`.

| New | Deprecated alias |
|---|---|
| `PRIVATE_SIDECAR_API_BASE_URL` | `PRIVATE_OBACO_API_BASE_URL` |
| `PRIVATE_SIDECAR_REGION_ID` | `PRIVATE_REGION_ID` |
| `PRIVATE_SIDECAR_SHOW_TEST_ALERTS` | `PRIVATE_OBACO_SHOW_TEST_ALERTS` |

Precedence: a `PRIVATE_SIDECAR_*` variable wins **only when non-empty**. Set to an empty string it counts as unset and the pre-rename name still takes effect — which matters because a half-migrated `.env` hits exactly that case. A stale legacy value that loses to a new name is now reported rather than silently ignored.

## Behavior changes beyond the rename

These came out of review and go slightly past a pure rename — flagging them explicitly:

1. **No more `regions/undefined/` requests.** The region path was previously built at module load with no guard, so a missing region ID produced a real request to `.../regions/undefined/alerts.pb`. `getSidecarRegionPath()` now returns `undefined` and the two region-scoped GET routes degrade instead (alerts → 204, surveys → `{surveys: []}`). The POST survey-response routes are unaffected; they aren't region-scoped.
2. **`.env.example` ships the Sidecar block commented out.** It previously set a base URL with an empty region ID — the exact configuration that silently disabled both features.
3. **`requiredWith` makes "region ID required once a base URL is set" enforceable** rather than prose. This is what lets `validate-env` catch (1) at config time.
4. The region path is now read per request rather than captured at import.

## Test plan

- [x] `npx vitest run` — 1827 passed, 105 files
- [x] `npm run lint` — prettier + eslint clean
- [x] Drove all four endpoints against a stub Sidecar via `vite dev`, in four configurations: new names only, legacy names only, neither set, and both set. Verified the exact upstream URLs (`/api/v1/regions/9/surveys.json?user_id=…`, `/api/v1/regions/9/alerts.pb`), that alerts decodes a real GTFS-RT protobuf and returns the alert, and that with both names set the request goes to the **new** name's region.
- [x] Region-less config issues **zero** upstream requests and logs `Sidecar is disabled: PRIVATE_SIDECAR_REGION_ID is not set` — once per route across repeated requests, not once per map load.
- [x] `validate-env` across new-names / legacy-only / half-migrated / superseded-with-a-bad-value / base-URL-without-region.
- [x] Mutation-tested the new guards: removing the region guard, or restoring `regions/undefined/`, both turn the suite red.

## Review notes

Pre-existing issues in these files that I did **not** change, surfaced by the review — worth separate work:

- `src/routes/api/oba/alerts/+server.js` never checks `response.ok` before feeding the body to the protobuf decoder. A Sidecar outage decodes to an empty feed and returns **204 — indistinguishable from "no alerts right now"**, so riders can miss a real disruption silently. This is the most significant of the three.
- The three survey routes discard the upstream status/body (`throw new Error('Failed to…')`), so a 422, a 401, and a local `TypeError` all reach operators as one generic 500.
- `submit-survey` and `update-survey` return raw `error.message` to the client, which can leak internal detail.

Also: `npm run test` maps to bare `vitest`, which parks in watch mode rather than running once as CLAUDE.md describes — it probably wants `vitest run`. Left alone as out of scope.

https://claude.ai/code/session_01Upr2AgZoRfC3YLjT2bQQPs

## Summary by CodeRabbit

* **New Features**
  * Added Sidecar configuration using the new `PRIVATE_SIDECAR_*` environment variables.
  * Added an optional OBA agency filter setting.
  * Legacy configuration names continue to work as fallbacks with deprecation warnings.
  * Sidecar region settings are required only when a Sidecar API URL is configured.

* **Bug Fixes**
  * Improved alerts and survey behavior when Sidecar configuration is incomplete, including clearer warnings and safe empty responses.

* **Documentation**
  * Updated setup guidance and environment-variable validation documentation for renamed settings and configuration requirements.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,7 +12,7 @@ PRIVATE_OBA_GEOCODER_PROVIDER="google"
 # PRIVATE_OBACO_SHOW_TEST_ALERTS still work, but they are deprecated.
 # PRIVATE_SIDECAR_API_BASE_URL=https://sidecar.onebusaway.org/api/v1
 # PRIVATE_SIDECAR_REGION_ID=
-PRIVATE_SIDECAR_SHOW_TEST_ALERTS=false
+# PRIVATE_SIDECAR_SHOW_TEST_ALERTS=false
 
 # Comma-separated agency IDs to restrict this instance to.
 # Leave empty to show all agencies (default).
--- a/.env.example
+++ b/.env.example
@@ -3,9 +3,16 @@ PRIVATE_OBA_API_KEY="test"
 PRIVATE_OBA_GEOCODER_API_KEY=""
 PRIVATE_OBA_GEOCODER_PROVIDER="google"
 
-PRIVATE_OBACO_API_BASE_URL=https://onebusaway.co/api/v1
-PRIVATE_REGION_ID=
-PRIVATE_OBACO_SHOW_TEST_ALERTS=false
+# Sidecar (formerly Obaco / OneBusAway.co) supplies service alerts and surveys.
+# It is optional; leave these unset to run without alerts and surveys.
+# Set the base URL and the region ID together — a base URL without a region ID
+# disables both features. A PRIVATE_SIDECAR_* value takes effect only when
+# non-empty, so leaving one blank keeps any pre-rename name you already have.
+# The pre-rename names PRIVATE_OBACO_API_BASE_URL, PRIVATE_REGION_ID, and
+# PRIVATE_OBACO_SHOW_TEST_ALERTS still work, but they are deprecated.
+# PRIVATE_SIDECAR_API_BASE_URL=https://sidecar.onebusaway.org/api/v1
+# PRIVATE_SIDECAR_REGION_ID=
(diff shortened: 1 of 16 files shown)


aaronbrethorst:
### Code review

PR #605, opened 2026-09-03
fix(favorites): stop outside-click dismissal stealing focus

### LINKED ISSUE #604: Favorites: outside-click dismissal steals focus from the clicked element
Follow-up from #585, which is otherwise merged and in good shape.

### Outside-click dismissal steals focus from the element the user just clicked

`FavoritesFloatingControl.handleWindowClick` calls `close()` with no arguments, so `restoreFocus` defaults to `true` and this runs:

```js
tick().then(() => toggleBtn?.focus());
```

That's a microtask, so it lands *after* the clicked element has already taken focus. Open the favorites panel, then click the search field next to it on the map page: the panel closes (correct) and the caret is immediately yanked out of the field and onto the favorites star (wrong).

The two deliberate call sites already distinguish this — `handleStopClick` and `handleRouteClick` both pass `restoreFocus: false`, and Escape correctly keeps it `true`. The outside-click path just wasn't separated out.

**The fix is not simply flipping it to `false`.** Most of the map is non-focusable, and clicking there genuinely does need the restore — otherwise focus drops to `<body>` when the panel unmounts, which is worse. What's wanted is roughly "restore focus only if the click target didn't take focus itself":

```js
// in handleWindowClick, for the outside-click branch
close({ restoreFocus: !isFocusable(target) });
```

or equivalently, check `document.activeElement` after the click settles.

**File:** `src/components/favorites/FavoritesFloatingControl.svelte`

### Smaller, while someone is in there

- `FavoritesList.itemTitle` uses `item.shortName || item.id` while `MapExperience.handleFavoriteRouteClick` uses `favorite.shortName ?? removeAgencyPrefix(favorite.id)`. Those disagree on an empty-string `shortName`, and the store's own test asserts empty-string short names are retained, so the path is reachable — just unlikely against real OBA data.
- The 50-item cap is enforced in `add`/`toggle` but not on the localStorage load path. Only reachable by hand-editing storage, so cosmetic.

omlahore:
Fixes #604.

## The change

`handleWindowClick` called `close()` with no arguments, so `restoreFocus` defaulted to `true` and the queued `tick().then(() => toggleBtn?.focus())` ran after the browser had already focused the clicked element.

Taking the approach from the issue rather than flipping the flag to `false`, since most of the map is non-focusable and clicking there does need the restore:

```js
const active = document.activeElement;
close({ restoreFocus: !active || active === document.body || rootEl.contains(active) });
```

The `rootEl.contains(active)` clause is there for the case the issue does not name: if focus is still inside the panel when it closes, skipping the restore strands it once the panel unmounts. That branch keeps the restore.

Escape is untouched, and `handleStopClick` / `handleRouteClick` keep their explicit `restoreFocus: false`.

8 lines.

## Verification

Two tests, one per branch:

- clicking an outside `<input>` closes the panel and **leaves focus on the input**
- clicking an outside non-focusable `<div>` closes the panel and **restores focus to the toggle**

I checked they fail for the right reason rather than passing by accident. With the component change stashed, the first fails and the second still passes, which is exactly the asymmetry the bug describes:

```
× leaves focus on an outside element that took it
✓ restores focus to the toggle when the outside click took none
  Tests  1 failed | 8 passed (9)
```

With the fix, the whole suite is green: **110 files, 1879 tests**. Prettier clean.

## Not included

The two smaller items at the bottom of #604 are left out, since #604 asks for one thing and those are separable:

- `FavoritesList.itemTitle` using `item.shortName || item.id` where `MapExperience.handleFavoriteRouteClick` uses `?? removeAgencyPrefix(...)`. Those disagree only on an empty-string `shortName`.
- The 50-item cap not being enforced on the localStorage load path.

Can take either in a follow-up.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved focus handling when closing the favorites panel by clicking outside.
  - Preserves focus on clicked inputs and other focusable elements.
  - Restores focus to the favorites toggle when the outside click does not move focus.

- **Tests**
  - Added coverage for focus preservation and restoration when the panel closes.

### DIFF
--- a/src/components/favorites/FavoritesFloatingControl.svelte
+++ b/src/components/favorites/FavoritesFloatingControl.svelte
@@ -68,7 +68,14 @@
 		if (!target.isConnected) return;
 
 		if (!rootEl.contains(target)) {
-			close();
+			// The browser focuses a clicked element on mousedown, before this click
+			// reaches the window listener, so restoring focus here would yank the
+			// caret back out of whatever the user just clicked onto. Restore only
+			// when the click did not take focus itself: a click on the map leaves
+			// activeElement at <body>, and focus still inside the closing panel
+			// would be stranded when it unmounts.
+			const active = document.activeElement;
+			close({ restoreFocus: !active || active === document.body || rootEl.contains(active) });
 		}
 	}
 
--- a/src/components/favorites/__tests__/FavoritesFloatingControl.test.js
+++ b/src/components/favorites/__tests__/FavoritesFloatingControl.test.js
@@ -161,6 +161,40 @@ describe('FavoritesFloatingControl', () => {
 		outside.remove();
 	});
 
+	it('leaves focus on an outside element that took it', async () => {
+		render(FavoritesFloatingControl);
+
+		const toggle = screen.getByRole('button', { name: 'Open favorites' });
+		await user.click(toggle);
+		expect(screen.getByRole('dialog')).toBeInTheDocument();
+
+		const input = document.createElement('input');
+		document.body.appendChild(input);
+
+		await user.click(input);
+
+		expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
+		expect(input).toHaveFocus();
+		input.remove();
+	});
+
+	it('restores focus to the toggle when the outside click took none', async () => {
+		render(FavoritesFloatingControl);
+
+		const toggle = screen.getByRole('button', { name: 'Open favorites' });
+		await user.click(toggle);
+		expect(screen.getByRole('dialog')).toBeInTheDocument();
+
+		const outside = document.createElement('div');
+		document.body.appendChild(outside);
+
+		await user.click(outside);
+
+		expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
+		expect(toggle).toHaveFocus();
+		outside.remove();
+	});
+
 	it('closes on Escape and restores focus to the toggle', async () => {
 		render(FavoritesFloatingControl);
 


aaronbrethorst:
### Code review

The fix itself is right, and I verified the test claim rather than taking it on faith. In an isolated copy at this head I reverted only `FavoritesFloatingControl.svelte` and re-ran the file: **1 failed | 8 passed**, and the failure is exactly `leaves focus on an outside element that took it` at `expect(input).toHaveFocus()` (focus had been yanked onto the toggle). With the component change back: 9 passed, and 1036 passed across `src/components/favorites` + `src/tests`. So the new test genuinely exercises the change — that was the sticking point on the last two rounds of #585, and it holds here.

Found 1 issue:

1. The comment's central claim is wrong for the default map provider, and the `active === document.body` clause it justifies is effectively dead on that path. `PUBLIC_OBA_MAP_PROVIDER` defaults to `"osm"` (`.env.example:31`), and `OpenStreetMapProvider.initMap` builds the map with `this.L.map(element, { zoomControl: false })` — `keyboard` is left at Leaflet's default of `true`. Leaflet 1.9.4's `Map.Keyboard.addHooks` then does `if (container.tabIndex <= 0) { container.tabIndex = '0'; }` on `#map` and binds a native `mousedown` listener whose `_onMouseDown` calls `this._map._container.focus()` (`node_modules/leaflet/src/map/handler/Map.Keyboard.js:42-86`). `#map` carries no tabindex of its own (`MapView.svelte:406`), and the favorites wrapper is deliberately `pointer-events` transparent so map clicks reach it. So a click on the map sets `document.activeElement` to the map container — a focusable element outside `rootEl` — not `<body>`.

   Net effect: for the single most common outside click, `restoreFocus` is now `false`, and the panel closes leaving focus on the Leaflet container instead of the toggle. That outcome is defensible on its own (the container is focusable and arrow-key pannable, so focus follows the click and isn't stranded), which is why I'm not calling it a behavior bug. But it is not what the comment, the PR description, or #604's rationale say happens, and the `<body>` branch that the whole design was built around only fires for genuinely non-focusable chrome — e.g. the mobile in-flow area — never for the map. Worth correcting the comment so the next person doesn't reason from it.

https://github.com/OneBusAway/wayfinder/blob/b7edc6f0c9ea67eede29203382698dcd26b43f68/src/components/favorites/FavoritesFloatingControl.svelte#L73-L79

aaronbrethorst (approved):
This is the right fix for #604, and it's the minimal one — reading `document.activeElement` and only restoring when the click didn't take focus itself, rather than adding listeners or reaching for a timeout. It reuses `close({ restoreFocus })`, which already existed for `handleStopClick`, so there's no new API surface, and `<svelte:window>` keeps the teardown burden at zero. I confirmed the new test genuinely fails without the component change.

One correction to the comment, worth folding in whenever you next touch the file. It says a click on the map leaves `activeElement` at `<body>`, but Leaflet's keyboard handler sets `tabIndex = '0'` on the map container and focuses it on mousedown, and we don't pass `keyboard: false` in `OpenStreetMapProvider.initMap()`. So on the default OSM provider a map click lands focus on the container, not `<body>` — `restoreFocus` is `false` there, which is the behavior you want anyway (focus sits on the thing the user just clicked). The code is right; only the stated reason for the `activeElement === document.body` branch is off. Not worth another round.

PR #606, opened 2026-09-03
fix(insights): return 400 for malformed event envelopes

### LINKED ISSUE #603: Analytics: harden forwardEvent envelope handling (null bodies, Umami parity, 400 vs 500)
Follow-ups from the review of #593, which closed items 1 and 3 of #333. Neither is a regression from that PR — both are pre-existing gaps that became visible while reviewing it.

### 1. `envelope = {}` doesn't cover the input the endpoint can actually produce

`PlausibleAdapter.forwardEvent(envelope = {}, requestContext)` uses a default parameter, which only fires on `undefined`. But `/api/events` gets its envelope from `await request.json()`, and `JSON.parse` never returns `undefined` — it returns `null` for a literal `null` body. So:

```
POST /api/events  with body: null
  -> envelope === null
  -> TypeError: Cannot destructure property 'name' of 'envelope' as it is null
  -> 500, with the destructuring message in the response body
```

`envelope ?? {}` closes this and makes the guard cover the reachable case, not just the direct-API-misuse case.

**File:** `src/lib/Insights/adapters/PlausibleAdapter.js`

### 2. `UmamiAdapter` has the same shape and didn't get the same treatment

`src/lib/Insights/adapters/UmamiAdapter.js:86` has the identical `forwardEvent(envelope, requestContext)` signature and the same destructure-then-validate body. Since `createAdapter` switches on `PUBLIC_ANALYTICS_PROVIDER`, the same malformed request now behaves differently depending on which provider a deployment runs. Worth applying whatever we do in (1) to both, and adding the matching no-envelope test.

**File:** `src/lib/Insights/adapters/UmamiAdapter.js`

### 3. Client validation errors surface as HTTP 500

`new Error('forwardEvent requires name and url')` carries no `upstreamStatus`, so the endpoint's `error.upstreamStatus || 500` fallback reports a malformed *client* envelope as a server fault. `insightsError(msg, 400)` from `src/lib/Insights/upstreamError.js` is the existing idiom. Entirely pre-existing — already true today for a body like `{"url": "/x"}` with no `name`.

omlahore:
Fixes #603, all three items.

## 1 and 2. `envelope ?? {}` in both adapters

`PlausibleAdapter.forwardEvent(envelope = {}, ...)` used a default parameter, which only fires on `undefined`. `/api/events` builds the envelope from `await request.json()`, and that returns `null` for a literal `null` body and never `undefined`, so the guard covered direct API misuse but not the reachable case. `UmamiAdapter` had no default at all.

Both now destructure from `envelope ?? {}`.

## 3. `insightsError(msg, 400)` instead of a bare `Error`

`new Error('forwardEvent requires name and url')` carries no `upstreamStatus`, so `+server.js:37`'s `error.upstreamStatus || (isTimeout ? 504 : 500)` reported a malformed client body as a server fault. Both adapters now throw `insightsError(..., 400)`, which is the idiom already used for upstream failures in the same files. `insightsError` was already imported in `UmamiAdapter`; added to `PlausibleAdapter`'s existing import.

Applying it to both matters for the reason the issue gives: `createAdapter` switches on `PUBLIC_ANALYTICS_PROVIDER`, so leaving one behind means the same malformed request behaves differently depending on which provider a deployment runs.

## Verification

Four tests, two per adapter: a `null` envelope and a missing one, each asserting both the message and `upstreamStatus === 400`.

I checked they fail for the right reason rather than passing by accident. With both adapter changes stashed:

```
× PlausibleAdapter > rejects a null envelope as a 400 rather than a destructuring TypeError
× PlausibleAdapter > marks a missing name as a client error
× UmamiAdapter     > rejects a null envelope as a 400 rather than a destructuring TypeError
× UmamiAdapter     > rejects a missing envelope as a 400
  Tests  4 failed | 64 passed (68)
```

With the fix, the whole suite is green: **110 files, 1881 tests**. Prettier clean.

The existing `throws when name is missing` / `throws when url is missing` tests are untouched and still pass, since the message is unchanged and only the carried status is new.

### DIFF
--- a/src/lib/Insights/adapters/PlausibleAdapter.js
+++ b/src/lib/Insights/adapters/PlausibleAdapter.js
@@ -1,4 +1,4 @@
-import { upstreamError } from '../upstreamError.js';
+import { insightsError, upstreamError } from '../upstreamError.js';
 
 const UPSTREAM_TIMEOUT_MS = 5000;
 
@@ -25,17 +25,20 @@ export class PlausibleAdapter {
 		return `${this.env.PUBLIC_ANALYTICS_API_HOST}/api/event`;
 	}
 
-	async forwardEvent(envelope = {}, requestContext) {
+	async forwardEvent(envelope, requestContext) {
 		// Disabled analytics intentionally ignores malformed events so local development
 		// and tests without a complete analytics configuration remain quiet.
 		if (!this.isEnabled()) {
 			return { status: 'analytics disabled' };
 		}
 
-		const { name, url, referrer = '', props = {} } = envelope;
+		// `??` rather than a default parameter: the endpoint gets this from
+		// `await request.json()`, which returns null for a literal null body and
+		// never undefined, so a default parameter never fires on the reachable case.
+		const { name, url, referrer = '', props = {} } = envelope ?? {};
 
 		if (!name || !url) {
-			throw new Error('forwardEvent requires name and url');
+			throw insightsError('forwardEvent requires name and url', 400);
 		}
 
 		const headers = { 'Content-Type': 'application/json' };
--- a/src/lib/Insights/adapters/UmamiAdapter.js
+++ b/src/lib/Insights/adapters/UmamiAdapter.js
@@ -96,10 +96,10 @@ export class UmamiAdapter {
 			language = '',
 			screen = '',
 			props = {}
-		} = envelope;
+		} = envelope ?? {};
 
 		if (!name || !url) {
-			throw new Error('forwardEvent requires name and url');
+			throw insightsError('forwardEvent requires name and url', 400);
 		}
 
 		const body = {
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This does exactly what #603 asked for, on both adapters, and the reasoning in the comment is the part I appreciated most — `request.json()` returning `null` rather than `undefined` is precisely why the default parameter never fired, and `?? {}` is the right answer rather than a defensive shrug. I checked that `400` survives the `error.upstreamStatus || (isTimeout ? 504 : 500)` mapping in the endpoint and doesn't collide with the timeout path, and that the guard throws before the `requestContext.userAgent` deref, so the zero-arg tests still exercise validation rather than a TypeError. The new tests fail for the right reason without the fix. Merging.

PR #608, opened 2026-09-04
fix(map): recover missing route shapes

### LINKED ISSUE #607: Some routes are not drawn in `onebusaway-mcp-ui` but are visible in WayFinder
While working with `onebusaway-mcp-ui`, I noticed that some routes are drawn on the map but are not displayed in WayFinder.

This suggests there may be a difference in how WayFinder processes or renders certain route shapes compared to `onebusaway-mcp-ui`.

### Screenshot

The screenshot below shows an example of the difference:

Image

soumajitgh:
## Summary

- align map routes with the non-departed arrivals shown to riders
- try up to three upcoming trip shapes before falling back to route-level geometry
- cache successful shapes and briefly cache failures to avoid request amplification
- keep arrival polling from redrawing healthy routes and clean up fallback segments and vehicles
- add regression coverage for retries, fallback deduplication, caching, and redraw behavior

## Testing

- npm test -- --run (1,799 tests passed)
- Prettier and ESLint checks on changed files
- npm run build

Closes #607

## Summary by CodeRabbit

* **Improvements**
  * Map route displays now prioritize currently boardable arrivals and exclude trips that have already departed.
  * Route lines are more resilient when trip-specific geometry is unavailable, using alternate trips or route-level geometry when possible.
  * Multiple route segments can now be displayed for more complete coverage.
  * Map updates reuse previously loaded route geometry, improving redraw behavior.
  * Vehicle markers are cleaned up reliably when routes are removed from the map.

### DIFF
--- a/src/components/MapExperience.svelte
+++ b/src/components/MapExperience.svelte
@@ -109,7 +109,11 @@
 	let arrivalsMatchSelection = $derived(
 		stopArrivals?.data?.entry?.stopId != null && stopArrivals.data.entry.stopId === selectedStopId
 	);
-	let activeRoutes = $derived(arrivalsMatchSelection ? activeRoutesFromArrivals(stopArrivals) : []);
+	// Use the same non-departed arrival set the sheet renders. Keeping the raw
+	// response as the bound value is still useful to StopPane; the map derives a
+	// boardable subset whenever that response refreshes.
+	let boardableRoutes = $derived(activeRoutesFromArrivals(stopArrivals, { now: Date.now() }));
+	let activeRoutes = $derived(arrivalsMatchSelection ? boardableRoutes : []);
 	// Deliberately NOT gated on arrivalsMatchSelection (derived from stopArrivals
 	// directly, not from the gated activeRoutes above). StopPane's rendered rows are
 	// a one-time $state seed that doesn't react to stopArrivals going stale, so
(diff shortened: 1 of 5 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a careful piece of work on a genuinely awkward problem. The escalation — try up to three boardable trips in arrival order, then fall back to route-level geometry, rather than dropping the route off the map — is the right shape, and I like that the fallback is honest about being degraded rather than pretending it's the real trip shape.

The parts I checked hardest and found sound:

- The `polylinesByRouteId` single-to-array conversion is complete. `teardown()`, `reassertLinePaintOrder()`, the promotion effect, and `drawRoutes` all handle the array form, and both providers already accept `revealPolylines({only})` with multiple polylines and per-polyline `removePolyline`/`setPolylineLayer`, so nothing diverges between OSM and Google.
- `polledRouteIds` is the right fix for the actual leak: vehicle polling starts for every active route including ones whose shape never resolves, so teardown has to snapshot the union rather than just the drawn keys.
- The stale-token path inside the multi-segment loop correctly removes every segment the superseded route already attached, not just the one in flight.
- `filterDeparted(arrivals, now)` matches its real signature, and gating on `Number.isFinite(now)` keeps the no-options callers on the old unfiltered behavior.

One thing worth a follow-up, not a blocker. `routeColors` now derives from the departed-filtered `boardableRoutes`, but `StopPane` seeds its rows from the *unfiltered* entry on first render and only applies `filterDeparted` on subsequent polls. So a route whose only cold-load arrival has already departed renders a row with no assigned color and falls back to `route?.color` until the first poll corrects it. Cosmetic and self-healing, and the comment above `routeColors` about two consumers with different staleness semantics is now carrying a third axis it doesn't mention — worth a line when you're next in there.

Merging.

PR #612, opened 2026-09-07
fix(stops-for-location): send the longitude span as lonSpan

omlahore:
### What this fixes

`/api/oba/stops-for-location` forwards the client's `lngSpan` straight through to the SDK ([+server.js#L8-L18](https://github.com/OneBusAway/wayfinder/blob/6c5adb0b8f04305cf209fcf2e6ea6a9dd36b4a52/src/routes/api/oba/stops-for-location/%2Bserver.js#L8-L18)):

```js
const lngSpan = +url.searchParams.get('lngSpan');
const queryParams = { lat, lon: lng, latSpan, lngSpan, radius };
```

The SDK's parameter is `lonSpan`, declared in `onebusaway-sdk/resources/stops-for-location.d.ts`:

```ts
latSpan?: number;
...
lonSpan?: number;
```

`lngSpan` is not a parameter it knows, so the longitude span was dropped and the request fell back to the default window.

### How it shows up

`MapView` sends a real span computed from the visible bounding box ([MapView.svelte#L192](https://github.com/OneBusAway/wayfinder/blob/6c5adb0b8f04305cf209fcf2e6ea6a9dd36b4a52/src/components/map/MapView.svelte#L192)):

```js
`/api/oba/stops-for-location?lat=${lat}&lng=${lng}&latSpan=${boundingBox.north - boundingBox.south}&lngSpan=${boundingBox.east - boundingBox.west}&radius=1500`
```

So on a wide viewport the east-west extent the user can see is wider than the window actually queried, and stops near the left and right edges are missing from the map.

The rest of the codebase already uses `lonSpan` (`serverCache.test.js`, `mathUtils.test.js`), which is what makes this one the outlier.

### Change

One line, keeping the existing `lng` to `lon` mapping style in the same object:

```js
lonSpan: lngSpan,
```

The client keeps sending `lngSpan`; only the name handed to the SDK changes.

### Test

New `src/tests/api/stops-for-location.test.js` asserts the SDK receives `lonSpan`. On unmodified develop:

```
× forwards the client lngSpan as the SDK lonSpan
  → expected undefined to be 0.05
```

Passes with the change. Prettier clean.

Unrelated note: `src/tests/lib/dateTimeFormat.test.js` fails locally for me on develop too, expecting `9:15 AM` where an `en-IN` locale renders `9:15 am`. Not touched here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected location-based transit stop searches so longitude span parameters are properly handled, improving result accuracy.

- **Tests**
  - Added coverage to verify that longitude span requests are forwarded correctly to the transit service.

### DIFF
--- a/src/routes/api/oba/stops-for-location/+server.js
+++ b/src/routes/api/oba/stops-for-location/+server.js
@@ -13,7 +13,7 @@ export async function GET({ url }) {
 		lat: lat,
 		lon: lng,
 		latSpan: latSpan,
-		lngSpan: lngSpan,
+		lonSpan: lngSpan,
 		radius: radius
 	};
 
--- a/src/tests/api/stops-for-location.test.js
+++ b/src/tests/api/stops-for-location.test.js
@@ -1,0 +1,37 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const listMock = vi.hoisted(() => vi.fn());
+
+vi.mock('$lib/obaSdk', () => ({
+	default: { stopsForLocation: { list: listMock } },
+	handleOBAResponse: (response) => new Response(JSON.stringify(response))
+}));
+
+vi.mock('$lib/agencyFilter.js', () => ({
+	getAgencyFilter: () => null,
+	filterStops: (stops) => stops
+}));
+
+import { GET } from '../../routes/api/oba/stops-for-location/+server.js';
+
+describe('GET /api/oba/stops-for-location', () => {
+	beforeEach(() => {
+		listMock.mockReset();
+		listMock.mockResolvedValue({ code: 200, data: { list: [] } });
+	});
+
+	// MapView sends lngSpan; the SDK's parameter is lonSpan, so forwarding the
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Verified against the SDK: `onebusaway-sdk/resources/stops-for-location.d.ts:46` declares `lonSpan`, so the route was handing the SDK a key it doesn't read and the longitude span was being dropped on every single call. Silent, too — no error, just a subtly wrong bounding box.

Translating at the boundary is the right shape: keep reading `lngSpan` from the query string where `MapView.svelte:200` writes it, and hand the SDK the name it actually wants. The rest of the codebase already uses `lonSpan` consistently (`mathUtils.js`, `serverCache.js`), so this lines up.

The test is a real regression guard rather than a passing-by-construction one — asserting both `lonSpan === 0.05` and `lngSpan === undefined` fails against the old code. Good catch on something that left no trace.

PR #613, opened 2026-09-08
feat(a11y): implement the combobox pattern on the trip-planner autocomplete

### LINKED ISSUE #527: a11y: add full combobox/listbox ARIA pattern to trip-planner autocomplete

omlahore:
Closes #527.

Follow-up to #517 / #525. #525 fixed the invalid HTML; this adds the pattern itself.

The location field rendered suggestions as a plain `ul` of buttons. A screen reader got no announcement that a list had appeared, no count and no position in it, and there was no keyboard path through the suggestions at all: reaching them meant tabbing out of the input and through each result in turn.

The input is now the combobox. It carries `aria-expanded`, `aria-controls`, `aria-activedescendant` and `aria-autocomplete="list"`, the `ul` is a `listbox` and each result an `option` with `aria-selected`. Arrow keys move the active option and wrap at both ends, Enter takes the active one, Escape hides the list without discarding what was typed, and hovering syncs the active option so pointer and keyboard agree.

**Four existing tests changed, so it is worth saying why.** They pinned the old behaviour: focus a result, press Enter on it, assert the results are focusable. The combobox pattern deliberately removes that. Options are not tab stops, focus stays on the input, and `aria-activedescendant` carries the position. Keeping those assertions would have meant keeping two competing keyboard models in one widget. Each one is rewritten against the new contract rather than deleted, and six more are added covering the wiring, the wrap-around at both ends, Escape, the collapsed state and `aria-selected`. That file goes from 24 tests to 30, all passing.

No new i18n keys. The listbox is reachable and announced through `aria-controls` from the already-labelled input, so adding a string here would have meant 25 locale files of debt for no behavioural gain.

Verification: `vitest` on the trip-planner suite is 84 passing across 7 files, `svelte-check` reports 0 errors and 0 warnings, and `prettier` plus `eslint` are clean. `src/tests/lib/dateTimeFormat.test.js` has 15 failures on my machine, but they reproduce identically on an unmodified `develop`, so they are pre-existing and timezone-dependent rather than anything here.

## Summary by CodeRabbit

- **New Features**
  - Added keyboard navigation for search suggestions, including arrow-key movement, wrapping, Enter selection, and Escape dismissal.
  - Improved autocomplete accessibility with combobox and listbox semantics.
  - Added active-option highlighting and selection behavior while keeping focus in the search field.

- **Tests**
  - Expanded coverage for keyboard interactions, accessibility attributes, selection behavior, and empty results.

### DIFF
--- a/src/components/trip-planner/TripPlanSearchField.svelte
+++ b/src/components/trip-planner/TripPlanSearchField.svelte
@@ -24,16 +24,60 @@
 		onSelect
 	} = $props();
 
+	const listboxId = `${inputId}-listbox`;
+	const optionId = (index) => `${inputId}-option-${index}`;
+
+	// Which option aria-activedescendant points at. -1 means the user is on the
+	// input itself and Enter should submit whatever they typed.
+	let activeIndex = $state(-1);
+	// Escape hides the list without clearing the query. Typing brings it back.
+	let dismissed = $state(false);
+
+	let isOpen = $derived(!isLoading && !dismissed && results?.length > 0);
+	// The parent owns `results`, so it can shrink underneath a stale activeIndex.
+	let activeOption = $derived(isOpen && activeIndex < results.length ? activeIndex : -1);
+
 	function handleInput(event) {
+		activeIndex = -1;
+		dismissed = false;
 		onInput(event.target.value);
 	}
 
 	function handleClear() {
+		activeIndex = -1;
+		dismissed = false;
 		onClear();
 	}
 
 	function handleSelect(result) {
+		activeIndex = -1;
+		dismissed = false;
 		onSelect(result);
+	}
+
+	function handleKeydown(event) {
+		if (event.key === 'Escape') {
+			if (isOpen) {
+				event.preventDefault();
+				dismissed = true;
+				activeIndex = -1;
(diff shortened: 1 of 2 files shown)


omlahore:
Closing this in favour of #592, which came first and is further along.

I only checked that #527 was unassigned, not whether there was already an open PR against it, and that is on me. @soumajitgh opened theirs on 20 August and it covers the same ground plus the Escape propagation fix in `keybinding.js`, which mine does not touch.

